PR #537, opened 2014-11-03
Support v-class attr of svg in IE

tyage:
in IE11, `el.classList` is not defined when el is `SVGElement` (such as <svg>, <g>, etc...)

So `<svg v-class="">`, `<svg v-transition>` causes error in IE11.

example: http://jsfiddle.net/tyage/dpt6L1o1/

I fixed this problem.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -290,7 +290,7 @@ var utils = module.exports = {
      *  get class name for IE
      */
     getClassName: function(el) {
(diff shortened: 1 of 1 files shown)


Mikecarbon:
424

Mikecarbon:
> Thanks @tyage ! Note 0.10.6 will no longer be updated, but I will incorporate your change into 0.11 branch.

424

PR #572, opened 2014-11-17
Make `json` filter bi-directional

BigBlueHat:
I do this sort of thing pretty often:
http://bigbluehat.github.io/vue-uri/
(Output JSON into a textarea and react to it's changes).

So far, that's been with custom code. Would be very handy to have this be "native."

Thanks for considering this @yyx990803! :smile:

### DIFF
--- a/src/filters/index.js
+++ b/src/filters/index.js
@@ -6,8 +6,17 @@ var _ = require('../util')
  * @param {Number} indent
  */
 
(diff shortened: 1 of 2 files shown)


BigBlueHat:
Using this again right now, and am thankful you rewrote this and got this into master. Much thanks! :smile:

PR #650, opened 2014-12-22
Update dom.js

boussou:
I had to do this to fix exceptions on target (undefined). I don't really see why or when it happens (was not happening on v0.10.6)


yyx990803:
It's still a bit hard to identify what is really wrong here, since your code sample alone seems to be working fine. I feel that simply checking `target` existence in the replace function doesn't address the root cause. I'd appreciate it if you could provide a live reproduction of the error.

yyx990803:
Closing (inactivity)

PR #683, opened 2015-01-21
support for js File array in the model of a <input type="file">

dannydulai:
The input type='file' doesnt have a element.value like the rest of the inputs, and it pretty much always operates like a lazy input.

All this change does is make the model populated by the element's File array in element.files, so then you can use FileReader or whatever else to get the contents of the file that was marked for uploading, including it's size and filename and path.

### DIFF
--- a/src/directives/model/default.js
+++ b/src/directives/model/default.js
@@ -6,9 +6,12 @@ module.exports = {
     var self = this
     var el = this.el
 
+    // - type='file' passses the File array into the model
+    var file = el.type == "file"
+
     // check params
-    // - lazy: update model on "change" instead of "input"
-    var lazy = this._checkParam('lazy') != null
+    // - lazy: update model on "change" instead of "input". Also do this for type='file'.
+    var lazy = this._checkParam('lazy') != null || file
     // - number: cast value into number when updating model.
     var number = this._checkParam('number') != null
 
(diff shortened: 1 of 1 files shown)


dannydulai:
in the meanwhile, my workaround is:

1) add a `v-ref` to your input: `<input v-ref='myfileinput' type='file'>`
2) add this to your viewmodel: `compiled: function() this.$$.myfileinput.onchange = function(e) { v.files = v.$$.myfileinput.files; }`

src/directives/model/default.js
```diff
@@ -31,7 +34,7 @@ module.exports = {
     // shared setter
     function set () {
       self.set(
-        number ? _.toNumber(el.value) : el.value,
+        file ? el.files : ( number ? _.toNumber(el.value) : el.value ),
```
yyx990803: It should be noted that `el.files` is a `FileList` object which is array-like, but not really an array (thus when stringifying it looks different from a stringified array). The items inside are also `File` objects. In theory you should only be putting plain objects in the vm's data, so maybe we need to process `el.files` into a proper array of plain objects here.

dannydulai: Oh, but doing that would make it un-passable to other browser APIs, thus making it useless.

yyx990803: The point is Vue will try to convert the object properties into getter/setters and they may or may not break these native objects. And if we skip conversion for these objects, they won't be able to be used for data-binding.

yyx990803:
Sorry for leaving this one hanging for so long - we will likely revisit input[type="file"] handling post 1.0, so closing this.

PR #691, opened 2015-01-27
support jQuery change event

zhongxingdou:
jQuery have a lot of UI components，such as bootstrap-datepicker, select2, these components will change value of form control and won't trigger native change event. instead of,  these components trigger a internal jQuery change event, so we can listen jQuery change event in Vue for ViewModel

### DIFF
--- a/src/directives/model/default.js
+++ b/src/directives/model/default.js
@@ -79,6 +79,9 @@ module.exports = {
 
     this.event = lazy ? 'change' : 'input'
     _.on(el, this.event, this.listener)
+    if (typeof(jQuery) === 'function') {
+      jQuery(el).on('change', this.listener)
+    }
 
     // IE9 doesn't fire input event on backspace/del/cut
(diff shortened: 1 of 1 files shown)


dannydulai:
Great fix.. I was having similar issue using jQueryUI's autocomplete widget. I worked around it by ignoring v-model and just accessing the input directly, but this is so much cleaner

boussou:
+1 ^_^

406087475:
+++++++++1

yyx990803:
This is a smart fix :)
Can you use `===` and squash the commits into a single one?

zhongxingdou:
@yyx990803  ok, I changed.

yyx990803:
Ahh, so I merged this without realizing it was a PR against the `master` branch, so I reverted it. Sorry about this. Could you please make a new PR against the `dev` branch instead? (And ideally rebase before that)

PR #705, opened 2015-02-06
Add docs badge to README

rrrene:
Hi Evan,

I want to propose to add this badge to the README to show off inline-documentation: [Inline docs](http://inch-ci.org/github/yyx990803/vue)

The badge links to [Inch CI](http://inch-ci.org) and shows an evaluation by [InchJS](http://trivelop.de/inchjs), a project that tries to raise the visibility of inline-docs. Besides testing and other coverage, documenting your code is often neglected although it is a very engaging part of Open Source.

So far over 500 **Ruby** projects are sporting these badges to raise awareness for the importance of inline-docs and to show potential contributors that they can expect a certain level of code documentation when they dive into your project's code and motivate them to eventually document their own.

I would really like to do the same for the **JavaScript** community and roll out support for JS over the coming weeks (early adopters are [forever](https://github.com/foreverjs/forever), [node-sass](https://github.com/sass/node-sass) and [Shipit](https://github.com/shipitjs/shipit)).

Although this is "only" a passion project, I really would like to hear your thoughts, critique and suggestions. Your status page is http://inch-ci.org/github/yyx990803/vue

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center"><a href="http://vuejs.org" target="_blank"><img width="100"src="http://vuejs.org/images/logo.png"></a></p>
 
-# Vue.js [![Build Status](https://travis-ci.org/yyx990803/vue.svg?branch=master)](https://travis-ci.org/yyx990803/vue) [![Selenium Test Status](https://saucelabs.com/buildstatus/vuejs)](https://saucelabs.com/u/vuejs) [![Coverage Status](https://img.shields.io/coveralls/yyx990803/vue.svg)](https://coveralls.io/r/yyx990803/vue?branch=master)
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the suggestion, but looks like it misses a lot of inline docs from Vue's source code, and I don't really want to tweak my comment style just to get this badge working properly.

rrrene:
Yeah, as mentioned, Inch is still learning. And you should by no means have to change your source to accomodate a tool!

Could you point me to some of the "a lot of inline docs" that are missed? because a cursory look at the source yielded no false-positives regarding "is documented but detected as undocumented".

yyx990803:
For example, the first item in "not documented" was `apply` in transition module, which is exported and also have a function name:

``` js
exports.apply = function apply () {
  //...
}
```

Also, I put variables that are only used for a single function right above that function's signature, after the comments for that function. This also gets flagged as "undocumented".

PR #709, opened 2015-02-10
fix #707: v-repeat nested repeats on object

### LINKED ISSUE #707: 0.11.5 breaks repeat using `$data`
I'm writing an application that needs to list arbitrary values in a JSON object two levels deep. Some of these values are arrays whose constituent objects contain `id` and `value` properties.

Consider the following code:

```js
...
data : function() {
  return {
    model : {
      "arbitrary_key" : [
        { "id" : "1", "value" : "arbitrary value 1" },
        { "id" : "2", "value" : "arbitrary value 2" },
        { "id" : "3", "value" : "arbitrary value 3" },
        { "id" : "4", "value" : "arbitrary value 4" }
      ]
    }
  }
}
...
```

```html
<li v-repeat="model">
  
  {{ $key }}
  
  <ul>
    <li v-repeat="$data">
      <label>{{ id }}</label>
      <var>{{ value }}</var>
    </li>
  </ul>
  
</li>
```

The code above outputs the list of four items as expected using Vue 0.11.4, but outputs nothing when using Vue 0.11.5. I don't believe there is a test case to cover this particular use case.

calebboyd:
Fixes https://github.com/yyx990803/vue/issues/707

### DIFF
--- a/src/directives/repeat.js
+++ b/src/directives/repeat.js
@@ -268,7 +268,7 @@ module.exports = {
     }
     var raw = this.converted ? data.value : data
     var alias = this.arg
-    var hasAlias = !isPlainObject(raw) || alias
+    var hasAlias = !isObject(raw) || !isPlainObject(data) || alias
     // wrap the raw data with alias
     data = hasAlias ? {} : raw
     if (alias) {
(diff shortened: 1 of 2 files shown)


yyx990803:
Great work, thanks a lot!

PR #718, opened 2015-02-20
fix #712: allow filters on options expression of v-model

### LINKED ISSUE #712: Issues with the options directive for the select tag
I've notice two issues with `<select>`s in Vue recently.

1. The `options` directive cannot take filters. This is particularly annoying because it's rare that your data will already be in the format: `{text: "", value: ""}`
2. The `options` directive doesn't work without `v-model`. This is less likely to be a problem, but it still seems unnecessary. It's definitely possible to want dynamic select options without data binding.

Here's a [JSFiddle](http://jsfiddle.net/yye93s4g/1/) demonstrating the problems.

```html
A select with v-model and options works:
<select v-model="model1" options="arr"></select>
<br>But without v-model the options aren't shown:
<select options="arr"></select>
<br>Also filters don't work on the options argument:
<select v-model="model2" options="arr | doNothing"></select>
<br>
```

```javascript
new Vue({
    el: "body",
    data: {
        model1: "a",
        model2: "a",
        arr: ["a", "b", "c", "d"],
    },
    filters: {
        doNothing: function (value) {
            return value;
        }
    }
});
```

calebboyd:


TMiguelT:
+1!

src/directives/model/select.js
```diff
@@ -74,9 +76,11 @@ function initOptions (expression) {
   }
   this.optionWatcher = new Watcher(
     this.vm,
-    expression,
-    optionUpdateWatcher,
-    { deep: true }
+    descriptor.expression,
+    optionUpdateWatcher,{
```
yyx990803: Can you put the opening brace to a new line of its own here?

src/directives/model/select.js
```diff
@@ -74,9 +76,11 @@ function initOptions (expression) {
...
+      deep: true,
+      filters: _.resolveFilters(this.vm,descriptor.filters)
```
yyx990803: 1 space after the comma for arguments please :)

calebboyd:
Updated and rebased -- Sorry about the slop! :/

yyx990803:
Looking good - thanks!

PR #720, opened 2015-02-21
Return empty string for Infinity and properly format values such as 0.999

44px:
This PR fixes two issues in currency filter (demo for both: http://jsfiddle.net/k5bv3r62/ ):

1) x.999 values are shown as $x.00
Say you have model with value 0.999. With currency filter applied it will be shown as $0.00

2) Infinity value show as $In,finity.ty
I think, it would be good to return empty string as for falsy values

Also I added new test cases for described issues.

### DIFF
--- a/src/filters/index.js
+++ b/src/filters/index.js
@@ -61,14 +61,15 @@ var digitsRE = /(\d{3})(?=\d)/g
 
 exports.currency = function (value, sign) {
   value = parseFloat(value)
-  if (!value && value !== 0) return ''
+  if (!isFinite(value) || (!value && value !== 0)) return ''
   sign = sign || '$'
   var s = Math.floor(Math.abs(value)).toString(),
     i = s.length % 3,
     h = i > 0
       ? (s.slice(0, i) + (s.length > 3 ? ',' : ''))
       : '',
-    f = '.' + value.toFixed(2).slice(-2)
+	v = Math.abs(parseInt((value * 100) % 100, 10)),
+    f = '.' + (v < 10 ? ('0' + v) : v)
   return (value < 0 ? '-' : '') +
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks for the PR. Would it be weird to have `Inifinity` output as empty string though? Maybe it should just print `"Infinity"`?

44px:
Thanks for response.

That's interesting question. Here is my thoughts:
It seems that printing an empty string is a default way to say that value is wrong. Currency filter does it for `undefined` and for `NaN`. And it would be less weird to see 'Amount: ' than 'Amount: NaN' or 'Amount: Infinity' somewhere in interface.

Also I made this test with AngularJS, it prints empty string: http://jsfiddle.net/41Lga6bg/

src/filters/index.js
```diff
@@ -61,14 +61,15 @@ var digitsRE = /(\d{3})(?=\d)/g
...
-  if (!value && value !== 0) return ''
+  if (!isFinite(value) || (!value && value !== 0)) return ''
   sign = sign || '$'
   var s = Math.floor(Math.abs(value)).toString(),
     i = s.length % 3,
     h = i > 0
       ? (s.slice(0, i) + (s.length > 3 ? ',' : ''))
       : '',
-    f = '.' + value.toFixed(2).slice(-2)
+	v = Math.abs(parseInt((value * 100) % 100, 10)),
```
yyx990803: fix indent here?

yyx990803:
Sounds good, it's probably a good idea to keep the filter consistent with Angular's implementation.

src/filters/index.js
44px: Fixed.

yyx990803:
Thanks!

PR #731, opened 2015-03-01
fix #721: Support sorting by '$key' and '$value' in 'orderBy'

### LINKED ISSUE #721: Support sorting by `$key` and `$value` in `orderBy`
from https://github.com/vuejs/Discussion/issues/118

kazupon:


yyx990803:
Thanks for the PR! I actually was fixing the same issue, and I found an easier solution: https://github.com/yyx990803/vue/commit/fdfac339b0add3bcb9b1a01968ad0851f4bdd82d

kazupon:
Cool !! :+1:

PR #745, opened 2015-03-17
Update default.js

janschoenherr:
Added support for jQuery input event.


yyx990803:
The tests were failing and I found that we also need to change the remove listener part in `unbind()`. Turns out it's also a bit tricky to get test coverage right, so I just went ahead and changed things around. Thanks a lot for the PR though.

janschoenherr:
Thanks! Looking forward to the new release :thumbsup:

PR #755, opened 2015-03-25
data define support getter and setter

zhongxingdou:


yyx990803:
It's actually possible to support real ES5 getter/setters by using `getOwnPropertyDescriptor`, but I decided not to do that because:

1. Performance concerns;
2. Vue's data objects are meant to be static, which means they should be JSON.stringify compliant so that they can be easily persisted to the backend.

For dynamic view state we have computed properties. Do you think there are use cases that getter/setters on the data object can be particularly useful?

PR #756, opened 2015-03-25
support define a watcher whose key is symbol * to watch all properties of data

zhongxingdou:


zhongxingdou:
example
```javascript
new Vue({
    data: {
        name: "hal",
        age: 30
    },
    watch: {
       "*": function(prop, newVal, val){
            console.log(prop + " changed: " + val + " --> " + newVal);
       }
   }
})
```

TMiguelT:
I'm not at my computer to check this but can't you just watch "$data" with Vue to the same effect?

TMiguelT:
Ah no you can't! +1 Although perhaps watch $data might be more intuitive. 
http://jsfiddle.net/4qs8kkan/1/

yyx990803:
But this would not trigger on nested data changes, right?

PR #757, opened 2015-03-25
implementation new Vue option 'guard' which supported listen data before...

zhongxingdou:
... change event

### DIFF
--- a/src/api/data.js
+++ b/src/api/data.js
@@ -95,6 +95,35 @@ exports.$watch = function (exp, cb, deep, immediate) {
   }
 }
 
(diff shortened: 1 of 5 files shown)


yyx990803:
This has some overlap with write filters - although my general opinion remains that Vue data objects should behave just like plain objects and manipulating it should not give you surprises (i.e. if I set `a.b = 123` I expect `a.b === 123`)

zhongxingdou:
I'm glad to discuss all  with you. For save my English. Could you added my QQ 624998961

PR #762, opened 2015-04-01
fix <option value="" selected> is undefined

nazo:
I found a problem that initial value of v-model of <select> is treated as undefined if contains <option value="" selected>.
This fix, <option value="" selected> is treated as default value.


yyx990803:
Sorry, I merged it without realizing it's on `master`. Can you re-submit the PR against `dev` branch instead?

PR #764, opened 2015-04-03
make example of vm ready bug in phantom.js

tscanlin:
@yyx990803 this doesn't seem to be a bug in the browser, but from phantom.js. I would expect the ready function for a child component to fire when running in phantom / casper, but it seems that it is not. I think the culprit is util.inDoc returning false. So maybe the version of webkit that phantom / casper use doesn't support document.documentElement.contains (at least for comments)?

One thing to note is that I am using windows...

overriding util.inDoc with $.contains seems to fix it. I'm gonna try to find a better way though.

### DIFF
--- a/examples/nested/index.html
+++ b/examples/nested/index.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Vue.js nested demo</title>
+    <script src="../../dist/vue.js"></script>
+  </head>
+  <body>
+
+    <!-- item template -->
+    <script type="text/x-template" id="child-template">
+      <span class="three" v-text="state.created"></span>
+      <span class="four" v-text="state.ready"></span>
+    </script>
+
+    <!-- the demo root element -->
+    <div id="demo">
+      <span class="one" v-text="state.created"></span>
+      <span class="two" v-text="state.ready"></span>
+
+      <div v-component="child">
+      </div>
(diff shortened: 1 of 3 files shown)


tscanlin:
Both of the below are taken from jquery and seem to work as ok replacements:

```
exports.inDoc = function (node) {
	var adown = doc.nodeType === 9 ? doc.documentElement : doc,
		bup = node && node.parentNode;
	return doc === bup || !!( bup && bup.nodeType === 1 && (
		adown.contains ?
			adown.contains( bup ) :
			doc.compareDocumentPosition && doc.compareDocumentPosition( bup ) & 16
		));
}
```
---------- OR -----------
```
exports.inDoc = function (node) {
	if ( node ) {
		while ( (node = node.parentNode) ) {
			if ( node === doc ) {
				return true;
			}
		}
	}
	return false;
}
```

Let me know if there's a more appropriate place for this if you are able to confirm the issue. This PR isn't intended to be merged, just serve as an example to diagnose the problem.

yyx990803:
Nice! I've noticed this issue when running the unit tests in phantomjs and the first of your two implementations seem to be a good option. Do you want to open a PR for it?

tscanlin:
Sure, I'd be happy to! Here's a PR: https://github.com/yyx990803/vue/pull/767

tscanlin:
The fix for this was merged into dev in this PR: https://github.com/yyx990803/vue/pull/768

PR #772, opened 2015-04-08
Allow isReserved() to handle numeric inputs

njovin:
Currently, when an object has numeric keys and you try to $add or $delete any of those keys, an error is thrown since isReserved is expecting a string.  If this function is allowed to return false when a numeric key is passed to it, the delete operation works as expected.

### DIFF
--- a/src/util/lang.js
+++ b/src/util/lang.js
@@ -6,10 +6,7 @@
  */
 
 exports.isReserved = function (str) {
-  if(typeof str == 'number') {
-    return false;
-  }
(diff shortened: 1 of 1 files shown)


yyx990803:
It's probably easier to just cast the argument into a string (`str = '' + str`), and that would also handle more types than just numbers and strings.

njovin:
Yep, that's the better solution. Pull request modified.

yyx990803:
Looks good now, but can you submit against `dev` branch instead? `master` is only updated when there's a release.

PR #780, opened 2015-04-15
Allow isReserved() to handle non-string inputs

njovin:
Carried over from #772 - change made on dev branch as requested.


yyx990803:
Thanks!

PR #783, opened 2015-04-15
Fix nested repeats with track-by.

### LINKED ISSUE #782: nested track-by breaks when replacing the whole object
I don't know if I'm doing something wrong, but this breaks if you click the button a couple times:
http://jsfiddle.net/nkovacs/c08kdyka/

The most common error is `"Uncaught TypeError: Cannot read property '__vue__' of null".`, but there are others too.

If I remove the v-repeat of subitems, or remove the outer track-by directive, it works.

nkovacs:
When the parent vm is reused, newly created children will inherit
the _reused property, and they won't be inserted into the dom.

Fixes #782

### DIFF
--- a/src/directives/repeat.js
+++ b/src/directives/repeat.js
@@ -252,6 +252,7 @@ module.exports = {
       } else { // new instance
         vm = this.build(obj, i, true)
         vm._new = true
+        vm._reused = false
       }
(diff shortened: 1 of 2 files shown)


yyx990803:
Nice, thanks :)

PR #787, opened 2015-04-16
<content> select should only match children.

### LINKED ISSUE #786: <content> select should only select children
Nested components using multiple insertion points with select do not work properly, because the first level component's select selects the elements that are inside the nested component: http://jsfiddle.net/nkovacs/g0oq6xf8/3/

According to [this](http://www.html5rocks.com/en/tutorials/webcomponents/shadowdom/#toc-projection), select can only select elements which are immediate children of the host node. This is how it should work: http://jsfiddle.net/nkovacs/g0oq6xf8/2/

nkovacs:
Fixes #786

### DIFF
--- a/src/compiler/transclude.js
+++ b/src/compiler/transclude.js
@@ -87,11 +87,13 @@ function transcludeContent (el, raw) {
       select = outlet.getAttribute('select')
       if (select) {  // select content
         selected = raw.querySelectorAll(select)
-        outlet.content = _.toArray(
-          selected.length
-            ? selected
-            : outlet.childNodes
-        )
+        if (selected.length) {
+          outlet.content = _.toArray(selected).filter(function(element) {
+            return element.parentNode === raw
+          })
+        } else {
+          outlet.content = _.toArray(outlet.childNodes)
+        }
       } else { // default content
         main = outlet
       }
--- a/src/compiler/transclude.js
+++ b/src/compiler/transclude.js
@@ -88,12 +88,13 @@ function transcludeContent (el, raw) {
       if (select) {  // select content
         selected = raw.querySelectorAll(select)
         if (selected.length) {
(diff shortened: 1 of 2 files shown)


src/compiler/transclude.js
```diff
@@ -87,11 +87,13 @@ function transcludeContent (el, raw) {
       select = outlet.getAttribute('select')
       if (select) {  // select content
         selected = raw.querySelectorAll(select)
-        outlet.content = _.toArray(
-          selected.length
-            ? selected
-            : outlet.childNodes
-        )
+        if (selected.length) {
```
yyx990803: One possible issue here: if all nodes in `selected` are nested ones (thus should be filtered out), we end up outputting empty content, but in that case we should actually use the fallback content (`outlet.childNodes`).

According to http://w3c.github.io/webcomponents/spec/shadow/#dfn-pool-distribution-algorithm, the algorithm here should be:

- For each direct child nodes of `raw`, check if it matches `select`.
- If yes, push it into `selected`.
- If `selected` ends up empty, use `outlet.childNodes` instead.

This also means we need a `matches()` helper in `util/dom.js`:

``` js
  var p = Element.prototype
	var matches = p.matches || p.webkitMatchesSelector || p.mozMatchesSelector || p.msMatchesSelector
  exports.matches = function (el, selector) {
    return matches.call(el, selector)
  }
```

nkovacs: I wanted to avoid element.matches because its browser support is slightly worse than querySelectorAll (http://caniuse.com/#feat=matchesselector vs http://caniuse.com/#search=queryselectorall). In practice it probably doesn't matter (opera mini 8 is the only browser that's still used of those, and I don't know if vue even supports it), but I decided to play it safe.

yyx990803:
Thanks!

PR #808, opened 2015-04-27
let v-events accept statements

dgerber:
This is a tentative implementation... It allows to use metaproperties like $index, $key in the handler without the child vm having to know about them.


yyx990803:
Good work, thanks!

PR #834, opened 2015-05-12
allow for object definitions as filter arguments

janschoenherr:
I would like to propose a change to the filter token regex. This will allow for object definitions to be passed as a single argument to the filter. This will only work for shallow objects. For deep objects we would have to use a better parser.

e.g.:

    {{ '%count% Pull Requests' | trans {count: count} }}

currently it works without spaces only:

    {{ '%count% Pull Requests' | trans {count:count} }}


yyx990803:
How would the filter implementation work in this case? Do you `JSON.parse` the thing?

janschoenherr:
Currently I am using the Vue expression parser to evaluate the expression.

PR #836, opened 2015-05-13
mis pull...

TerenceZ:
update from vue/dev


TerenceZ:
This is a miss pulling... Please ignore it...

PR #839, opened 2015-05-13
Digest watcher undefined fix, fixes #838

### LINKED ISSUE #838: Possible Digest bug
https://github.com/yyx990803/vue/blob/master/src/instance/scope.js#L117

If length is 1, the index will be `1` and the result will be `undefined`.

knownasilya:
Maybe there's a better way to do this..

This basically breaks all subsequent watches..

### DIFF
--- a/src/instance/scope.js
+++ b/src/instance/scope.js
@@ -113,8 +113,11 @@ exports._unproxy = function (key) {
 
 exports._digest = function () {
   var i = this._watcherList.length
+  var watcher
+  
   while (i--) {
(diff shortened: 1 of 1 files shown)


knownasilya:
ping @yyx990803

yyx990803:
Thanks for the PR, but this fix doesn't address the real cause of the error you mentioned. I'd like to understand how you ran into the issue (you mentioned sometimes it doesn't work), and ideally have a jsfiddle that reproduces the error.

PR #861, opened 2015-05-22
Support `disabled` attribute on select options

holic:
This PR adds support for an optional `disabled` setting in an `options` array for dynamic `<select>` elements.

Local tests ran fine, but Saucelabs choked so I wasn't able to test full browser compatibility. I'm worried that the `disabled=""` test isn't going to be consistent across browsers.


yyx990803:
`disabled=""` works fine across browsers, it's actually the `value: null` that makes the test fail in some browsers. I'll merge this and add the compatibility fix, thanks!

PR #882, opened 2015-05-30
New orderBy implementation

thelinuxlich:
This is much, much faster and respects unicode characters when sorting.


yyx990803:
Interesting, do you have some sort of benchmark?
Also, can you take `dist/vue.js` out of the commit?

yyx990803:
Hmm, did some benchmark but it seems to be slower than the current implementation in V8: https://gist.github.com/yyx990803/5b0193f1675274f8c43a

thelinuxlich:
True, months ago it was the opposite :)

Here is a new implementation which is slightly slower than native sort, but when the array has more than 100 elements and sorting by string, it starts to get faster

https://gist.github.com/thelinuxlich/17b2eef28870a407371a

yyx990803:
Hmm, I think most of the time we will be sorting less than 100 items though. Also, the sorting itself will actually never be the bottleneck with 1000 items taking 0.2ms, it's mostly the DOM manipulations.

Thanks for the PR though!

thelinuxlich:
You should take into consideration the noTilde function for ordering correctly unicode strings

yyx990803:
Yeah, I'll consider making it part of `util`, seems it can be used in `filterBy` as well

PR #893, opened 2015-06-04
fix getContainedComponents in IE 11

KuroGuo:


yyx990803:
Is it possible to provide a repro of the error before the fix?

KuroGuo:
In IE, nextSibling returns a text node, but text node has no method 'contains'.

src/directives/if.js
```diff
@@ -88,9 +88,11 @@ module.exports = {
     function contains (c) {
       var cur = start
       var next
+      var isElementNode
       while (next !== end) {
         next = cur.nextSibling
-        if (cur.contains(c.$el)) {
+        isElementNode = cur.nodeType === Node.ELEMENT_NODE
+        if (isElementNode && cur.contains(c.$el)) {
```
yyx990803: Thanks for the fix! The situation can be a bit more complicated though, in the case where the component is a block instance (thus its `$el` can be a TextNode too).

So I think it's better to change the check to:

``` js
if (
  cur === c.$el || // both are textNodes
  (cur.contains && cur.contains(c.$el)) // other situations
) {
  return true
}
```

src/directives/if.js
```diff
@@ -90,7 +90,7 @@ module.exports = {
       var next
       while (next !== end) {
         next = cur.nextSibling
-        if (cur.contains(c.$el)) {
+        if (cur.contains && cur.contains(c.$el)) {
```
yyx990803: we still need to check the case where `cur === c.$el`

yyx990803:
Merging so you can get credit, I'll fix with tests. :) thanks

salazr:
FYI, we the same issue, and came to the same workaround in testing (+cur.contains(el)), but kept digging, and it was triggered by IE9 when doing v-if inside a <template> tag.

yyx990803:
yep, the whole IE family doesn't implement `contains` for text nodes. In the [spec](http://www.w3.org/TR/domcore/#node) `contains` is part of the `Node` interface, which should be available for text nodes as well, but oh well, it's IE...

PR #901, opened 2015-06-06
v-repeat does not completely teardown when reduced in length

alchen:
In the latest dev branch, I ran into an issue with assigning new values to `props` used by a `component` after its associated list of items in `v-repeat` has reduced in length.

It appears that some expired items are still watching for changes; they try to set `undefined[prop]` and fail. 

JSFiddle: http://jsfiddle.net/alchen/42ht4eg0/


yyx990803:
Thanks for reporting this! This is indeed caused by the latest commit on `dev`, but the fix doesn't really address the root cause. I've made the proper fix in the commit above.

PR #902, opened 2015-06-06
Add objects support to v-attr

OEvgeny:
The goal is to add to ``v-attr`` directive support for objects:
```
<div v-attr="obj"></div>
```
Demo here: http://jsfiddle.net/nd9baygn/3/

Waiting for suggestions patiently :)

### DIFF
--- a/src/directives/attr.js
+++ b/src/directives/attr.js
@@ -6,27 +6,43 @@ module.exports = {
 
   priority: 850,
 
-  bind: function () {
-    var name = this.arg
-    this.update = xlinkRE.test(name)
-      ? xlinkHandler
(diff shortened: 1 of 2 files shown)


yyx990803:
Looking good! Let's just make things consistent to resemble the updated `v-style` implementation: https://github.com/yyx990803/vue/blob/dev/src/directives/style.js

I think the logic should be equivalent and I don't see the need to store the previous object as `this.obj`.

Nirazul:
The demo does not seem to work in latest Chrome.

OEvgeny:
Done!

yyx990803:
good job!

PR #982, opened 2015-06-28
Cater for possibility of el not existing

fergaldoyle:
While writing a drag and drop vue sortable directive which uses jQuery UI sortable (both same list and connected list sortable), I found vue was throwing `el` undefined errors a lot with various actions, e.g. dragging the first item in a list then updating the underlying array model. The element had been removed by jQuery while vue was expecting it to exist.

By checking for the existance of `el`, and returning and empty object if it does not exist, allow these two dom manipulation libraries play nicely together.

### DIFF
--- a/src/directives/repeat.js
+++ b/src/directives/repeat.js
@@ -683,9 +683,7 @@ module.exports = {
 
 function findPrevVm (vm, anchor, id) {
   var el = vm.$el.previousSibling
-  if (!el) {
-      return
-  }
+  if (!el) return
   while (
     (!el.__vue__ || el.__vue__.$options._repeatId !== id) &&
     el !== anchor
--- a/src/directives/repeat.js
+++ b/src/directives/repeat.js
@@ -684,7 +684,7 @@ module.exports = {
 function findPrevVm (vm, anchor, id) {
   var el = vm.$el.previousSibling
   if (!el) {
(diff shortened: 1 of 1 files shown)


yyx990803:
This will likely cause the list to go out of sync. I need to see a demo of jQuery UI sortable working properly with this change.

fergaldoyle:
Here is the sortable directive with stock vue.js (throws errors)
http://jsfiddle.net/fergal_doyle/5Lrtte9v/7/

And here is the same directive but with a built of vue.js including the change I made in this PR:
http://jsfiddle.net/fergal_doyle/5Lrtte9v/6/

You can drag and drop items, use methods to reverse and add items to the model in any combination and the list stays in sync.

src/directives/repeat.js
```diff
@@ -683,6 +683,9 @@ module.exports = {
 
 function findPrevVm (vm, anchor, id) {
   var el = vm.$el.previousSibling
+  if (!el) {
```
yyx990803: I finally got time to investigate this and it seems to indeed work! One thing here is that we can just `if (!el) return` here, no need for the empty object.

fergaldoyle:
I've made the changes, no empty object now being returned.

yyx990803:
Thanks!

PR #1129, opened 2015-08-07
v-model radio and checkbox expressions

fergaldoyle:
**Radio input**
Added the ability to specify an expression, via attribute `exp`, which will be evaluated on the current scope and set as the v-model value when the radio is selected, similar to Angular `ng-value`. e.g.
```html
<input type="radio" name="sample" v-model="model.radio" value="1" exp="true" />
<input type="radio" name="sample" v-model="model.radio" value="0" exp="false" />
```
If the first radio is selected, `model.radio === true` instead of `"1"`. When the model updates and `exp` is present, we match the evaluated expression rather than the `value` string when choosing which radio to set as `checked`.

**Checkbox input**
Added the ability to specify a `true-value` and `false-value` which will override the default `true`/`false` values. Similar to Angular `ng-true-value` and `ng-false-value`. e.g:
```html
<input type="checkbox" v-model="model.checkbox1" true-value="'someTrueValue'" false-value="'someFalseValue'"/>
<input type="checkbox" v-model="model.checkbox2" true-value="expression1" false-value="expression2"/>
```
If present `true-value` and `false-value` will be evaluated on the current scope, and set as the v-model value when the checkbox changes. Similar to the radio, if the model changes a comparison is made against `true-value` and `false-value` values to determine the checkbox state.


src/directives/model/checkbox.js
```diff
@@ -5,17 +5,52 @@ module.exports = {
...
+      }
+      if (!val && falseExp !== null) {
+        val = self.vm.$eval(falseExp)
+      }
+      return val
+    }
+    this._getValue = getValue
+
+    function matchValue (value) {
+      var trueValue = true
```
yyx990803: Since `el.checked` can be only `true` or `false`, we just need to check if `value` equals `trueValue` here - in all other cases we just return false. This could make `matchValue` much simpler.

src/directives/model/checkbox.js
```diff
@@ -5,17 +5,52 @@ module.exports = {
   bind: function () {
     var self = this
     var el = this.el
+    var trueExp = this._checkParam('true-value')
```
yyx990803: To be consistent with `exp` for radio input, maybe we should use `true-exp` and `false-exp`?

fergaldoyle:
Yes makes sense to use true-exp and false-exp. Also simplified `matchValue`, either the true/trueExp matches the incoming value or it does not, which implies a false match.

https://github.com/yyx990803/vue/commit/b8e364fd3b0a1231452f0e425a0e75e95d174f93

yyx990803:
Cool! Mind rebasing against latest `dev` so the CI build can pass?

fergaldoyle:
Hitting the limits of my git knowledge here, I've re-based `v-model-enhancments` onto `dev`, will I have to open a new pull request?

yyx990803:
Here's what you should **probably** do:

```
git checkout dev
git pull https://github.com/yyx990803/vue.git dev
git checkout v-model-enchancements
git rebase dev
git push -f
```

fergaldoyle:
That seemed work

yyx990803:
Thanks!

PR #1133, opened 2015-08-08
Hide directive

pespantelis:
It is the opposite of `show` directive.

### DIFF
--- a/src/directives/hide.js
+++ b/src/directives/hide.js
@@ -1,0 +1,8 @@
+var transition = require('../transition')
+
+module.exports = function (value) {
+  var el = this.el
+  transition.apply(el, value ? 1 : -1, function () {
+    el.style.display = value ? 'none' : ''
+  }, this.vm)
+}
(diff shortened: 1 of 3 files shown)


azamat-sharapov:
Doesn't this make API more verbose and add more size to library? Why not just `v-show="!show"`?

pespantelis:
IMO `v-hide="show"` is more clean and readable from `v-show="!show"`.
Regarding the library, it is increased less than half of KB.

yyx990803:
Thanks for the PR, but let me explain why I don't want to add `v-hide`: the general philosophy behind Vue is we want to keep it as lightweight as possible, so when we add code into the codebase, we want to make sure the code either provides substantial improvement to an existing feature, or adds a new functionality that cannot be trivially achieved with the current API. In general, we want to avoid adding code just for the sake of stylistic alternatives. In fact we have been trying to removing stylistic alternatives as much as possible (e.g. `v-with` vs `props`).

`v-hide`, in my opinion, is just a stylistic alternative to `v-show`, and its benefit over `v-show` is somehow subjective and marginal.

Regardless, thanks for your PR and hope to see you contributing to Vue again in the future :)

pespantelis:
Thank you for the detailed explanation :)

Issue #1131, opened 2015-08-08
Suggestion about `currency` filter


ratiw:
Hi, I've been experimenting on Vue.js a little bit and feel really great about it. 

However, in my opinion, the `currency` filter is a bit restricted that it doesn't allow `''` empty string to be used. As in my country, the currency symbol is not very popular and most of the time get ignored by the users; therefore, no symbol is used in that place. But we do format the currency using thousand separator and decimal point as usual.

I know that I can pass a space to get rid of the currency symbol, but that actually use a space in front of the currency value is not very ideal. I think allowing an empty string to be used in the currency symbol would be more appropriate (so there will be no space in front of the currency value).

This is just a suggestion. Thanks for you excellent work on Vue.js anyway.

Cheers,
Rati

PR #1135, opened 2015-08-08
Currency symbol

azamat-sharapov:
Looks like `''` and `null` are casted into `undefined`, so I chose `false` to be as "no-symbol param". 

What do you think?

### DIFF
--- a/src/filters/index.js
+++ b/src/filters/index.js
@@ -61,7 +61,7 @@ var digitsRE = /(\d{3})(?=\d)/g
 exports.currency = function (value, currency) {
   value = parseFloat(value)
   if (!isFinite(value) || (!value && value !== 0)) return ''
-  currency = currency || '$'
+  currency = currency != null ? currency : '$'
   var stringified = Math.abs(value).toFixed(2)
   var _int = stringified.slice(0, -3)
   var i = _int.length % 3
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,4 +81,4 @@ The default task (by simply running `grunt`) will do the following: lint -> buil
 
 The unit tests are written with Jasmine and run with Karma. The e2e tests are written for and run with CasperJS.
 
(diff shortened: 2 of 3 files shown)


src/filters/index.js
```diff
@@ -61,7 +61,7 @@ var digitsRE = /(\d{3})(?=\d)/g
 exports.currency = function (value, currency) {
   value = parseFloat(value)
   if (!isFinite(value) || (!value && value !== 0)) return ''
-  currency = currency || '$'
+  currency = currency === false ? '' : currency || '$'
```
yyx990803: Looking at #1131 , it seems we should just check if `currency` is undefined or null.
So something like `currency = currency != null ? currency : '$'`

azamat-sharapov: Hmm, in suggestion, OP needs ability to filter number through currency filter *without adding symbol*. How would `currency = currency != null ? currency : '$'` do that?

yyx990803: so that he can do `money | currency ''`. Currently empty string is considered falsy and gets casted into `$`.

azamat-sharapov: well, as I said in PR description, `''` is casted to `undefined` :) Is this bug somewhere in parser?  

Otherwise I was going to use empty string as a *no-symbol param* also. Only "ok" choice was to take `false` as *no-symbol param*. Another choice is to remove this slightly discriminating `$` default :smile: , but that would be breaking change.

yyx990803: Oh, interesting, that sounds like a filter bug! I'll take a look.

yyx990803: fixed!

azamat-sharapov: heh, did I ever speak about speed of you coding? :)
is it fine if I force push into this branch? wanna re-create branch. Or better open another PR?

yyx990803: just force push is fine

yyx990803:
:tada:

PR #1140, opened 2015-08-10
Debounce filter

karevn:
An implementation for event debounce filter inspired by this thread: https://github.com/vuejs/Discussion/issues/245


src/filters/index.js
```diff
@@ -130,6 +130,20 @@ exports.key = function (handler, key) {
 // expose keycode hash
 exports.key.keyCodes = keyCodes
 
+exports.debounce = function (handler, delay) {
```
yyx990803: This implementation doesn't pass on the arguments the handler gets. In fact you can just use the internal `_.debounce` to wrap the handler.

test/unit/specs/filters/filters_spec.js
```diff
@@ -96,6 +96,17 @@ describe('Filters', function () {
...
   })
 
+  it('debounce', function (done) {
+    var filter = filters.debounce
+    expect(filter(null)).toBeUndefined()
+    var spy = jasmine.createSpy('filter:debounce')
+    var handler = filter(spy)
+    handler()
+    expect(spy).not.toHaveBeenCalled()
+    handler = filter(function () { done() })
```
yyx990803: make sure to test receiving proper arguments here

karevn:
Btw, why not to make `_.debounce` available as Vue global function?

yyx990803:
@karevn it is available as `Vue.util.debounce`.
Could you please rebase against current `dev` so that the CI build can pass? Thanks!

yyx990803:
This PR itself is trying to merge into `master`, but `master` only holds the latest stable release; all PRs should be submitted against `dev`. I don't think you can change the target branch of an open PR, so you will have to make a new one (and preferably make clean commits on top of `dev`)

karevn:
https://github.com/yyx990803/vue/pull/1145

PR #1151, opened 2015-08-12
Added throttle util and filter

karevn:
Hello! I think this small addition will be helpful for some cases (at least it was really helpful for my current project).

### DIFF
--- a/src/util/lang.js
+++ b/src/util/lang.js
@@ -255,6 +255,47 @@ exports.debounce = function (func, wait) {
 }
 
 /**
+* Throttle a function so it will only gets called at most
+* once during a given timespan.
+*
+* @param {Function} func
+* @param {Number} wait
+* @param {Object} options
+* @return {Function} - throttled function
+*/
+exports.throttle = function (func, wait, options) {
+    var context, args, result
+    var timeout = null
+    var previous = 0
+    if (!options) options = {}
+    var later = function () {
+      previous = options.leading === false ? 0 : Date.now()
+      timeout = null
+      result = func.apply(context, args)
+      if (!timeout) context = args = null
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks for the PR - can you give some more information on the actual use case? e.g. what type of UI are you using this for?

karevn:
I am building a "custom cursor" thingie, which should only be displayed when hovering over specific areas of the app. So, I need to check if I should display a custom cursor or default one based on `jQuery(el).closest('bla-blah').size()` call result, which is expensive. But, I don't need an immediate reaction, so if I throttle the event to be only fired once in 100msec - all works fine.

So basically, this thing will be useful for working with mouseover / scroll events when they should be fired "in-between" the action.

src/util/lang.js
```diff
@@ -255,6 +255,47 @@ exports.debounce = function (func, wait) {
...
 
 /**
+* Throttle a function so it will only gets called at most
+* once during a given timespan.
+*
+* @param {Function} func
+* @param {Number} wait
+* @param {Object} options
+* @return {Function} - throttled function
+*/
```
rafaell-lycan: :clap:

Codebryo:
is this still open/valid?

karevn:
It's just sitting and waiting :)

yyx990803:
Hey, sorry for letting this sit for so long. The reason I hesitated and decided not to merging this: it's adding weight to the library for a very edge use case, which can be easily implemented as a userland thing. In comparison, `debounce` is in core because it is commonly used on input with `v-model` and there's no easy way for a user to implement that for `v-model`.

Again, thanks for the PR.

PR #1203, opened 2015-08-25
Added namespaces for events

VladimirShestakov:


VladimirShestakov:
We can dispatch, broadcast and emit event with namespaces:

```js
this.$emit('name.xxx`);
```

```js
this.$on('name', function(..)); //work
this.$on('name.xxx', function(..)); //work
this.$on('name.xxx.yyy', function(..)); //not work
```

VladimirShestakov:
In the listener we have access to the all namespaces by this._eventNamespaces
```js
this.$on('name', function(
    if (this._eventNamespaces.indexOf('xxx')){
         console.log('is xxx');
    };
));

``

**Comment by yyx990803** [comment]
Thanks for the PR, but I don't think this design is intuitive enough... what problem does it solve? Looks like I can achieve something similar with current API:

``` js
this.$on('namespace', function (type) {
  switch (type) {
    // ...
  }
})

this.$emit('namespace', 'xxx')
this.$emit('namespace', 'yyy')
```

VladimirShestakov:
But you used similarity of namespace for lifecycle events (hooks). Namespaces make modular system more flexible for routing/listening events. They are used in the largest applications with many vendor components. Later we can improve core behavior of the Vue with namespaces.

VladimirShestakov:
It might be better to use ":" instead of "." for names splitting. Because it is used in the core

VladimirShestakov:
And I would like to add a listener of all events by an empty string (name):

```js
this.$on('', function () {
    console.log('all events');
});

This listener can be used in the root component for broadcasting all events dispatched from the children

yyx990803:
Overall I think event communication is something I want to de-emphasize for new users, because it is too easy to be abused and create hard-to-maintain situations. In the future the recommendation would be adopting a flux-like architecture for large apps, and only use events in small use cases for convenience. Making it more powerful/complicated doesn't fit with that vision. Thanks for the PR though.

VladimirShestakov:
It's not good idea to make analog of ReactJS.

PR #1206, opened 2015-08-25
New filter: `prevent`

lepture:
> Prevent default event behavior.

It is useful on `<a>` link when you want to call a function instead of visiting the link:

```html
<a href="/login" v-on="click: showLogin=true | prevent">Login</a>
```

Another use case is dropdown. GitHub has a user menu dropdown on the top right. The avatar has a link of the user profile, but it won't redirect to the profile page when clicking the avatar. Instead, it will show a dropdown menu.

```html
<a href="/{{ username }}" v-on="click: showDropdown=true | prevent">avatar</a>
<dropdown show="{{@ showDropdown }}">
    ....
</dropdown>
```

### DIFF
--- a/src/filters/index.js
+++ b/src/filters/index.js
@@ -139,6 +139,17 @@ exports.debounce = function (handler, delay) {
 }
 
 /**
+ * Prevent event default behavior
+ */
+exports.prevent = function (handler) {
+  if (!handler) return;
+  return function (e) {
+    e.preventDefault()
+    return handler.call(this, e)
+  }
+}
+
+/**
  * Install special array filters
  */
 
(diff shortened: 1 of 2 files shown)


VladimirShestakov:
:+1:

Nirazul:
I don't see why this must be in core. Maybe you additionally want to stop propagation or use prevent just under some circumstances.
It'd be perfect as a plugin, however :)

lepture:
@Nirazul it is useful, i think it is better a built-in filter like `key`. let @yyx990803 decide it.

yyx990803:
@lepture Thanks for the PR. I generally want to discourage inline state manipulations. In a use case like this I actually prefer to have an explicit method for it. And as @Nirazul pointed out, this type of sugar can be trivially done in userland, and ideally the core should be as lean as possible. (I know we have a lot of other filters, which I once considered removing, but people already use them...)

PR #1245, opened 2015-09-03
fix eslint ternary operator error

nkovacs:
eslint's operator-linebreak rule now supports the ternary operator (https://github.com/eslint/eslint/issues/3274).

Changed the rule to match the code, and fixed the single instance
in the code that was different.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -137,7 +137,7 @@
     "object-shorthand": 0,
     "one-var": [2, { "initialized": "never" }],
     "operator-assignment": 0,
-    "operator-linebreak": [2, "after"],
+    "operator-linebreak": [2, "after", { "overrides": { "?": "before", ":": "before" } }],
     "padded-blocks": 0,
     "prefer-const": 0,
     "quote-props": 0,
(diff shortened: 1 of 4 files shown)


nkovacs:
I don't know what's going on on circleci, it installed eslint 1.3.1, which supports this, and it works on my computer with eslint 1.3.1 (which is how I ran into this issue in the first place).

yyx990803:
It's a bit tricky here because the build setup currently uses `grunt-eslint` which depends on eslint `^1.1.0`. So if we update the root eslint to `1.3.1`, npm will install a separate copy of eslint under grunt-eslint.

I will likely update the build setup to get rid of `grunt-eslint` later, but for now let's stay away from altering the build setup to avoid rebase headaches from the alpha/beta branches. Thanks for raising the issue.

nkovacs:
Won't grunt-eslint's ^1.1.0 install 1.3.1 though? it did that on my computer.

yyx990803:
Ha, looks like it's because CircleCI used a cached version of eslint.

PR #1246, opened 2015-09-03
Don't set up css transition callback if there's a js callback.

### LINKED ISSUE #1241: triggering a transitionend event after enter transition is done in in-out mode causes error
It only happens with in-out mode. The css transition that triggers the transitionend event is entirely unrelated to the component transition, which in the example uses js callbacks, and has already finished by the time the transitionend event is called.

http://jsfiddle.net/nkovacs/h800dpx2/

Uncaught TypeError: Cannot read property 'parentNode' of null caused by child.$el being null [here](https://github.com/yyx990803/vue/blob/master/src/directives/component.js#L259)

nkovacs:
Fixes #1241

Leave doesn't set up the css callback if there's a js callback, so I did the same for enter.

### DIFF
--- a/src/transition/transition.js
+++ b/src/transition/transition.js
@@ -95,16 +95,20 @@ p.enterNextTick = function () {
   _.nextTick(function () {
     this.justEntered = false
   }, this)
-  var type = this.getCssTransitionType(this.enterClass)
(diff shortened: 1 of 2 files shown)


yyx990803:
:+1: nice work

Issue #1148, opened 2015-08-12
auto transform property to number even if its property assert type is string.


TerenceZ:
For example as followed, after initialisation, a warn "Expected string, got number" will be raised up.

```html
<comp a="10"></comp>
<script>
Vue.component({
    name: "comp",
    props: {
        a: { type: String }
    }
});
</script>
```

yyx990803:
This is auto casted because in most cases people do want a Number when they pass a number. If you want a literal string of numbers you'd have to do `<comp a="{{'10'}}"></comp>`.

PR #1285, opened 2015-09-09
Add mixed type properties.

### LINKED ISSUE #1284: Use the property specification for auto-casting in literal properties
Instead of always auto-casting, which leads to confusing behavior like #1148, you could use the property specification's type to determine what, if any, auto-casting should occur. For example:

```js
Vue.component('example', {
  props: {
    propString: {
      type: String
    },
    propStringOrNumber: {
      type: [String, Number]
    },
    propAllThree: {
      type: [Boolean, Number, String]
    },
  }
})
```

(I hope it's not impossible to extend type like that)
`propAllThree` would work like the current auto-casting, `propStringOrNumber` would be auto-casted to number, but not boolean, and `propString` would never be auto-casted.

If type is not specified, it could default to either `[Boolean, Number, String]` or just `[String]` *for the purposes of auto-casting only*. I'm not sure which one would be better, but in any case, you can explicitly specify the type to avoid auto-casting.

nkovacs:
Literal property auto-casting now respects property type(s),
and will not cast a property to boolean or number if it
cannot be a boolean or number.

Fixes #1284, #1148 

There were a few special cases that are a bit different for mixed types, but it should be backwards compatible:
* `default` will always be called even if one of the types is `Function`. But if the only type is `Function`, it won't be called, like before. In practice it probably doesn't makes sense to have a property that can be either a Function or something else.
* If the default is not specified and the property is omitted, the default is undefined, except for Boolean, where it is false. For a mixed type, the default will be undefined even if one of the possible types is Boolean. It might make sense to use false as the default here, but in practice both are falsy, and you can specify a default, so it's not a big issue.
* In contrast to the previous point, if one of the possible types is Boolean (explicitly specified), and the attribute value is omitted (or is an empty string), it will be cast to true, even if String is also an accepted type for the property. This is how it worked previously for Boolean type properties, so I extended that to mixed properties that include Boolean. It made more sense for a property without a value to be true in this case, instead of a falsy empty string.

For example:
```
prop1: [Boolean, String]
prop2: String

<example></example> <!-- undefined undefined -->
<example prop1 prop2></example> <!-- true, '' -->
<example prop1="foo" prop2="foo"></example> <!-- 'foo' 'foo' -->
<example prop1="123" prop2="123"></example> <!-- '123' '123' -->
```

### DIFF
--- a/src/compiler/compile-props.js
+++ b/src/compiler/compile-props.js
@@ -139,13 +139,7 @@ function makePropsLinkFn (props) {
       } else {
         // literal, cast it and just set once
         var raw = prop.raw
-        value = options.type === Boolean && raw === ''
-          ? true
-          // do not cast emptry string.
-          // _.toNumber casts empty string to 0.
-          : raw.trim()
-            ? _.toBoolean(_.toNumber(raw))
-            : raw
+        value = _.castProp(prop, raw)
         _.initProp(vm, prop, value)
       }
     }
@@ -160,12 +154,13 @@ function makePropsLinkFn (props) {
  */
 
 function getDefault (options) {
+  var types = _.normalizePropTypes(options.type)
   // no default, return undefined
   if (!options.hasOwnProperty('default')) {
     // absent boolean value defaults to false
-    return options.type === Boolean
-      ? false
-      : undefined
+    return types && types.length === 1 && types[0] === Boolean
+        ? false
+        : undefined
   }
   var def = options.default
   // warn against non-factory defaults for Object & Array
@@ -177,7 +172,7 @@ function getDefault (options) {
     )
   }
   // call factory function for non-Function types
-  return typeof def === 'function' && options.type !== Function
+  return typeof def === 'function' && !(types && types.length === 1 && types[0] === Function)
     ? def()
     : def
 }
--- a/src/util/component.js
+++ b/src/util/component.js
@@ -52,10 +52,50 @@ exports.initProp = function (vm, prop, value) {
 }
 
 /**
+ * Automatically cast a prop based on its specified type.
+ *
+ * @param {Object} prop
+ * @param {String} raw
+ * @return {String|Boolean|Number}
+ */
+exports.castProp = function (prop, raw) {
+  var hasBoolean = false
+  var hasNumber = false
+  var types = exports.normalizePropTypes(prop.options.type)
+  if (types) {
+    var i, l
+    for (i = 0, l = types.length; i < l; i++) {
+      var type = types[i]
+      if (type === Number) {
+        hasNumber = true
+      } else if (type === Boolean) {
+        hasBoolean = true
+      }
+    }
+  }
+
+  if (hasBoolean && raw === '') {
+    return true
+  }
+  var value = raw
+  if (!value.trim()) {
+    return value
+  }
+  if (!types || hasNumber) {
+    value = _.toNumber(value)
+  }
+  if (!types || hasBoolean) {
+    value = _.toBoolean(value)
+  }
+  return value
+}
+
+/**
  * Assert whether a prop is valid.
  *
  * @param {Object} prop
  * @param {*} value
+ * @return {Boolean}
  */
 
 exports.assertProp = function (prop, value) {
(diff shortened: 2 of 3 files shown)


yyx990803:
Hmm I didn't have time to respond to #1285, but with 1.0.0's syntax there's no longer the need for auto casting. Literal values are always strings; boolean and number literals need to be passed via dynamic bindings. The only auto-cast case is omitted value for Boolean props.

The ability to allow multiple types is a good improvement, but I think that should itself be a PR. Also, I somewhat prefer only auto-casting omitted value when Boolean is the only expected type.

nkovacs:
>  with 1.0.0's syntax there's no longer the need for auto casting

Why not? For example with `<example prop="123"></example>`, if prop's type is just Number, there's no harm in auto-casting it. Not auto-casting would mean this would not pass validation, and require people to use binding, even though they're not binding to a variable. That feels like a workaround for a bug to me.
Similarly, if the type is [Number, String, Boolean], the user has opted into it, so auto-casting would not cause surprise.
You could disable auto-casting when the type is not specified, but I kept it enabled for backwards compatibility, and because you said in #1148 that most people expect auto-casting.

> I somewhat prefer only auto-casting omitted value when Boolean is the only expected type

What about when the type is [Boolean, Number]? Empty string isn't allowed there, so it makes sense to  cast to true.
The alternative is to auto-cast to true if Boolean is an allowed type, unless String is also an allowed type. But I think it's simpler and more consistent to always auto-cast to true if Boolean is a possible type.

yyx990803:
In 1.0.0 `prop="123"` is a literal, thus it is expected to pass "123" as a literal string. If we do autocast, the user would have to do `bind-prop="'123'"` in order to pass the string, which is less intuitive imo. "Literal syntax = string" is a much easier rule to remember.

I'd like to treat multiple types as "any one of these types is allowed". Omitting value = true is a behavior that is strongly tied to boolean attributes only - if the user wants to use this syntax, then the prop needs to be Boolean and Boolean only. Casting it when multiple types are allowed only makes the rules more complicated than necessary.

nkovacs:
> If we do autocast, the user would have to do bind-prop="'123'" in order to pass the string, which is less intuitive imo.

Not if the property's type is just String. This would also mean you could never use literals with Number or Boolean properties, it would always fail validation.

The only case where the value is ambiguous is with mixed properties that include String, e.g. `[String, Number]` - should it auto-cast to Number or not? I'm in favor of auto-casting, since you opted into it, and you probably don't want something that looks like a number to be a string if you explicitly said that you accept a number.

For type=null, I could change it to treat it as `[String]` instead of the current backwards compatible behavior. That way by default literal properties would always be strings, and you'd have to opt into casting.

>  Omitting value = true is a behavior that is strongly tied to boolean attributes only - if the user wants to use this syntax, then the prop needs to be Boolean and Boolean only.

If the type doesn't include String, the empty string will create an error, so you can't omit the value. It's an unnecessary restriction. 
For example:

```js
props: {
  centered: {
    type: [Number, Boolean]
    default: false // not centered by default
  }
}
```
```html
<column centered></column> <!-- centered === true, i.e. use default width -->
<column centered="3"></column> <!-- centered === 3, explicit width -->
<column></column> <!-- centered === false -->
```

Also this doesn't happen for type=null, so you have to opt into it by listing Boolean as one of the possible values.
The only ambiguity is with `[String, Boolean]` and such, but you can look at that from both sides, for example:

```html
<example highlighted></example> <!-- use default highlighting -->
<example highlighted="true"></example> <!-- use default highlighting -->
<example highlighted="green"></example> <!-- use green highlighting -->
<example></example> <!-- no highlighting -->
```

If empty string is not cast to true, in this case, the first example would be `'' == false` which is wrong.

A possible solution is to allow the property to specify the value for when the value is empty, e.g.:

```js
props: {
  highlighted: {
    type: [String, Boolean],
    default: false,
    empty: true,
    validator: function(value) {
      // only allow predefined colors
      return ["green", "blue", "red", true, false].indexOf(value) !== -1;
    }
  }
}
```

yyx990803:
The issue I have with your proposal is that the same syntax, e.g. `prop="123"`, will result in different values based on the component's prop type expectations. The consumer of the component will need to understand these complex conversion rules (A only gets casted into B when the props expects type A & B together...) in order to infer what eventual value something will be casted into.

In comparison, a simpler rule set is:

1. `prop="xxx"`: always a string.
2. `:prop="xxx"`: always an expression. `123`, `true` and `false` should use this as well.
3. `prop`: only true when the type is boolean. A special case, but makes sense and easy to tell.

nkovacs:
That's true. But I don't think it's a big problem, you already have to know what the component expects when you're using it, so that you don't try to pass a number in a literal property.
The only uncertainty occurs when you're reading a template and you don't know the component's property rules.
Anyway, I can't remove the auto-casting from this, since that would break the expected behavior in 0.12. I'll rework the pull request for 1.0

yyx990803:
Thanks, closing this for now.

Issue #1249, opened 2015-09-03
v1.0.0-alpha.2: Uncaught TypeError: Cannot read property '__vfrag__' of null


weislanes:
Update: [Reproduction here](http://jsfiddle.net/mikeemisme/64scphgn/)

Occurring when dragging a list item from one unordered list to another - in this case when using RubaXa/Sortable to manage lists.

Appears to be caused by the location of the comment node after items are moved, according to comments under #437 - typeError 'Cannot read property '__vue__' of null'. That similar issue was closed as could not be reproduced. 

Billiam later provided [more details as well as a jsfiddle example](https://github.com/yyx990803/vue/issues/437#issuecomment-66904584). 

Now that v1.0.0 uses 'v-for' instead of 'v-repeat' it generates the error: Cannot read property '__vfrag__' of null.

yyx990803:
Can you provide a repro using 1.0.0-alpha.2?

weislanes:
Sure thing. (Repro created here)[http://jsfiddle.net/mikeemisme/64scphgn/]

weislanes:
Spent more time on this and don't think it's due to the comment issue that affected #437. Looks like comment nodes are not output in the DOM in Vue 1.0.0 when using v-for, as far as I can tell. 

So issue is caused by something else that I can't seem to resolve.

yyx990803:
Ok, so in 0.12, the anchor nodes for `v-repeat` will be an empty textNode in normal mode, and comment node in debug mode. So the code that hacks around comment nodes should also check for empty textNodes: http://jsfiddle.net/64scphgn/5/

weislanes:
Thanks for that info but this doesn't resolve the issue. If you walk through the reproduction steps in your updated Fiddle the 'Cannot read property '__vfrag__' of null' error still appears. 

Note that this only happens after dragging between two lists, dragging within a list appears to be OK. Any ideas?

PR #1293, opened 2015-09-10
Return if no previous sibling of el

weislanes:
Cater for scenario where there is no previous sibling of el. Solution to #1249 and similar to #982. Occurs in scenario where dragging between lists using library like RubaXa Sortable.

### DIFF
--- a/dist/vue.js
+++ b/dist/vue.js
@@ -9057,6 +9057,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	    el !== anchor
 	  ) {
 	    el = el.previousSibling
+	    if (!el) return
 	    frag = el.__vfrag__
 	  }
 	  return frag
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks for the PR, but please make sure to work on source files in `src` instead of the built file in `dist`.

weislanes:
Whoops. Very first PR - still learning. Can go ahead and make this change in correct place if required?

yyx990803:
Ideally you should submit the PR with one single commit affecting only the source file that needs to be changed. You can either squash the commits and force-push to your fork, or, if you don't know how to do that, submit a new PR instead.

weislanes:
Agreed. Didn't realise I was compounding my mistakes - however I learned a lot in those 5 commits :)  Will submit a new PR.

PR #1299, opened 2015-09-11
fix select element can't removeChild optgroup's option

evantre:
demo: http://jsfiddle.net/evantre/eokkguxx/
if change select's options, it well report error about "Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node."


yyx990803:
Thanks! :+1:

evantre:
sorry, it also has a bug. I fix it later.

PR #1304, opened 2015-09-12
update npm ignore file

kazupon:


yyx990803:
Thanks, but the `explorations` folder is a local folder that I use for debugging and experiments. It's not checked in but still needs to be ignored by npm.

PR #1316, opened 2015-09-15
Set main file to minified version

BYODKM:
To make `.bowerrc` file more useful.


yyx990803:
Thanks, but per [bower.json spec](https://github.com/bower/bower.json-spec#main), the main field should not include minified files.

PR #1372, opened 2015-09-29
add attributes available for Aria accessibility

wprater:


src/directives/public/bind.js
```diff
@@ -34,6 +34,10 @@ module.exports = {
         attr === 'class' ||
         // data attributes are allowed globally
         /^data-/.test(attr) ||
+        // aria attributes are allowed globally
+        /^aria-/.test(attr) ||
+        // role available
+        (attr === 'role' && 'htmlRole' in this.el) ||
```
yyx990803: It seems the `role` attribute doesn't have a corresponding property. We can just allow it without checking for `'htmlRole' in this.el`.

wprater: will modify; wasn't sure what the `html<attr>` was for.

wprater:
@yyx990803 removed the `htmlRole` check

PR #1437, opened 2015-10-17
allow optional `(key,val) in arr` syntax for `v-for`

vprimachenko:
optionally binding the iteration index (or key when iterating over an
object) to a name to simplify work with nested `v-for`s

### DIFF
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -14,7 +14,13 @@ module.exports = {
     // support "item in items" syntax
     var inMatch = this.expression.match(/(.*) in (.*)/)
     if (inMatch) {
-      this.alias = inMatch[1]
+      var itMatch = inMatch[1].match(/\((.*),(.*)\)/)
+      if (itMatch) {
+        this.iterator = itMatch[1]
+        this.alias = itMatch[2]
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks! Can you please add an simple test case like [this](https://github.com/yyx990803/vue/blob/1.0.0-alpha/test/unit/specs/directives/for/for_spec.js#L598-L604)?

src/directives/for.js
```diff
@@ -14,7 +14,13 @@ module.exports = {
     // support "item in items" syntax
     var inMatch = this.expression.match(/(.*) in (.*)/)
     if (inMatch) {
-      this.alias = inMatch[1]
+      var itMatch = inMatch[1].match(/\((.*),(.*)\)/)
+      if (itMatch) {
+        this.iterator = itMatch[1]
```
yyx990803: Small thing: we should call `trim()` on the iterator and alias here so that cases with extra spaces like `( key, val )` would also work.

yyx990803:
Thanks, note that further development for 0.12 compat version will be pushed to the `1.0.0-migration` branch now.

Issue #1433, opened 2015-10-16
1.0.0-rc1: v-for + filters + .sync does not work as expected


karevn:
This fiddle: http://jsfiddle.net/kyhqq375/ is expected to change "Items: one, two" when clicking buttons, but "write" filter is never called, and "one, two" is never changed. Obviously, not a simple use case, but it should work or at least show a warning...

marcvdm:
I was "fiddling" around in you jsfiddle and got the following error
``[Vue warn]: It seems you are using two-way binding on a v-for alias, and the v-for has filters. This will not work properly. Either remove the filters or use an array of objects and bind to object properties instead.``

The result of the filter is not bound to the current vm so there can't be a 2 way bind.

There are better ways to solve the problem that you are facing
http://jsfiddle.net/kyhqq375/1/

karevn:
@marcvdm that is exactly how I worked it around in my app, but it still is a workaround, not a solution.

Btw, the warning does not show in my fiddle. Probably, you replaced `item.value` binding expression with `binding`.

karevn:
Hmm. I think, a watcher should be created with "deep" option to fix it (instead of "shallow" watcher). Am I right?

yyx990803:
This is a use case that is by design not going to work. "deep" has nothing to do with this. Two-way bindings on a filtered copy doesn't really make sense. I anticipated cases like this but your usage somehow bypassed the warning check ;)

It will give a warning in such scenarios now.

karevn:
@yyx990803 may be there's still some kind of a solution for this? I think it's possible to deep-watch for changes in the filtering result, and then run the write filter to re-assign the initial value.

karevn:
@yyx990803 and I don't see reasons (other than code complexity) why should not it work. I think of it the next way: We have an initial object, which is passed through a filter to build a new one (or an array), which is then iterated by `v-for`. If filter results are changed, then we run the write filter to re-build the initial value (which is not always possible, but it's an app-specific detail). It's just a "streamlined" version of @marcvdm 's workaround, IMO.

yyx990803:
Yes, this is a marignally useful feature that involves a lot of code complexity - maybe requires rewriting some parts of how filters work. Although theoretically it "should" work, given the limited time I have I don't see it ever become a priority in the foreseeable future. You are free to investigate and submit a PR if you feel it's worth the effort.

karevn:
Okay, i'll look for a solution for this thing.
[...]

PR #1441, opened 2015-10-19
Added reverse filter support for v-for

karevn:
An implementation for this issue: https://github.com/vuejs/vue/issues/1433

### DIFF
--- a/src/directives/public/for.js
+++ b/src/directives/public/for.js
@@ -516,7 +516,7 @@ module.exports = {
     }
   },
 
-  _bubbleChanges: function (scope, value) {
+  _syncChanges: function (scope, value) {
     // two-way sync for v-for alias
     var subjectValue = this.filters ? this.processedValue
       : this.rawValue
--- a/src/directives/public/for.js
+++ b/src/directives/public/for.js
@@ -514,6 +514,23 @@ module.exports = {
         frag.destroy()
       }
     }
+  },
+
+  _bubbleChanges: function (scope, value) {
+    // two-way sync for v-for alias
+    var subjectValue = this.filters ? this.processedValue
+      : this.rawValue
+    if (scope.$key) { // original is an object
+      subjectValue[scope.$key] = value
+    } else {
+      subjectValue.$set(scope.$index, value)
+    }
+    if (this.filters) {
+      var vm = this._scope || this.vm
+      var reverseValue = vm._applyFilters(subjectValue,
+        subjectValue, this.filters, true)
+      vm.$set(this.expression, reverseValue)
+    }
   }
 }
 
--- a/src/directives/public/for.js
+++ b/src/directives/public/for.js
@@ -473,6 +473,7 @@ module.exports = {
    */
 
   _postProcess: function (value) {
+    this.processedValue = value
     if (_.isArray(value)) {
       return value
     } else if (_.isPlainObject(value)) {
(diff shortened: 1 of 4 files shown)


yyx990803:
Nice one. Since this is `v-for` specific, maybe we can move the part into `v-for` as a method, and just call it in the watcher.

karevn:
Agree. Btw, remember I offered to use `<v-for>` component instead of `v-for` directive? It would not only be a syntactic sugar, but would also allow to get rid of $forContext at all (it would all be about child - parent context relationships).

src/instance/misc.js
```diff
@@ -39,6 +39,14 @@ exports._applyFilters = function (value, oldValue, filters, write) {
   return value
 }
 
+exports._bubbleChanges = function (value, expression) {
```
yyx990803: I think splitting this here makes the code harder to follow. It's ok that this check stay in the watcher.

src/directives/public/for.js
```diff
@@ -513,6 +514,23 @@ module.exports = {
         frag.destroy()
       }
     }
+  },
+
+  _bubbleChanges: function (scope, value) {
```
yyx990803: Not sure about the method name here. There's no "bubbling" here - maybe something like `syncChanges`?

karevn: Yep, you're right.

src/instance/misc.js
karevn: It's more about responsibilities, IMO. Written this way, misc.js mixin is responsible for both forward and reverse filters, and abstracts forContext existence from the watcher.

Also, `alias === expression` does not look nice (because I easily bypassed this logic in my initial fiddle). Any ideas?

yyx990803: When I saw `scope._bubbleChanges` it's a bit confusing, because this is a v-for-scope-only thing, but is attached to Vue's prototype, which imo is the wrong place. It only works because of the prototypal inheritance.

yyx990803: As for expression check - a reverse write should be applied as long as the expression starts with the alias. e.g. `item.test` and `item[test]` both should trigger reverse write. So you need a regular expression, e.g. `new RegExp(alias + '\\b')`.

yyx990803:
Btw, if possible please re-submit this to the `dev` branch. Latest 1.0 work is now in the dev branch.

IMKAIER:
00

MicroDreamIT:
this.reverse=(this.sortKey==sortKey)?!this.reverse:false; is not working in latest version for example

<tr v-for="people in peoples | orderBy sortKey reverse">
		<td>{{ people.name}}</td>
		<td>{{ people.age}}</td>
	</tr>

	sortBy: function(sortKey){
			this.reverse=(this.sortKey==sortKey)?!this.reverse:false;
			this.sortKey=sortKey;

		},
	}

karevn:
@MicroDreamIT I think you're missing the point of this PR. It's intended to allow to re-apply changes done to the array processed with two-way filters. Sorting and reverse order sorting are irrelevant to this.

MicroDreamIT:
@karevn I am new in vuejs and javascript world, how can i make it happen?

karevn:
@MicroDreamIT please ask this question at Vue chat: https://gitter.im/vuejs/vue

PR #1450, opened 2015-10-20
v-model directive on <div contenteditable>

bdedardel:
v-model directive can only be used on <input>, <select> or <textarea> elements for now.
It would be nice to use it on 'contenteditable' !

Best regards,
Benjamin


yyx990803:
Hmm, did you mean to open an issue instead of a PR? This doesn't seem to contain anything and is merging the wrong branches.

PR #1497, opened 2015-10-27
fix `(k,v) in o` bugs with falsy values and updates

vprimachenko:


yyx990803:
Hmm, can you upgrade to 1.0 and submit the patch to the dev branch instead?

PR #1639, opened 2015-10-28
IE title binding issue

fergaldoyle:
With IE 10 & 11 (haven't tested on Edge yet), we cannot use title as a prop. Using `:title` or `v-bind:title`, IE seems to convert it to a static title attribute, which Vue will pick up instead of the bound attribute. Example:
http://codepen.io/anon/pen/epryMp

In this PR, a bound `title` attribute is preferred over a static title attribute. 

Perhaps all bound attributes should be preferred over static if both are specified on an element?

### DIFF
--- a/src/compiler/compile-props.js
+++ b/src/compiler/compile-props.js
@@ -21,7 +21,7 @@ module.exports = function compileProps (el, propOptions) {
   var props = []
   var names = Object.keys(propOptions)
   var i = names.length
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks, can you add a test case in `test/unit/specs/misc_spec.js`?

fergaldoyle:
Done

yyx990803:
Thanks!

PR #1647, opened 2015-10-29
fix dynamic attribute on custom directive

kazupon:
dynamic attribute don't work on custom directive.

please see the following jsfiddle:
https://jsfiddle.net/kazupon/of17drkk/6/


yyx990803:
Good job!

PR #1657, opened 2015-10-30
v-for: able to access parent scope's $refs and $els in v-for context.

TerenceZ:
As the title says.


TerenceZ:
Oh~The test cases are wrong...

PR #1658, opened 2015-10-30
v-for: support access for parent scope's $refs and $els in for context.

TerenceZ:
Allow the scope created by `v-for` to access its parent scope's $refs and $els.


yyx990803:
:+1:

PR #1694, opened 2015-11-04
v-on: add $target like $event

ccqgithub:
v-on: Let the eventHander can find the real DOM where binded. 
now：the event.target is the DOM where happend, can not find the DOM where binded……

    <span class="tit" @click="toggleLi($event, $target)">
        <i class="i fa fa-bar-chart"></i>
        <i class="arrow fa fa-angle-right"></i>
        菜单一
     </span>

### DIFF
--- a/src/directives/public/on.js
+++ b/src/directives/public/on.js
@@ -95,8 +95,10 @@ module.exports = {
 
     this.reset()
     var scope = this._scope || this.vm
+    var self = this
     this.handler = function (e) {
       scope.$event = e
+      scope.$target = self.el
       var res = handler(e)
       scope.$event = null
       return res


yyx990803:
Thanks, but there's the native `event.currentTarget` for that purpose...

ccqgithub:
@yyx990803 But in my test, `event.currentTarget` is not correct, some times is null, some times is TextNode...

yyx990803:
看不懂你什么意思，还是用中文吧...

ccqgithub:
@yyx990803 
非常不好意思，都怪我英文太差了。。。看见都用英文不好意思用中文~~~

你说的是对的，是因为传递给事件处理函数的event对象不知道被什么修改过了（事件执行之后被修改的，函数中alert阻断时event.currentTarget是正确的），所以我在控制台打印的event对象看到的时候其实是被修改过的，才让我误解不正确……

PR #1699, opened 2015-11-04
Add offset facility to limit filter aka mysql limit

Mat-Moo:
The limit filter is too limited by the fact that it's always a 0 offset, this commit allows a second parameter defining the offset allowing for onscreen pagination etc.

### DIFF
--- a/src/filters/array-filters.js
+++ b/src/filters/array-filters.js
@@ -9,10 +9,8 @@ var toArray = require('../directives/public/for')._postProcess
  * @param {Number} offset (Decimal expected)
  */
 
-exports.limitBy = function (arr, n, offset) {
-  var off = (!offset)
-    ? 0
-    : parseInt(offset, 10)
+exports.limitBy = function (arr, n, off) {
+  off = off ? parseInt(off, 10) : 0
   return typeof n === 'number'
     ? arr.slice(off, off + n)
     : arr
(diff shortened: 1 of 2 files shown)


src/filters/array-filters.js
```diff
@@ -6,11 +6,15 @@ var toArray = require('../directives/public/for')._postProcess
  * Limit filter for arrays
  *
  * @param {Number} n
+ * @param {Number} offset (Decimal expected)
  */
 
-exports.limitBy = function (arr, n) {
+exports.limitBy = function (arr, n, offset) {
+  var off = (!offset)
```
yyx990803: Can we just do `offset = offset ? parseInt(offset, 10) : 0` ?

Mat-Moo: Good point I'll update it

src/filters/array-filters.js
```diff
@@ -6,11 +6,13 @@ var toArray = require('../directives/public/for')._postProcess
  * Limit filter for arrays
  *
  * @param {Number} n
+ * @param {Number} offset (Decimal expected)
  */
 
-exports.limitBy = function (arr, n) {
+exports.limitBy = function (arr, n, off) {
```
yyx990803: one more thing: the comment above is `offset`. Let's be consistent and use `offset` in the code too. LGTM after that.

PR #1703, opened 2015-11-04
IE should prefer bound props over static props other than class

fergaldoyle:
A follow on from: https://github.com/vuejs/vue/pull/1639

There are more attributes other than `title` which IE has issues with. (`id`, `lang`, `language`,`dir`, etc)
e.g: http://codepen.io/anon/pen/yYxBPy

In this PR, we feature detect binding syntax issues, if the current browser does have this issue, we check every prop for a bound alternative, (`:`, `v-bind:`, `*.sync`, `*.once`) and if present, use the bound prop over a static prop.

`class` is ignored. Buy ignoring it the behavior matches other browsers (bound classes get passed down to component root element)

### DIFF
--- a/src/compiler/compile-props.js
+++ b/src/compiler/compile-props.js
@@ -8,6 +8,15 @@ var empty = {}
 var identRE = require('../parsers/path').identRE
 var settablePathRE = /^[A-Za-z_$][\w$]*(\.[A-Za-z_$][\w$]*|\[[^\[\]]+\])*$/
 
+// feature detect browsers (IE) that have trouble
+// with binding syntax on certain attributes
+var div, preferBinding
+if (_.inBrowser) {
+  div = document.createElement('div')
+  div.setAttribute(':title', '')
+  preferBinding = div.getAttribute('title') !== null
+}
+
 /**
  * Compile props on a root element and return
  * a props link function.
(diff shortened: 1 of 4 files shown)


src/compiler/compile-props.js
```diff
@@ -8,6 +8,15 @@ var empty = {}
 var identRE = require('../parsers/path').identRE
 var settablePathRE = /^[A-Za-z_$][\w$]*(\.[A-Za-z_$][\w$]*|\[[^\[\]]+\])*$/
 
+// feature detect browsers (IE) that have trouble
+// with binding syntax on certain attributes
+var div, preferBinding
```
yyx990803: Let's move this check in to `util/env.js` and expose it as `_.preferBinding`.

src/compiler/compile-props.js
```diff
@@ -120,6 +125,28 @@ module.exports = function compileProps (el, propOptions) {
...
+ * @return {String} attr
+ */
+
+function checkBindingAttr (el, attr) {
+  if (attr === 'class') {
+    return null
+  }
+
+  return (
+    el.getAttribute(':' + attr) ||
```
yyx990803: Looks like we can just use `el.hasAttribute` here and the returned value will be a boolean.

src/compiler/compile-props.js
```diff
@@ -120,6 +125,28 @@ module.exports = function compileProps (el, propOptions) {
...
+    el.getAttribute(':' + attr + '.once') ||
+    el.getAttribute(':' + attr + '.sync') ||
```
yyx990803: Is the check for `.once` and `.sync` necessary? Does IE also treat `:title.once` as `title`?

fergaldoyle:
`hasAttribute` now used and indeed there is no need to check for `.once` or `.sync`

PR #1762, opened 2015-11-08
Added support for properties which were created with defineProperty o…

jdavidberger:
…utside of vue.

Previously objects with properties with get/set defined outside of vue wouldn't propagate value changes after being made reactive in vue. This uses getOwnPropertyDescriptor to detect that case and account for it. 

Since it seems like every browser with supports defineProperty also supports getOwnPropertyDescriptor, I didn't add any logic to detect/work without it. 

Also added are two test cases testing the concept. 

If this addition gets pulled, it'd also probably be worthwhile checking the 'configurable' property. If configurable is set to false, defineReactive currently just throws an error. I imagine the correct behavior would be just to return out without changing anything; but this patch doesn't address it.

### DIFF
--- a/src/observer/index.js
+++ b/src/observer/index.js
@@ -182,8 +182,8 @@ function defineReactive (obj, key, val) {
     var property = Object.getOwnPropertyDescriptor(obj, key)
     if (property && property.get && property.set) {
       Object.defineProperty(target, 'val', {
-        get: property.get.bind(obj),
-        set: property.set.bind(obj)
+        get: _.bind(property.get, obj),
+        set: _.bind(property.set, obj)
       })
     }
   }
@@ -193,14 +193,15 @@ function defineReactive (obj, key, val) {
     enumerable: true,
     configurable: true,
     get: function metaGetter () {
+      var val = target.val
       if (Dep.target) {
         dep.depend()
         if (childOb) {
           childOb.dep.depend()
         }
-        if (_.isArray(target.val)) {
-          for (var e, i = 0, l = target.val.length; i < l; i++) {
-            e = target.val[i]
+        if (_.isArray(val)) {
+          for (var e, i = 0, l = val.length; i < l; i++) {
+            e = val[i]
             e && e.__ob__ && e.__ob__.dep.depend()
           }
         }
--- a/src/observer/index.js
+++ b/src/observer/index.js
@@ -172,7 +172,11 @@ function copyAugment (target, src, keys) {
 
 function defineReactive (obj, key, val) {
   var dep = new Dep()
-  var childOb = Observer.create(val)
+
+  var property = Object.getOwnPropertyDescriptor(obj, key)
+  var getter = (property && property.get) || function () { return val }
+  var setter = (property && property.set) || function (v) { val = v }
+  var childOb = Observer.create(getter())
   Object.defineProperty(obj, key, {
     enumerable: true,
     configurable: true,
@@ -182,18 +186,18 @@ function defineReactive (obj, key, val) {
         if (childOb) {
           childOb.dep.depend()
         }
-        if (_.isArray(val)) {
-          for (var e, i = 0, l = val.length; i < l; i++) {
-            e = val[i]
+        if (_.isArray(getter())) {
+          for (var e, i = 0, l = getter().length; i < l; i++) {
+            e = getter()[i]
             e && e.__ob__ && e.__ob__.dep.depend()
           }
         }
       }
-      return val
+      return getter()
     },
     set: function metaSetter (newVal) {
-      if (newVal === val) return
-      val = newVal
+      if (newVal === getter()) return
+      setter(newVal)
       childOb = Observer.create(newVal)
       dep.notify()
     }
--- a/src/observer/index.js
+++ b/src/observer/index.js
@@ -173,6 +173,8 @@ function copyAugment (target, src, keys) {
 
 function defineReactive (obj, key, val) {
   var dep = new Dep()
+  var hasGetter = true
+  var hasSetter = true
 
   var target = {
     val: val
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks for the PR!

I've considered implementing this before, but decided not to because of 2 reasons:

1. As stated, Vue expects your "data" to be plain objects so that they can be serialized and re-hydrated. Also, being plain objects make them easy to reason about. Allowing the data to contain arbitrary external getter will make component state less predictable. This is why Vue provides computed properties but intentionally limit them to the view layer. I don't really see any particularly convincing use case for data objects with pre-defined getter/setters. Most of them can be achieved using alternative patterns. 

2. Having to check `getOwnPropertyDescriptor` for every single property actually introduces a roughly 10% performance hit even if the user doesn't use this feature at all. This is the primary reason I did not implement this in Vue in the first place.

So unfortunately, I'll have to decline this - but let me know if you think there's a convincing use case that justifies the performance cost. In general, it's a good idea to discuss new features before submitting a PR. But I really appreciate your effort and please keep contributing in the future!

jdavidberger:
I'm not sure if the use case I'm targeting it for is going to be particularly compelling, but here is what I was using for it. I'm interfacing it with a C library ported into emscripten; the object properties on the JS side invoke getters and setters exposed from emscripten to set attributes in C; which are eventually used to calculate stuff in the C library.

At a 10% performance hit I completely understand not including it. If that wasn't the case, would you consider inclusion? I was curious about the number because I hadn't tested timing benchmarks at all under the assumption that getOwnPropertyDescriptor should be more or less free if you are only doing it once per defineProperty. Did you see the 10% hit on object creation or on object access? I wrote up some quick benchmarks and didn't see a hit in chrome, node, or firefox, but I only tested the latest versions.

yyx990803:
The 10% number was a brief test using one of the re-render benchmarks: https://github.com/vuejs/js-repaint-perfs The cost mostly comes at when replacing large nested objects and we need to convert these new objects to become reactive. This is probably not the most common use case though, so the overall perf hit would be less obvious.

In fact, it maybe a good idea to include this feature under a flag: it's off by default, but can be turned on when you need it and are ok with the perf tradeoff. For example, `Vue.config.convertAllProperties = true`.

src/observer/index.js
```diff
@@ -182,18 +186,18 @@ function defineReactive (obj, key, val) {
         if (childOb) {
           childOb.dep.depend()
         }
-        if (_.isArray(val)) {
-          for (var e, i = 0, l = val.length; i < l; i++) {
-            e = val[i]
+        if (_.isArray(getter())) {
```
yyx990803: Two things here:

1. The getter and setter should be invoked using `obj` as the `this` context, so `getter.call(obj)`.

2. The getter only needs to be called once inside the reactive getter. Calling it every time seems unnecessary.

jdavidberger:
Apologies if the above comments aren't formatted correctly; I've never actually used the in-diff commenting in github. 

This should address the key points; namely putting the functionality behind an option which defaults to off and minimizing impact to normal attributes when the option is turned on. I switched out the getter/setter function approach for one that, for the common case of normal attributes, doesn't have to call any functions which should perform better on some engines.

src/observer/index.js
```diff
@@ -172,7 +173,22 @@ function copyAugment (target, src, keys) {
...
+
+  var target = {
+    val: val
+  }
+
+  if (config.convertAllProperties) {
+    var property = Object.getOwnPropertyDescriptor(obj, key)
+    if (property && property.get && property.set) {
+      Object.defineProperty(target, 'val', {
+        get: property.get.bind(obj),
```
yyx990803: Since people may use Vue in PhantomJS, I've intentionally avoided using the native `bind`. You can use the `bind` provided by Vue's own util: `_.bind(get, obj)` and remove the bind shim in tests.

src/observer/index.js
```diff
@@ -182,18 +198,18 @@ function defineReactive (obj, key, val) {
         if (childOb) {
           childOb.dep.depend()
         }
-        if (_.isArray(val)) {
-          for (var e, i = 0, l = val.length; i < l; i++) {
-            e = val[i]
+        if (_.isArray(target.val)) {
```
yyx990803: Since `val` is now a getter, if the original is also a getter, we are still effectively invoking the original getter multiple times here. We should just cache the value as `var value = target.val` at the top of the metaGetter.

src/observer/index.js
```diff
@@ -172,11 +173,27 @@ function copyAugment (target, src, keys) {
...
   var dep = new Dep()
-  var childOb = Observer.create(val)
+
+  var target = {
+    val: val
+  }
+
+  if (config.convertAllProperties) {
+    var property = Object.getOwnPropertyDescriptor(obj, key)
+    if (property && property.get && property.set) {
```
yyx990803: Just noticed it's possible for a property to only have getters, or only have setters. We should handle all three possible situations here. Probably need to procedurally build up the descriptor based on the presence of the getter and the setter.

src/observer/index.js
```diff
@@ -189,11 +206,11 @@ function defineReactive (obj, key, val) {
           }
         }
       }
-      return val
+      return target.val
```
yyx990803: Should be fine to just return `val` here as well.

src/observer/index.js
yyx990803: Please bear with me, thanks for the effort so far :)

jdavidberger: No worries it's all good feedback. 

I think the getter/setter stuff makes sense, but might as well also make it safely handle 'configurable' being false. For now I'll make it so it just leaves everything alone; let me know if you think some other behavior makes more sense.

src/observer/index.js
jdavidberger: Not sure how I missed replacing this one out; I'll add a test condition which enforces only calling the underlying getter once.

yyx990803:
Thanks!

jdavidberger:
No problem! Thanks for reconsidering on the feature inclusion!

I did find a problem with the merged commit. 'property' is undefined when defineReactive is actually adding a field that wasn't there before to obj. I'm sure this happens in other places, but I first saw it in the for directive. I have a commit which tests and addresses this in my fork. It didn't flag in the unit tests since that config option is false by default. 

I reran the unit tests with the flag switched on by default and everything passed, so I think it should be good now. I'll be playing with it more in my project to see how it goes; would you rather I make pull requests if/when each thing comes up or group them up over a few days?

yyx990803:
Thanks for the follow up - note that I've made some modifications to the merged parts (mostly trying to further simplify it): https://github.com/vuejs/vue/blob/dev/src/observer/index.js#L174-L222

Ideally rebase your fork against the latest dev before submitting the new PR :)

PR #1765, opened 2015-11-08
Occurred TypeError in Observer.create

kazupon:
I create the object with `Object.create(null)` for plugin, and when it would try to define as reactivity object with `Vue.util.defineReactive`, occured the `TypeError` in `Observer.create`.

Please see the following jsfiddle:
https://jsfiddle.net/kazupon/yo2cmsLp/1/

In Vue.js, Can I define the reactity object was created with `Object.create(null)` ?

If this is bug, I hope that you will check PR.

### DIFF
--- a/src/observer/index.js
+++ b/src/observer/index.js
@@ -47,7 +47,7 @@ Observer.create = function (value, vm) {
   }
   var ob
   if (
-    '__ob__' in value &&
+    Object.prototype.hasOwnProperty.call(value, '__ob__') &&
     value.__ob__ instanceof Observer
   ) {
     ob = value.__ob__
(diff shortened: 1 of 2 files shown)


src/observer/index.js
```diff
@@ -47,7 +47,7 @@ Observer.create = function (value, vm) {
   }
   var ob
   if (
-    value.hasOwnProperty('__ob__') &&
+    '__ob__' in value &&
```
yyx990803: Thanks, but using `in` operator this would return `true` if the `__ob__` is defined on the prototype as well. We can use `Object.prototype.hasOwnProperty.call(value, '__ob__')`.

kazupon: Thanks!
You're right.

kazupon:
I fixed. :smile_cat:

yyx990803:
Thanks!

PR #1809, opened 2015-11-15
Occurred error in 'set' and 'delete' utility functions

kazupon:
we create the object with `Object.create(null)`, and after that when we use `Vue.util.set` or `Vue.util.delete`, occur the following error at these functions.

```
TypeError: obj.hasOwnProperty is not a function. (In 'obj.hasOwnProperty(key)', 'obj.hasOwnProperty' is undefined)
```

Please see the following jsfiddle:
https://jsfiddle.net/kazupon/1b7nytcd/1/

This issue is related #1765

I fixed the this issue, and refactored.
Please check the PR!! :smile_cat:

### DIFF
--- a/src/util/lang.js
+++ b/src/util/lang.js
@@ -10,7 +10,7 @@
  */
 
 exports.set = function set (obj, key, val) {
-  if (Object.prototype.hasOwnProperty.call(obj, key)) {
+  if (exports.hasOwnProperty(obj, key)) {
     obj[key] = val
     return
   }
@@ -43,7 +43,7 @@ exports.set = function set (obj, key, val) {
  */
 
 exports.delete = function (obj, key) {
-  if (!Object.prototype.hasOwnProperty.call(obj, key)) {
+  if (!exports.hasOwnProperty(obj, key)) {
     return
   }
   delete obj[key]
--- a/src/util/lang.js
+++ b/src/util/lang.js
@@ -10,7 +10,7 @@
  */
 
 exports.set = function set (obj, key, val) {
-  if (exports.hasOwnProperty(obj, key)) {
+  if (exports.hasOwn(obj, key)) {
     obj[key] = val
     return
   }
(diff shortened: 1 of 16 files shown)


src/util/lang.js
```diff
@@ -63,6 +63,17 @@ exports.delete = function (obj, key) {
...
 
 /**
+ * Check whether the object has the property.
+ *
+ * @param {Object} obj
+ * @param {String} key
+ * @return {Boolean}
+ */
+exports.hasOwnProperty = function (obj, key) {
+  return Object.prototype.hasOwnProperty.call(obj, key)
```
yyx990803: Let's cache `Object.prototype.hasOwnProperty` outside the function because this is going to be called a lot.

yyx990803:
Thanks! A few notes:

1. The changes to `v-model` directives are unnecessary because those are Elements so they will always inherit from `Object.prototype`.

2. Let's rename it to `_.hasOwn` (shorter and smaller file size)

src/util/lang.js
kazupon: OK !!
Should I use the `Cache` ?
https://github.com/vuejs/vue/blob/dev/src/cache.js

kazupon:
Thank you for your many review!! :)

I'll try :muscle:

src/util/lang.js
yyx990803: No, what I mean is just hold it in a variable outside the function as `var hasOwn = Object.prototype.hasOwnProperty`
It probably doesn't make much difference, but saves two property lookups on every call ;)

kazupon: I see. :)

kazupon:
Done.
Please check :)

yyx990803:
Looks good! Can you squash them into one commit please?

kazupon:
See the https://github.com/vuejs/vue/pull/1816 :)

PR #1818, opened 2015-11-16
Reverse filters support for v1.0

karevn:
Continuation of this work: https://github.com/vuejs/vue/pull/1441

### DIFF
--- a/src/directives/public/for.js
+++ b/src/directives/public/for.js
@@ -485,6 +485,7 @@ module.exports = {
    */
 
   _postProcess: function (value) {
+    this.processedValue = value
     if (_.isArray(value)) {
       return value
     } else if (_.isPlainObject(value)) {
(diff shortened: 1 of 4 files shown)


yyx990803:
:+1: thanks!

yyx990803:
Unfortunately, had to revert this because this actually neglects a use case where the user attempts to use `v-model` on a `v-for` alias with a read-only filter, e.g. `orderBy`. This is in fact pretty common and is what the original warning is trying to prevent.

With this PR, such use case no longer receives a warning and actually writes the value back to the original array at the wrong index.

karevn:
Ok. Any ideas for enhancements?
16 нояб. 2015 г. 21:50 пользователь "Evan You" <notifications@github.com>
написал:

yyx990803:
I tried working on top of your PR but haven't had a clean solution yet. Will take a deeper look later this week.

karevn:
I think we should figure out use cases first. :)
16 нояб. 2015 г. 22:05 пользователь "Evan You" <notifications@github.com>
написал:

PR #1826, opened 2015-11-18
add orderByKeys function

huangyifu:


yyx990803:
Thanks for the PR.

Please note that Vue.js core intentionally limits the functionality it provides to keep its size small. Therefore we only provide the most commonly-used filter functions as default ones. If the users need a highly custom sorting algorithm, they can just implement it as a custom filter or a computed property.

The basic idea is, we want avoid shipping code that adds size but only get used in a few very specific scenarios. So unfortunately I don't think we should add yet another sorting filter.

---

Additionally, please make sure to follow the code style and pass the lint/test build flow before submitting PRs.

PR #1832, opened 2015-11-19
make sure VS Code understand ES6 syntax

cnchanning:


yyx990803:
Why do I need this in Vue's source code if I am not using VS Code? Does this serve any user-land purpose?

fergaldoyle:
In user-land people should add `jsconfig.json` to the root of their project.

This would only benefit contributors to Vue who happen to use VS Code.

yyx990803:
I'm afraid this should be left to the user/contributor then. This seems simple enough to add it ad-hoc and it doesn't make sense to add config files for every possible IDE out there.

Issue #1945, opened 2015-12-01
Need more variables in inline statement


rhyzx:
~~Currently only `$event` variables is [provided](https://github.com/vuejs/vue/blob/d716ef46e781a4d65c71b2a1e94f6902ec14d7b4/src/directive.js#L232), and it represents the first argument this handler being called(DOM events in most case). But it is not suitable for custom vm events, especially with multi arguments.~~

eg.
```js
// parent
<my-component v-for="item in items" @my-event="handle($index, /* i want 'bar' here */)"></my-component>

// child
this.$emit('my-event', 'foo', 'bar')
```

I suggest add a more flexible variable(such as `$` or `$$`) to represents all arguments, so we can get any argument by `@my-event="handle($[0], $[1])"`.

For convenient, we can provide `$0` - `$9` to represents corresponding arguments like [Bash](http://tldp.org/LDP/abs/html/othertypesv.html#POSPARAMREF1).

@yyx990803 , if you agree with me and think this is necessary, please let me pull a PR for this :trollface:

rhyzx:
Well, after a researching of src codes, i found there is no variable(even `$event`) [provided](https://github.com/vuejs/vue/blob/dev/src/instance/api/data.js#L25) for custom events.

yyx990803:
Thanks, I think something like `$arguments` is more explicit. What we need to do is basically wrap the `handler` here: https://github.com/vuejs/vue/blob/dev/src/instance/internal/events.js#L41 - before calling the handler, we set `$arguments` on the current scope, and reset it afterwards.

rhyzx:
got it

Rolanddoda:
Does `$arguments` still exists in the template ? Bc I am trying to access it but it doesn't work. I am using a library which returns 2 arguments .

If I do:
```
<lib-component
    @some-event="$emit('some-event', $event)"
>
</lib-component
```

It catches only the first argument. So somehow I need to catch all events and seemed that $arguments is the best solution.

Rolanddoda:
Oh Ok I found that $arguments do not exists but `arguments` without $ sign exists.

PR #1949, opened 2015-12-02
Add $arguments variable in component events inline statement #1945

rhyzx:


src/instance/internal/events.js
```diff
@@ -32,13 +32,18 @@ export default function (Vue) {
...
-    var name, handler
     for (var i = 0, l = attrs.length; i < l; i++) {
-      name = attrs[i].name
+      let name = attrs[i].name
       if (eventRE.test(name)) {
         name = name.replace(eventRE, '')
-        handler = (vm._scope || vm._context).$eval(attrs[i].value, true)
-        vm.$on(name.replace(eventRE), handler)
+        const scope = vm._scope || vm._context
+        const handler = scope.$eval(attrs[i].value, true)
```
rhyzx: Does it ok to use ES6 syntax here? i searched whole src codes and found there is no ES6 syntax used except modules.

src/instance/internal/events.js
```diff
@@ -32,13 +32,18 @@ export default function (Vue) {
...
+        vm.$on(name.replace(eventRE), function (...args) {
+          scope.$arguments = args
+          const result = handler.apply(this, args)
```
rhyzx: One disadvantage of doing wrapper here instead of [here](https://github.com/vuejs/vue/blob/dev/src/instance/api/data.js#L26) is, **if this statement `isSimplePath`, it is no necessary to wrap it**

src/instance/internal/events.js
yyx990803: it's ok to use `let` and `const`. I am only using features that do not increase compiled file size at the moment.

src/instance/internal/events.js
yyx990803: Good point. Let's wrap it in `$get`. Also, use `util.toArray` to convert `arguments` to an array, the `...args` syntax adds a long for loop in the compiled code.

rhyzx: got it

yyx990803:
:+1:

PR #1955, opened 2015-12-03
add coerce option for component props

lepture:
Add a force convert option for component. Here is an example:

```html
<!-- Component Avatar -->
<script>
export default {
  props: {
    size: {
      type: Number,
      coerce: 'number',
      default: 48,
    }
  }
}
</script>
```

And then I can use:

```html
<avatar size="38"></avatar>
```

Otherwise, size will always be String, it won't pass the validation.

### DIFF
--- a/src/directives/internal/prop.js
+++ b/src/directives/internal/prop.js
@@ -5,7 +5,7 @@
 
 import Watcher from '../../watcher'
 import config from '../../config'
-import { assertProp, initProp } from '../../util/index'
+import { assertProp, initProp, coerceProp } from '../../util/index'
 
 const bindingModes = config._propBindingModes
 
@@ -25,6 +25,7 @@ export default {
       parent,
       parentKey,
       function (val) {
+        val = coerceProp(prop, val)
         if (assertProp(prop, val)) {
           child[childKey] = val
         }
--- a/src/util/component.js
+++ b/src/util/component.js
@@ -76,6 +76,7 @@ function getIsBinding (el) {
 
 export function initProp (vm, prop, value) {
   const key = prop.path
+  value = coerceProp(prop, value)
   vm[key] = vm._data[key] = assertProp(prop, value)
     ? value
     : undefined
@@ -141,6 +142,23 @@ export function assertProp (prop, value) {
     }
   }
   return true
+}
+
+/**
+ * Force parsing value with coerce option.
+ *
+ * @param {*} value
+ * @param {Object} options
+ * @return {*}
+ */
+
+export function coerceProp (prop, value) {
+  var coerce = prop.options.coerce
+  if (!coerce) {
+    return value
+  }
+  // coerce is a function
+  return coerce(value)
 }
 
 function formatType (val) {
--- a/test/unit/specs/directives/internal/prop_spec.js
+++ b/test/unit/specs/directives/internal/prop_spec.js
@@ -295,7 +295,7 @@ describe('prop', function () {
 
   describe('assertions', function () {
 
-    function makeInstance (value, type, validator) {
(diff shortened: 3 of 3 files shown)


lepture:
I will add tests and documentation after your acceptance.

src/compiler/compile-props.js
```diff
@@ -220,3 +220,31 @@ function getDefault (vm, options) {
...
+ *
+ * @param {*} value
+ * @param {Object} options
+ * @return {*}
+ */
+
+function getCoercedValue (value, options) {
+  var coerce = options.coerce
+  if (coerce) {
+    return value
```
Pumpuli: So if `options.coerce` exists the function does nothing...?

yyx990803:
1. Instead of specifying coerce type separately, how about just `coerce: true`? Since the type is declared already. For example:

``` js
{
  type: Number,
  coerce: true
}
```

Will simply coerce the value into number.

2. In addition to initialization, the coercion should also happen when prop is updated: https://github.com/vuejs/vue/blob/dev/src/directives/internal/prop.js#L28-L30

lepture:
@yyx990803 coerce can be a function.

src/compiler/compile-props.js
lepture: if (! coerce)

lepture:
@yyx990803 Maybe it is better to coerce value in assertProp function.

yyx990803:
@lepture yeah, coerce in `assertProp` sounds better.

For the value of the option, `number` and `int` sounds pretty arbitrary for me. How about `boolean` and `string`? It seems the following would be more reasonable:

1. If `coerce` is a function, use that function to coerce the value;
2. In other cases, `coerce` must be used together when `type` is also present, and it will coerce the value to the type specified. (there are internal util functions for these)
3. Special needs like `int` should just use a custom coerce function.

lepture:
How about coerce accepts only function by now? We could implement other values later.

BTW, Boolean, String, Number are valid functions too.

lepture:
Actually it should be in initProp

yyx990803:
initProp doesn't affect when the prop is updated from the parent though

lepture:
Yes. But it should be in initProp, since we have to assign the coerced value to vm. I'll parse the value in update bind.

yyx990803:
We can extract `coerceProp` as a function into util, so that it can be used in both places like `assertProp` and `initProp`.

lepture:
@yyx990803 updated. Should I export `coerceProp`?

yyx990803:
Yeah, we should export it, and then call it before `assertProp` here: https://github.com/vuejs/vue/blob/dev/src/directives/internal/prop.js#L28

lepture:
@yyx990803 updated. It should be documented and tested now.

src/directives/internal/prop.js
```diff
@@ -25,6 +25,7 @@ export default {
       parent,
       parentKey,
       function (val) {
+        val = coerceProp(val)
```
yyx990803: should be `coerceProp(prop, val)`

lepture: OMG.

yyx990803:
Cool it passes now, can you add a test case for it?

lepture:
@yyx990803 updated with tests.

yyx990803:
:+1:  Thanks!

bekite:
Is it possible to access the vue instance inside the coerce function.
I have defined some helper methods I want to use.

```
    someObj: {
      coerce (val) {
        // return this.$helper.copy(val) <-- would be nice
        return JSON.parse(JSON.stringify(val))
      }
    },

PR #1982, opened 2015-12-07
Dev

xueyaodeai:
window.ontransitionend and window.onanimationend now is null in Chrome and Safari instead of undefined!


yyx990803:
Thanks, the logic is only for detecting old browsers that still require the webkit prefix. It works as intended...

xueyaodeai:
I get it. I'm just learning the code of vuejs. Could you tell me some way to learn it?

posva:
The guide is neaaat http://vuejs.org/guide/

PR #2014, opened 2015-12-11
Fix package module ID / Cannot find module "vue"

tilleps:
Getting errors while trying to install via npm:
```
npm install vue
```

Was able to install the github version via:
```
npm install vuejs/vue#dev
```

However, I got the following error:
```
Error: Cannot find module 'vue'
```

Looks like the package.json main is pointing to a non-existing file:
`dist/vue.common.js`

I checked both `dev` and `master` branch, neither had the file.  This PR is removing
`dist/vue.common.js` from files and updating the package main to point to `dist.vue.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,8 @@
     "browser",
     "framework"
   ],
-  "main": "dist/vue.common.js",
+  "main": "dist/vue.js",
   "files": [
-    "dist/vue.common.js",
     "dist/vue.js",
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks, `vue.common.js` is only built and distributed with the npm module.

See updated instructions on using latest source code: http://localhost:4000/guide/installation.html#Dev_Build

PR #2043, opened 2015-12-17
make Vue work on node env that forbid `new Function` (ie. Atom)

jeremyfa:
_Vue_ is excellent at making UI in a simple and expressive way. I wanted to use it to create UI inside an Atom plugin, but Atom environment doesn't allow to use `eval` or `new Function()`.

This pull request provides an alternate implementation of `Function` based on node.js `vm` module when the native `Function` doesn't work. These changes make _Vue_ work fine on Atom (and similar node.js based environment that may have forbidden `new Function()`)

I ran browserify and webpack over the modified project, so far it doesn't break.

EDIT: I saw the csp-compliant build, which works on Atom as well, but when available, the `vm` based implementation may be more efficient than `notevil.Function`.


jeremyfa:
Mmmh, seems that I don't respect the indentation. I will fix that.

jeremyfa:
Well, still having issues. I will look into what I have broken -_-. Stay tuned.

jeremyfa:
On my side the modified `vue.js` still works just fine in the browser. Not sure if we should worry about the failed tests, that are skipped in the `dev` branch. Should we?

azamat-sharapov:
I might have misunderstood purpose of this PR, but.. have you checked **csp** branch? https://github.com/vuejs/vue/releases/tag/v1.0.11-csp

jeremyfa:
Yes thanks! Just edited the first post of this PR: I did see the *csp* branch, but the `vm` based implementation, when available (node.js/atom) may be more efficient than parsing AST with pure javascript. I am wondering if it could be considered to use it in that case?

yyx990803:
Thanks for this PR, but I don't think this should be part of the main distributed build of Vue.

On the other hand, the advantage of CSP build is that it can be used in environments that enforces CSP but do not have a Node runtime (e.g. Chrome apps).

It may indeed be the most efficient solution in Electron-based environments like Atom. But that's a really narrow use case. I think this could either be used as a fork or a special branch like `vm`.

jeremyfa:
Yes, makes sense.

The `vm` branch might be a good option (also to be able to do something like `npm install vue@vm`).

I guess for now I will use a fork.

Thanks for the feedback!

PR #2054, opened 2015-12-18
Fix "lazy" attribute for input type="text"

PavelCSS:


yyx990803:
I don't think this PR does what it says - it also breaks existing behavior. Please make sure your PR is correct.

PavelCSS:
v-model attribute lazy is activated twice when blur and change. I think enough to a change event

PR #2061, opened 2015-12-21
[fix] Convert string to number

pespantelis:
I found this issue by a [question](http://stackoverflow.com/questions/34388349/vue-js-binding-radio-elements-of-same-name-to-array) on StackOverflow.

If the `index` is string, the `this.length` assigned with the concatenated value of `index` and `1`.


ludo237:
:+1:

Issue #1960, opened 2015-12-03
class="{{ foo }" and :class="foo" not treated the same way during transitions.


crswll:
In this example we're using `class="{{ foo }}"` in the template but the elements aren't transitioning.

http://jsfiddle.net/crswll/t5jv7d98/1/

If we use `:class="foo"` they transition just fine.

http://jsfiddle.net/crswll/t5jv7d98/2/

PR #2076, opened 2015-12-23
Fix #2075

### LINKED ISSUE #2075: class="className {{ className }}" cause duplicated result
source:

``` html
<div class="icon icon-{{test}}"></div>
```

expect:

``` html
<div class="icon icon-test"></div>
```

result:

``` html
<div class="icon icon- icon-test"></div>
```

here's the code:

https://jsfiddle.net/10ons60y/

 是我的使用姿势不对吗  _(:3」∠)_

pespantelis:


yyx990803:
Thanks, this change was introduced in https://github.com/vuejs/vue/commit/ad2ac5070c830ecc7d892229d7ddf29a63e2a6c4 to fix #1960 - I will have to think about how to best handle it without re-introducing the issue in #1960.

PR #2102, opened 2016-01-01
Fix inline statement propagation

rhyzx:
When handle events by inline statement, propagation continuing is not working.

By the way: https://github.com/rhyzx/vue/blob/fix-inline-statement-propagation/test/unit/specs/api/events_spec.js#L186-L187 ,  binding handler using `v-on` directive to component will be treat as its inner handler, this is **non-intuitive** in my opinion, eg:

```js
// parent.vue
<foo @test="stopPropagation"></foo>
```

`test` event dispatched by `foo` will still propagate even if `stopPropagation` returned `false`, what do you think about this? @yyx990803

### DIFF
--- a/src/instance/api/data.js
+++ b/src/instance/api/data.js
@@ -24,8 +24,9 @@ export default function (Vue) {
         var self = this
         return function statementHandler () {
           self.$arguments = toArray(arguments)
-          res.get.call(self, self)
+          var result = res.get.call(self, self)
           self.$arguments = null
+          return result
         }
       } else {
         try {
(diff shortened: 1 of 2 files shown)


yyx990803:
After some investigation I'm starting to think that:

1. inline `v-on` should not interfere with event flow;
2. inline `v-on` should only be triggered by events emitted from that component only (i.e. not by dispatched or broadcasted events).

yyx990803:
... but these will be breaking changes. So I'll probably just make inline handlers behave as expected.

rhyzx:
:+1: rapid!

rhyzx:
Agreed, those can be introduced in next major release.

PR #2137, opened 2016-01-10
allow empty expression in terminal directives

rhyzx:
Eg: `<div v-if>`, i think this is a bug, expression validation should be in directive itself.

ps. Do i need to add test for this?


yyx990803:
Thanks, this should be fine without tests.

Issue #2161, opened 2016-01-14
__vfrag__ references prevent GC of Vue objects


elsassph:
Unless the host DOM element is fully removed (from DOM and from references), `__vfrag__` references seem to prevent full garbage collection of the Vue objects.

Using Chrome Profiler, after the `vm` object is destroyed we can still find references through `__frag__` between the DOM and the Vue objects:
```javascript
var data = {
    items:[
        { title:'Item 1', position:[0, 0] },
        { title:'Item 2', position:[200, 0] },
        { title:'Item 3', position:[0, 100] },
        { title:'Item 4', position:[200, 100] }
    ]
};

Vue.component('tile', {
    props: ["item"],
    template:`
        <div class="tile" :style="position">
            <b>{{item.title}}</b>
        </div>`,
    computed: {
        position: function() {
            var pos = this.item.position;
            return `left:${pos[0]}px; top:${pos[1]}px`;
        }
    }
});

var app = new Vue({
    el: '#app',
    template:`
        <div class="grid">
            <tile v-for="item in items" :item="item"/>
        </div>`,
    data: data
});

setTimeout(function() {
    app.$destroy();
    app = null;
    console.info('destroyed');
}, 2000);
```

PR #2162, opened 2016-01-14
Fixes: __vfrag__ references prevent GC of Vue objects 

elsassph:
See issue #2161


src/fragment/fragment.js
```diff
@@ -181,6 +181,9 @@ Fragment.prototype.destroy = function () {
   if (this.parentFrag) {
     this.parentFrag.childFrags.$remove(this)
   }
+  if (this.node) {
```
yyx990803: `this.node` will always be present, we can remove the `if` check here

PR #2165, opened 2016-01-15
Fix test about assigning to frozen object in chrome.

Gerhut:
When assigning property to a frozen object, an error is occurred in my chrome 47.0.2526.106

### DIFF
--- a/test/unit/specs/misc_spec.js
+++ b/test/unit/specs/misc_spec.js
@@ -223,7 +223,13 @@ describe('Misc', function () {
     })
     expect(vm.$el.textContent).toBe('hi frozen')
     vm.msg = 'ho'
-    vm.frozen.msg = 'changed'
+    try {
+      vm.frozen.msg = 'changed'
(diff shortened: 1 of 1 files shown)


yyx990803:
Interesting... it works on my Chrome with the same version. Are you using Windows?

Gerhut:
Sorry for forgetting notice that I am using Windows 7. Screenshot as follows.

After looking into [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Converting_mistakes_into_errors), writting to properties of frozen object will throw TypeError in strict mode, and silently fails without strict mode.

Here is a [demo](http://codepen.io/Gerhut/pen/QyqeKB?editors=001), my browser shows

    done
    boom in strict.

So maybe checking your test environment is in strict mode or not will helps?

image

yyx990803:
Ah, ok. Thanks for the explanation.

PR #2166, opened 2016-01-15
Feature: warn when mounting to body.

Gerhut:
[BrowserSync](http://browsersync.io/) sometimes failed when mounting vue instance to `document.body`. Although I have not recognized where the specific cause is, it recovered when mounting to a child element. Whatever, I think warning about this issue is a good idea. Moreover, React [warned about it](https://github.com/facebook/react/blob/3b96650e39ddda5ba49245713ef16dbc52d25e9e/src/renderers/dom/client/ReactMount.js#L398-L406) as well (Discussion: facebook/react#3207)

### DIFF
--- a/src/instance/api/lifecycle.js
+++ b/src/instance/api/lifecycle.js
@@ -24,6 +24,14 @@ export default function (Vue) {
     if (!el) {
       el = document.createElement('div')
     }
+    if (process.env.NODE_ENV !== 'production') {
+      if (el.tagName && el.tagName.toUpperCase() === 'BODY') {
+        warn(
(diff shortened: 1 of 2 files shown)


yyx990803:
[Vue already warns if an instance is mounted to body with `replace: true`](https://github.com/vuejs/vue/blob/dev/src/compiler/transclude.js#L79-L85). Mounting without `replace: true` shouldn't cause problems because it leaves `document.body` untouched. I'd be more interested in figuring out what is actually causing BrowserSync to fail.

Gerhut:
OK, I will continue working on it.

PR #2182, opened 2016-01-18
stringToFragment not using cache when templateString != templateStrin…

tgeorgiev:
…g.trim()

The issue was that initially we check the cache by an un-trimmed key, while later on, before putting a value in the cache, we trim the key. This would lead to the fragment of a templateString that has spaces at the beginning or the end, not to be retrieved from the cache.

### DIFF
--- a/src/parsers/template.js
+++ b/src/parsers/template.js
@@ -108,10 +108,8 @@ function stringToFragment (templateString, raw) {
     var suffix = wrap[2]
     var node = document.createElement('div')
 
-    if (!raw) {
-      templateString = templateString.trim()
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks, one issue here is that we want to use the `raw` behavior for text-only templates. So we should probably assign the trimmed string to a separate variable so that we can use the original string when a text-only string is detected.

tgeorgiev:
Oh, I see what you mean. I will fix this and update the request shortly. Thanks!

yyx990803:
:+1:  thanks!

PR #2183, opened 2016-01-18
Cache fixes

tgeorgiev:
- Fixing cache polution when putting an item with the same key multiple times
- Making shift() to reduce the current size of the cache


yyx990803:
wow, didn't know the cache had bugs all along!

tgeorgiev:
yeah, it happens so that I have this other issue that I have templates with whitespace in the beginning that started putting a lot of items in the cache, this is how I found it :)

PR #2200, opened 2016-01-20
improve nextTick by setImmediate

Jinjiang:


yyx990803:
We should still prioritize the MutationObserver approach, because it uses **microtask** semantics, and `setImmediate` uses **macrotask** semantics. In environments where both are present (e.g. IE11), we should prioritize the one that is consistent with other browsers where only MutationObserver is present.

For microtask vs macrotask, see https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

smolinari:
Sorry to bump in a bit offtopic Evan ( @yyx990803 ), but what is a macrotask? The article only mentions tasks and microtasks (and it is a great article btw). 

Scott

Jinjiang:
@yyx990803 gotcha. So mutation observer, promise belongs to microtask and setImmediate, setTimeout, process.nextTick belongs to (macro)task?

And Q1: mutation observer / promise >> (process.nextTick in Node.js > ) setImmediate > setTimeout(0) / process.nextTick polyfill in Browser. Is that correct?
Q2: why a DOM related solution (mutation observer) not promise?

smolinari:
Ahh...ok. A regular task is a macrotask. Right?

Scott

yyx990803:
Q1: yes except that process.nextTick is also microtask in Node > 0.9

Q2: because MutationObserver is more widely supported than native Promise,
and the user may be using a Promise polyfill that doesn't adhere to the
microtask semantics.

Jinjiang:
Well, so this mr will be merged, closed, later or what else? (I have updated the order: MutationObserver, setImmediate, setTimeout)

PR #2211, opened 2016-01-22
Support backspace with "delete" keycode alias

thecrypticace:
Currently when using the `delete` alias in a on directive with keyup `delete" does not support using backspace. This seems quite counterintuitive. Additionally, on Mac keyboards the key normally identified as "backspace" on PC keyboards is called "delete." I was fairly surprised when using the delete modifier didn't detect the key.

I came upon this when developing a search / tag field that turned a search query into tokens when hitting enter or space. I wanted a way to delete tokens without using the mouse. The solution I came up with was detecting the delete key and removing the last token if the search field was empty. I spent 10 minutes debugging code until I realized that I had to press fn+delete (equivalent of DEL / forward delete on a Mac laptop-sized keyboard) for it to actually remove the token.

You can see a reduced test case / demo of this here: http://codepen.io/thecrypticace/pen/YwGKmJ

This PR makes a change to allow aliases to map to multiple key codes. And, using that, it adds support for delete to react to both backspace and DEL / forward delete.

### DIFF
--- a/src/directives/public/on.js
+++ b/src/directives/public/on.js
@@ -7,7 +7,7 @@ const keyCodes = {
   tab: 9,
   enter: 13,
   space: 32,
-  'delete': 46,
+  'delete': [8, 46],
   up: 38,
   left: 37,
   right: 39,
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks for the detailed explanation and test coverage :+1:

thecrypticace:
Glad to help.

PR #2213, opened 2016-01-23
removed unnecessary methods _initMeta()

Jinjiang:
Latest version has used `defineReactive` directly in `for` directive

(And how to deal with the test case? No better idea...)


yyx990803:
I kept it there because it is used in vue-router. It's a bit weird if upgrading a minor version of Vue requires upgrading vue-router as well, so I'll keep this PR open until 1.1.

For the tests we can simply remove them.

yyx990803:
We'll just keep this in v1 for compatibility. v2 already got rid of it.

PR #2234, opened 2016-01-26
fix(util/env): wrong parameter type in document.createTextNode.

Brooooooklyn:
It wouldn't cause any issues, but not substandard.

https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode


yyx990803:
Thanks for the PR. I believe all DOM APIs automatically apply JavaScript type coercions when DOMString is expected, so when the underlying native method is called, the value has already been coerced. Otherwise, we'd need to do manual type coercion for all calls to `setAttribute()`, `Text.data = xxx` etc. as well, and I don't think this is necessary.

Issue #2130, opened 2016-01-08
transfer DOM


rhyzx:
We may need DOM place in some other location in some situations, eg:  place `<modal>` in root element(body).

`position: fixed` is not a perfect solution, when inside parent with z-index, it may be overlap by other elements: http://stackoverflow.com/a/16756277

I made some experiments and it seems impossible to achieve this by custom directive, so i made some modifications in source code and it works as expected:
http://jsfiddle.net/rhyzx/s3L0emxy/

commit is here: https://github.com/rhyzx/vue/commit/8430843342beed71953da8de00c87dd45cbb250b

How do you think about this feature? @yyx990803

yyx990803:
Thanks for the suggestion! This is definitely useful, however my concern is:

1. The rendered DOM structure would be different from the template structure;
2. The use case is a bit narrow for a new core directive: it is only necessary when we don't have control over the `z-index` of our app element on the page.

rhyzx:
> 1. The rendered DOM structure would be different from the template structure;
> 2. The use case is a bit narrow for a new core directive: it is only necessary when we don't have control over the z-index of our app element on the page.

That's true. this is a edge case and seems no perfect way to solve, eg. bootstrap's solution: http://v4-alpha.getbootstrap.com/components/modal/#live-demo , by this way, modal's lifecycle is not inherit from parent and not context aware. ng's solution is little cumbersome: https://angular-ui.github.io/bootstrap/#/modal .

How about providing some ways(may be an API) to extending this feature?

yyx990803:
Yeah, it definitely should be extendable... I think the only thing preventing it is the internal `terminalDirectives` array, since both `util` and `FragmentFactory` are exposed already.

yyx990803:
So, with be0d1cd75b4d84ece2b97cac5701376070884ab2 you should have everything you need to implement `v-inject` as a third party directive:

- `Vue.FragmentFactory`
- `Vue.util`
- `Vue.compiler.terminalDirectives`

I'll close this now - feel free to share `v-inject` and add it to awesome-vue :)

rhyzx:
Great, thanks.

geraldlrh:
how vue2.0 to transfer dom like this?

PR #2272, opened 2016-01-31
Improve terminalable directive

kazupon:
There are times when we want to use the `arg` and `modifires` in custom terminal directive, and control the priority.

e.g.
```html
<input type="text" v-validate:feild1.input-off="['required']">
```

Recently, in the following related issues or PR, we became able to define the terminal directive.

- allow empty expression in terminal directives
https://github.com/vuejs/vue/pull/2137

- transfer DOM
https://github.com/vuejs/vue/issues/2130

However, we can not yet.

I fixed this issues.
I hope that this PR are merged with you.

Please check the codes.

### DIFF
--- a/src/compiler/compile.js
+++ b/src/compiler/compile.js
@@ -25,14 +25,9 @@ const dirAttrRE = /^v-([^:]+)(?:$|:(.*)$)/
 const modifierRE = /\.[^\.]+/g
 const transitionRE = /^(v-bind:|:)?transition$/
 
-// terminal directives
-export const terminalDirectives = [
-  'for',
-  'if'
-]
-
 // default directive priority
 const DEFAULT_PRIORITY = 1000
+const DEFAULT_TERMINAL_PRIORITY = 2000
 
 /**
  * Compile a template and return a reusable composite link
@@ -588,13 +583,27 @@ function checkTerminalDirectives (el, options) {
       return skip
     }
   }
-  var value, dirName
-  for (var i = 0, l = terminalDirectives.length; i < l; i++) {
-    dirName = terminalDirectives[i]
-    value = el.getAttribute('v-' + dirName)
-    if (value != null) {
-      return makeTerminalNodeLinkFn(el, dirName, value, options)
+
+  var attrs, attr, name, value, matched, dirName, arg, def, termDef
+  attrs = el.attributes
+  for (var i = 0, j = attrs.length; i < j; i++) {
+    attr = attrs[i]
+    if ((matched = attr.name.match(dirAttrRE))) {
+      def = resolveAsset(options, 'directives', matched[1])
+      if (def && def.terminal) {
+        if (!termDef || ((def.priority || DEFAULT_TERMINAL_PRIORITY) > termDef.priority)) {
+          termDef = def
+          name = attr.name
+          value = attr.value
+          dirName = matched[1]
+          arg = matched[2]
+        }
+      }
     }
+  }
+
+  if (termDef) {
+    return makeTerminalNodeLinkFn(el, dirName, value, options, termDef, name, arg)
   }
 }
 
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks for the PR! This one takes some time to review, but I'll do it as soon as I can :+1:

kazupon:
Thanks!! :smiley_cat:

test/unit/specs/compiler/compile_spec.js
```diff
@@ -248,6 +248,52 @@ describe('Compile', function () {
...
+    var defTerminal = { priority: Vue.options.directives.if.priority + 1 }
+    var options = _.mergeOptions(Vue.options, {
+      directives: { term: defTerminal }
+    })
+    el.innerHTML = '<div v-term:arg1.modifier1.modifier2="foo"></div>'
+    var linker = compile(el, options)
+    linker(vm, el)
+    expect(vm._bindDir.calls.count()).toBe(1)
+    var args = vm._bindDir.calls.argsFor(0)
+    expect(args[0].name).toBe('term')
```
blake-newman: Perhaps assign the args[0] to a variable.

kazupon:
ping @yyx990803

src/compiler/compile.js
```diff
@@ -588,12 +590,31 @@ function checkTerminalDirectives (el, options) {
...
     }
   }
-  var value, dirName
-  for (var i = 0, l = terminalDirectives.length; i < l; i++) {
-    dirName = terminalDirectives[i]
-    value = el.getAttribute('v-' + dirName)
-    if (value != null) {
-      return makeTerminalNodeLinkFn(el, dirName, value, options)
+
+  var terminals = terminalDirectives.map(function (terminal) {
```
yyx990803: These change seems pretty expensive since it's now map + sort + nested for loop for every single element compiled. Ideally we want to move as much work out of this function as possible.

yyx990803:
Here's what I think what should be the ideal API: the custom directive should just declare `terminal: true` instead of monkey patching the internal `terminalDirectives` array. We should also change `v-if` and `v-for` to use the same `terminal: true` contract.

With this, the `checkTerminalDirectives` function would largely be the same with `compileDirectives` - pseudo code:

```
let A = "current terminal directive found with highest priority"
- for each attribute on current element:
  - if it's a directive, and has `terminal: true`:
    - if it have higher priority than A:
      - assign it to A
- If A exists, return a terminalLinkFn for it.
```

Does this make sense? I may have overlooked something but I think this could make it more efficient and more consistent.

simplesmiler:
> custom directive should just declare `terminal: true`

Do want. This will allow to make things like [vue-transfer-dom](https://github.com/rhyzx/vue-transfer-dom) truly reusable (i.e. without `Vue.use(...)`).

blake-newman:
I totally agree with the approach not to monkey patch parts of the core API.  This clearly has a use case that makes sense to easily implement, with future extentions this will become ever more requested.

kazupon:
@yyx990803 
Thank you for your code review , new API proposal for terminal directive, and pseudo code.

When I'm implementing the patch of this PR, I felt that `terminalDirectives` array is monkey patch. I think that your API proposal make sense.

I'll try to challenge based on your feedback. :smile_cat:

kazupon:
@yyx990803 updated

src/compiler/compile.js
```diff
@@ -588,14 +590,27 @@ function checkTerminalDirectives (el, options) {
...
+          name = attr.name
+          value = attr.value
+          dirName = matched[1]
+        }
+      }
     }
   }
+
+  if (termDef) {
+    return makeTerminalNodeLinkFn(el, dirName, value, options, termDef, name)
```
yyx990803: we can already extract arg as `matched[2]` here, and can just use that and pass it to `makeTerminalNodeLinkFn` instead of using `argRE`.

kazupon:
@yyx990803 
Updated Done.

This PR is breaking change of custom terminal directives.
We need to write the document terminal directive feature.

yyx990803:
@kazupon yes. Custom terminal directives were not publicly documented though, so technically this is a new feature. One case I can think of is https://github.com/rhyzx/vue-transfer-dom may need to be updated to use the new API. /cc @rhyzx

yyx990803:
Merged! :tada: 
I think this is the first PR that makes non-trivial updates to the compiler mechanism, great job @kazupon !

kazupon:
@yyx990803 
Thanks !! 

As next step, I'll try to write the documentation.

PR #2290, opened 2016-02-03
Check parentNode property before remove el from DOM

VincentGarreau:
Fore some reason I had an issue with this method. I fixed it by adding this condition.

### DIFF
--- a/src/util/dom.js
+++ b/src/util/dom.js
@@ -122,7 +122,9 @@ export function after (el, target) {
  */
 
 export function remove (el) {
-  el.parentNode.removeChild(el)
+  if (el.parentNode) {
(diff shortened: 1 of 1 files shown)


azamat-sharapov:
are you using vue-router on your app (where you got that error)?

VincentGarreau:
I use vue-router, but it's not caused by transition hooks. It's when a v-for list is updated (from a computed source).

yyx990803:
It's better to provide a repro of the scenario where you saw this error, because it's more likely a bug somewhere else that led to this. Fixing it here is not the ideal solution.

VincentGarreau:
Right, i'll try to provide a repro. I close the the PR for now, thanks!

PR #2303, opened 2016-02-05
Add non-developer-friendly description to README.md

meowwwls:
As someone who is currently a web development student, I often find it difficult to quickly understand what certain libraries and frameworks set out to accomplish when reading the technical introductions, or to imagine a real-world use case for the technology.

I have written what I hope is a clear, easy to understand, basic introduction to what Vue.js does at its core, one that I hope will be useful to everyone, whether they are seasoned developers, new programmers, or non-technical visitors. The goal is to help expose Vue.js to a broader range of users who at first might otherwise be confused by the more technical, thorough overview.


yyx990803:
Hi, thanks for the PR, and I appreciate the effort in making it more non-dev friendly. However, I intentionally kept the README short and succinct so that people don't miss the points.

Detailed intro can be found on the Vue.js documentation website: http://vuejs.org/guide/overview.html , which provides a very similar explanation. I'd be happy to take a pull request that adds a link to these sections in README instead of pulling in full paragraphs. Thanks!

PR #2346, opened 2016-02-14
Pass a context to factory functions

mtpc:
Allow this use case:

```javascript
const component = Vue.extend({
  template: '<div><dynamic></dynamic></div>',
  data() {
    return {
      path: 'some/path';
    }
  },
  components: {
    dynamic: function(resolve) {
      // "this" would be undefined otherwise, preventing us from using the instance's data
      require([`${this.path}/dynamic_module`], resolve);
      // System.import(`${this.path}/dynamic_module`).then(resolve)
    }
  }
});
```

### DIFF
--- a/src/instance/internal/misc.js
+++ b/src/instance/internal/misc.js
@@ -76,7 +76,7 @@ export default function (Vue) {
       } else {
         factory.requested = true
         var cbs = factory.pendingCallbacks = [cb]
-        factory(function resolve (res) {
+        factory.call(this, function resolve (res) {
           if (isPlainObject(res)) {
             res = Vue.extend(res)
           }


yyx990803:
I don't think this would work though. The resolved component is cached on the factory itself, so you cannot achieve dynamic loading like this, unless you add some mechanism for disabling caching: https://github.com/vuejs/vue/pull/2346/files#diff-2cde61012c979d215472e5f6ce1f8070R72

mtpc:
Indeed, it breaks caching since the returned vue instance might be different every time.
Would something like 
```javascript
components: {
  dynamic: function(resolve) {
    require([`${this.path}/dynamic_module`], function(resolvedComp) {
      resolvedComp.isDynamic = true;
      resolve(resolvedComp);
    }
  }
}
```

and then [here](https://github.com/vuejs/vue/blob/dev/src/instance/internal/misc.js#L84)
```javascript
if (!res.options.isDynamic) {
  factory.resolved = res
}
```

works? This is just the first idea crossing my mind, but it kinda looks like a dirty hack.
Or maybe there's a better way to resolve dynamic components I'm not aware of?

yyx990803:
One thing you can do:

``` js
components: {
  dynamic: function factory (resolve) {
    require([`${this.path}/dynamic_module`], function(resolvedComp) {
      resolve(resolvedComp); // this call sets the cache
      factory.resolved = null; // disable the cache
    }
  }
}
```

yyx990803:
Ok, I'll merge this since it seems cache busting can be achieved without other changes.

PR #2366, opened 2016-02-19
Fix custom terminal directive with global mixin

rhyzx:
Repo: https://jsfiddle.net/rhyzx/qdyq4z7z/2/

This is because Vue will [replace](https://github.com/vuejs/vue/blob/dev/src/instance/api/global.js#L157) old options with new one instead of extending exist options when `mixin()`, so compiler will get an outdated reference of [directives definition object](https://github.com/vuejs/vue/blob/dev/src/compiler/compile.js#L630).

### DIFF
--- a/src/compiler/compile.js
+++ b/src/compiler/compile.js
@@ -627,7 +627,8 @@ function makeTerminalNodeLinkFn (el, dirName, value, options, def) {
     filters: parsed.filters,
     raw: value,
     // either an element directive, or if/for
-    def: def || publicDirectives[dirName]
+    // #2366 or custom terminal directive
+    def: def || resolveAsset(options, 'directives', dirName)
   }
   // check ref for v-for and router-view
   if (dirName === 'for' || dirName === 'router-view') {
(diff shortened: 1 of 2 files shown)


src/compiler/compile.js
```diff
@@ -627,7 +628,8 @@ function makeTerminalNodeLinkFn (el, dirName, value, options, def) {
     filters: parsed.filters,
     raw: value,
     // either an element directive, or if/for
-    def: def || publicDirectives[dirName]
+    // #2366 or custom terminal directive
+    def: def || publicDirectives[dirName] || Vue.options.directives[dirName]
```
yyx990803: Thanks, this should actually be `def || resolveAsset(options, 'directives', dirName)` so that:

1. We don't need to import `Vue` here (feels dirty)
2. We reuse the same asset resolution logic (auto warnings if no directive found)

rhyzx: Got it

rhyzx: Done

Issue #2382, opened 2016-02-23
Feature : v-show modifier for toggling visibility


weotch:
It would be nice to be able to do:

```html
<div v-show.visible="bool"></div>
```

... to toggle between `visibility:hidden` and `visibility:visible`. 

My use case is an element with a transition.  When the element transitions out, I'd prefer for the space it took up to not collapse.

vuchl:
You could create a [custom directive](http://vuejs.org/guide/custom-directive.html) for that and use the [instance properties](http://vuejs.org/guide/custom-directive.html#Directive_Instance_Properties) (this.modifiers) to get the `v-directive.modifier`.

The only thing i can not really make out right now is if you can override the `v-show`directive.

simplesmiler:
@weotch is `opacity: 0;` different to `visibility: hidden;` for your use case?

yyx990803:
It's probably better to just create your custom directive like `v-visible`. Unfortunately the internal transition API is not exposed at the moment, but it will be in the next release and you will be able to achieve what you want using `Vue.transition.applyTransition()`.

weotch:
@vuchl Thanks for the input, but I don't believe that solution would trigger transitions like normal `v-show` does.

@simplesmiler no, those would be equivalent for my use case.

@yyx990803 I don't think a custom directive would work because I wouldn't be able to communicate to the transition system to do an enter / leave.  But cool to know about that new API, thanks!

yyx990803:
@weotch if changing the opacity is what you want, then you can do it with a simple class toggle with CSS transitions instead of using `v-show`.

azamat-sharapov:
> Unfortunately the internal transition API is not exposed at the moment

Ha! this got me not long ago, but there is solution: `var apply = Vue.directive('show').apply`

canamo06:
@yyx990803 is there API of transitions in the new version? because I want to trigger the transition in my component or with a custom directive.

 I'm doing some calculations with the offsetHeight, so I need to use 'visibility' instead of 'display'(v-show)

bennyschudel:
how does this work in 2.0?

noudadrichem:
I had the same prob, here is the fix I found. Custom directive in the form of a Node module:
https://www.npmjs.com/package/vue-visible

m4heshd:
[...]

PR #2389, opened 2016-02-24
visiblity modifier for v-show

crswll:
I saw #2382 and figured I'd give contributing a shot. I'm almost certain I followed the guidelines but if I didn't let me know.

### DIFF
--- a/src/directives/public/show.js
+++ b/src/directives/public/show.js
@@ -12,20 +12,24 @@ export default {
   },
 
   update (value) {
-    this.apply(this.el, value)
+    this.apply(this.el, value, this.modifiers)
     if (this.elseEl) {
-      this.apply(this.elseEl, !value)
+      this.apply(this.elseEl, !value, this.modifiers)
     }
   },
 
-  apply (el, value) {
+  apply (el, value, modifiers) {
     if (inDoc(el)) {
       applyTransition(el, value ? 1 : -1, toggle, this.vm)
     } else {
       toggle()
     }
     function toggle () {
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks @crswll - sorry I didn't communicate this clearly, I didn't think adding a modifier to `v-show` was the best way to approach this. What he wanted can simply be solved by using a CSS class to toggle the element's opacity with CSS transitions.

simplesmiler:
I asked https://github.com/vuejs/vue/issues/2382#issuecomment-188131890 because `opacity: 0` and `visibility: hidden` do not produce identical results. Namely, content with `opacity: 0` is interactive (links and buttons can be clicked, inputs are focusable), while content with `visibility: hidden` is not.

`opacity: 0; pointer-events: none` produce similar result to `visibility: hidden`, but only works in IE11+.

Thus I believe a directive for toggling `visibility` instead of `display` is handy, either in core or in ecosystem.

crswll:
Ah I got you. I was just thinking in terms of getting the transitions right. Honestly I was just kind of bored and wanted to see if I can do it!

yyx990803:
@simplesmiler ah that's true. That's why I exposed the transition API though. With that you can easily write your own `v-visible` directive.

milesingrams:
I've been trying to write a v-visible directive that triggers transitions and I cant see any way to trigger enter/leave transitions without importing directly from 'vue/src/platforms/web/runtime/modules/transition'.
You mentioned exposing the transition API but I cant see any place to import it! Where might I find this exposed transition api?

Issue #2393, opened 2016-02-25
Vue tries to observe enumerable but non-configurable properties


saul:
### Vue.js version
1.0.16

### Reproduction Link
https://jsfiddle.net/5sH6A/275/
Check JavaScript console for error.

### Steps to reproduce
Add a non-configurable, but enumerable property to component data/props.

### What is expected?
Vue.js should *not* be redefining properties that are not configurable in `defineReactive` as by virtue of being non-configurable, they cannot change anyway so observing them is pointless (and also throws an error).

This bug currently makes it impossible to add Three.js objects to Vue components without causing this error. However a current workaround is to set `convertAllProperties = true`, as this triggers checks to ensure that non-configurable properties are not redefined.

### What is actually happening?
`Uncaught TypeError: Cannot redefine property: <property name>`

Mat-Moo:
FYI the fiddle is blank?

saul:
Thanks @Mat-Moo, updated the link

yyx990803:
It seems it's better to skip observing the entire object? I saw your other comment about a non-reactive object being converted when passed down as a prop. Maybe we need a way to explicitly declare non-reactive properties on a vm instance.

saul:
I agree that that would be a good feature, but think it should be separate to this fix, which fixes objects that you want to be reactive with some non-configurable properties.

I think the way to solve the 'non-reactive object' suggestion would be to never observe objects that are frozen (e.g., with `Object.freeze`), and can be tested with `Object.isFrozen`.

simplesmiler:
> Maybe we need a way to explicitly declare non-reactive properties on a vm instance.

Yes please.

yyx990803:
@saul makes sense, let's work off #2394 to fix this.

On the other hand, Vue already skips observing objects that are not [extensible](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isExtensible). But the user may have objects that should not be observed but should be mutable. So maybe a new option like `nonReactive`?

``` js
export default {
  // reactive data
  data () {
    return { a: 1 }
  },
  nonReactive () {
    return {
      vector: new THREE.Vector3( 1, 0, 0 )
    }
  }
})
```

Basically, this simply attaches the property to the instance without converting it - in addition, it should probably add a private marker to the object (if it's extensible) so that when the object is passed down as a prop, it doesn't get converted again.

Mat-Moo:
[...]

PR #2394, opened 2016-02-25
Test whether a property is configurable before defining it as reactive

saul:
Fixes vuejs/vue#2393

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -36,14 +36,6 @@ const config = {
   warnExpressionErrors: true,
 
   /**
-   * Whether or not to handle fully object properties which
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks, after some thinking, I think we can just use `convertAllProperties` behavior as the default and remove that config altogether.

saul:
Do you want me to make that change in this PR?

yyx990803:
Yes please.

saul:
Done :)

yyx990803:
Great job, thanks! :+1:

PR #2413, opened 2016-02-29
Added CoffeeScript TodoMVC example code

anfelor:
This is actually more a pull _proposal_ rather than a request.

I found CoffeeScript to work quite well with Vue.js
as there are lots of objects and anonymous functions.

Moreover it shortens the source code to 129 lines (instead of 168).

Everything works, so I would like to see it in the vue repository,
but as said, I understand that it isn't a must.


yyx990803:
Hey @anfelor , thanks for the proposal. In general, Vue does not assume language preference and I personally don't think a CoffeeScript example belongs in the core repo.

However, I think this would be helpful for people who use both CoffeeScript and Vue. I would suggest you make a standalone repo of this example, and add it to [awesome-vue](https://github.com/vuejs/awesome-vue#examples).

SusanthCom:
Thanks @yyx990803  for your great work (Vue) and the Suggestion (awesome-vue)

PR #2499, opened 2016-03-16
fix: directive I/F consistency

kazupon:


vuchl:
I'm no JS master so can someone explain why you got rid of the function keyword and what it does?

Is it just another syntax? I see the consistency with e.g. the `v-on`directive code.

blake-newman:
@vuchl This is es6 syntax,  it basically sugar syntax for functions in objects.

PR #2508, opened 2016-03-18
Allow currency filter to specify places

jbrooksuk:
This adds another parameter to the `currency` filter which allows specifying of the places to display. This is useful if you only want to display a currency as £100.

### DIFF
--- a/src/filters/index.js
+++ b/src/filters/index.js
@@ -63,20 +63,23 @@ export default {
   /**
    * 12345 => $12,345.00
    *
-   * @param {String} sign
+   * @param {String} currency
+   * @param {Number} places
    */
 
-  currency (value, currency) {
+  currency (value, currency, places) {
     value = parseFloat(value)
     if (!isFinite(value) || (!value && value !== 0)) return ''
     currency = currency != null ? currency : '$'
-    var stringified = Math.abs(value).toFixed(2)
-    var _int = stringified.slice(0, -3)
-    var i = _int.length % 3
+    places = places != null ? places : 2
+    var stringified = Math.abs(value).toFixed(places)
+    var dotplace = Math.abs(places + 1)
+    var _int = stringified.slice(0, -dotplace)
+    var i = _int.length % dotplace
     var head = i > 0
-      ? (_int.slice(0, i) + (_int.length > 3 ? ',' : ''))
+      ? (_int.slice(0, i) + (_int.length > dotplace ? ',' : ''))
       : ''
-    var _float = stringified.slice(-3)
+    var _float = stringified.slice(-dotplace)
     var sign = value < 0 ? '-' : ''
     return sign + currency + head +
       _int.slice(i).replace(digitsRE, '$1,') +
--- a/test/unit/specs/filters/filters_spec.js
+++ b/test/unit/specs/filters/filters_spec.js
@@ -77,6 +77,11 @@ describe('Filters', function () {
     expect(filter(-50)).toBe('-$50.00')
     expect(filter(-150.43)).toBe('-$150.43')
     expect(filter(-1500.4343434)).toBe('-$1,500.43')
+    // places
+    expect(filter(123, '$', 2)).toBe('$123.00')
+    expect(filter(123.45, '$', 0)).toBe('$123')
+    expect(filter(123.456, '$', 2)).toBe('$123.46')
+    expect(filter(123.456, '$', 3)).toBe('$123.456')
   })
 
   it('debounce', function (done) {


src/filters/index.js
```diff
@@ -63,20 +63,23 @@ export default {
   /**
    * 12345 => $12,345.00
    *
-   * @param {String} sign
+   * @param {String} currency
```
jbrooksuk: Pretty sure this is right.

src/filters/index.js
```diff
@@ -63,20 +63,23 @@ export default {
...
 
-  currency (value, currency) {
+  currency (value, currency, places) {
     value = parseFloat(value)
     if (!isFinite(value) || (!value && value !== 0)) return ''
     currency = currency != null ? currency : '$'
-    var stringified = Math.abs(value).toFixed(2)
-    var _int = stringified.slice(0, -3)
-    var i = _int.length % 3
+    places = places != null ? places : 2
```
posva: Instead of having a default value of 2, I think it's better to not use `toFixed` is no value is passed

What do you think?

src/filters/index.js
```diff
@@ -63,20 +63,23 @@ export default {
   /**
    * 12345 => $12,345.00
    *
-   * @param {String} sign
+   * @param {String} currency
+   * @param {Number} places
    */
 
-  currency (value, currency) {
+  currency (value, currency, places) {
```
posva: I'd rather call places decimals, but I'm not native English so I prefer someone to confirm or reject this

posva:
Where are my modals?! I jumped right into the code review 😳

Thanks for the PR. The jsdoc is indeed wrong

posva:
@jbrooksuk What is the use case for displaying 0, 1 or more than 2 decimals on a currency?

src/filters/index.js
chrisvfritz: @posva How about decimalPlaces?

posva: I think that dotPosition or dotPos is also a better name :smile:

jbrooksuk:
> @jbrooksuk What is the use case for displaying 0, 1 or more than 2 decimals on a currency?

@posva mainly it's just when you have a more precise number but need to display it differently.

src/filters/index.js
jbrooksuk: I don't think there is an issue with this name?

posva: There's not, I never heard about places :sweat_smile:, sorry
@chrisvfritz Places seems fine to me

posva:
@jbrooksuk I see the concern on arbitrary numbers but what about currencies?

jbrooksuk:
> @jbrooksuk I see the concern on arbitrary numbers but what about currencies?

When you don't want the trailing ".00" on the end,

chrisvfritz:
@posva @jbrooksuk What about instead of an option to specify decimal places, having an option to display whole numbers without decimals? So for example in USD, `22` would become `"$22"`, but `22.2` would become `$22.20`. That way, you can guarantee you're not hiding rounding errors or other miscalculations.

chrisvfritz:
Perhaps a boolean such as `preferIntegers`?

FadyMak:
@jbrooksuk @posva  I believe your concern is valid for removing trailing '.00'.

For inspiration: Angular provides a similar parameter called *fractionSize* on their currency filter which you can reference here: https://docs.angularjs.org/api/ng/filter/currency

@chrisvfritz In that case an optional flag to the currency filter such as *no_trailing_zeros* could be an option.

jbrooksuk:
> @posva @jbrooksuk What about instead of an option to specify decimal places, having an option to display whole numbers without decimals? So for example in USD, 22 would become "$22", but 22.2 would become "$22.20". That way, you can guarantee you're not hiding rounding errors or other miscalculations.

@chrisvfritz that doesn't work in some scenarios. For example, the app I'm building needs to display the whole price, not the pence, but elsewhere it needs the pence.

jbrooksuk:
Remember that this is an optional parameter and the default behaviour continues to be the same. I feel this adds more flexibility.

chrisvfritz:
@jbrooksuk So in some cases there is a non-zero pence in the price, but you still prefer to hide it?

jbrooksuk:
> @jbrooksuk So in some cases there is a non-zero pence in the price, but you still prefer to hide it?

Yeah, and without re-implementing the filter, it's not "impossible". Like I said, this is an *optional* extension of existing functionality. It's just not strict anymore.

chrisvfritz:
@FadyMak I like the `noTrailingZeroes` name! :smiley: 

@jbrooksuk Hmm, I worry that might be an edge case. I feel like suppressing non-zero pence/cents/etc could very easily (and unintentionally) hide errors in other applications, if people used it hoping to only suppress trailing zeroes.

jbrooksuk:
> @jbrooksuk Hmm, I worry that might be an edge case. I feel like suppressing non-zero pence/cents/etc could very easily (and unintentionally) hide errors in other applications, as people might use it hoping to only suppress trailing zeroes.

If they do, then that's a matter of not understanding what the filter is doing.

chrisvfritz:
@jbrooksuk I agree, but I think in designing APIs, it's important to be wary of releasing features that are especially _prone_ to misunderstanding. I also have a preference for explicitness though. Since that's a feature that could easily hide rounding/calculation errors, I would want to make sure I (and other team members) remained very aware of that. So I would personally never use such an option even if it existed. I would instead opt for explicitness - with something like a computed value called `priceWithoutPence` that I used the currency filter on instead.

What are your thoughts @posva @FadyMak?

FadyMak:
@jbrooksuk I agree with the point of being explicit. I believe the core API should provide basic functionality such as specifying the *decimalPlaces* as an **optional** parameter to the currency filter. This would result in the currency filter rounding to the specified number of decimal places.

Since the number of decimal places varies by locale (i.e.: the US Dollar ($) is usually represented with 2 decimal places whereas the Kuwaiti Dinar (KWD) is represented with 3 decimal places), the **decimalPlaces** parameter would default to the maximum number of decimal places for that specific locale.

Please reference the [ISO 4217 World Currency Format](http://www.thefinancials.com/?SubSectionID=curformat) to see the default number of decimal places a specific currency displays.

If a developer wishes to perform more complex formatting, they can achieve that by defining a custom filter of their own and chaining it to the currency filter.

jbrooksuk:
Defaulting based on locale could be dangerous?

chrisvfritz:
@FadyMak Good to know about some currencies using 3 decimal places instead of 2 or 0 - thanks for linking to that resource!

src/filters/index.js
blake-newman: I would use `precision` as the terminology is a bit more precise.  As you are defining the accuracy of the output.

blake-newman:
I would actually create a precision filter away from this to allow for more use cases,  and chain the filters.  This will help add more to the API at the same time as resolving this issue.

yyx990803:
@jbrooksuk thanks for the PR! So to sum it up:

1. We want to avoid making built-in filters more complex than they are right now, because Vue is not a utility library and domain-specific logic is best left in userland. The default filters were provided for convenience but not for completeness.

2. As @blake-newman suggested, chaining an extra filter seems like a cleaner and smarter solution.

Although we are not merging this, we really appreciate your contribution!

PR #2515, opened 2016-03-19
update internal explain comments

kazupon:


blake-newman:
Looks good to me.  Perhaps we can update the eslint config to include jsdoc to validate the comments. 

@yyx990803 yours thoughts?

yyx990803:
@blake-newman sounds like a good idea, although there might be a lot to fix since current jsdoc in Vue source may not be fully compliant ;)

blake-newman:
@yyx990803 always the case when adding more linting,  will look into it at somepoint this week.

PR #2521, opened 2016-03-21
Initialize underlying _data with prop value

### LINKED ISSUE #2520: $data replacement doesn't trigger change detection
### Vue.js version
1.0.18

### Reproduction Link
http://jsfiddle.net/6byd5w3q/3/

### Steps to reproduce
1. Create simple component `my-component`
2. Pass props object with single property `myProp`
3. Pass method `update$Data` which sets `this.$data = {myProp: 'new value'}`
4. Pass template which contains input field, which is binded to myProp and button which has click handler binded to `update$Data`
5. Execute `update$Data` by pressing on the button 

### What is Expected?
`myProp` should be updated 

`myProp` is not updated

### Extra Info
Works in 1.0.17 and bellow.

Possible cause: https://github.com/vuejs/vue/commit/d1788b98ac8dce4352e9f78810d13380449bd321#diff-b994c9d50c338882ce8ff1f2b6e6b5b5R107

oaleynik:
Closes #2520

### DIFF
--- a/src/util/component.js
+++ b/src/util/component.js
@@ -105,6 +105,7 @@ export function initProp (vm, prop, value) {
     value = getPropDefaultValue(vm, prop.options)
   }
   if (assertProp(prop, value)) {
+    vm._data[key] = value
     defineReactive(vm, key, value, true /* doNotObserve */)
   }
 }


yyx990803:
Thanks, please see https://github.com/vuejs/vue/issues/2520#issuecomment-199082346

Issue #2530, opened 2016-03-22
support "classnames" syntax in :class


luwenjin:
From knockout to angular to react, now i'm new to Vue. Really love this library, feels like "angular+react for humans".

:class now support array **OR** object, but sometimes it's still a little tedious
if :class support classnames(https://github.com/JedWatson/classnames) like syntax will be better.

example:
from
`:class="[type, checked?'checked':'', readOnly?'read-only':'', disabled?'disabled':'', fitted?'fitted':''"`
into
`:class="type, {checked, disabled, fitted}, readOnly&&'read-only'"`

Sure, I can always use a computed method, but by doing this class names are defined in one more place(props+template -> props+template+computed), which harms readability.

rpkilby:
This is not a real answer, but for readability in the short term you could break the parameter across multiple lines. eg,

```html
<div :class="{
 'type': true,
 'checked': checked,
 'read-only': readOnly,
 'disabled': disabled,
 'fitted': fitted}">
</div>
```

fullfs:
@luwenjin Hello! I'm also having problems with multiple conditioned classnames like in your example. But please note you're proposing to bring new directive syntax for a sole case. It does not look to like a wise decision to me, because @yyx990803 tried to make directives' syntax as compact as possible and make it look like common javascript with simple expressions, arrays and objects.

@rpkilby 's version looks better to me :)

simplesmiler:
@luwenjin `{ checked, disabled, fitted }` is ES6 and will only. Other than that you can make a custom method `cn: (classes) => classNames(classes)` and use that:
```
:class="cn(type, { checked, disabled, fitted, 'ready-only': readOnly })"
```
Or, with support for ES5 browsers:
```
:class="cn(type, { checked: checked, disabled: disabled, fitted: fitted, 'ready-only': readOnly })"
```

luwenjin:
@rpkilby in my example, the read-only class is a special case showing false value will be ignored. In common case, it will be like:
`:class="type, {checked, disabled, fitted, readonly}'"`
it's ES5 object literal vs ES6, if you can use ES6 style, you will almost never use the old style.

@fullfs I like the compact feeling of Vue. But just like v-for, IMHO, if one directive is used so frequently, it's worth breaking the rule in order to make the user wring code faster/clearer.
And in my proposal, it's still common javascript in some way, it's the spread arguments(like ...args in function(...args){}). 
Even if sticking to the list/object type, 
[...]

PR #2540, opened 2016-03-23
:class bindings with array + objects

FadyMak:
For feature request: #2530 

Feel free to suggest any improvements or a better way of going about the proposed solution.


src/directives/internal/class.js
```diff
@@ -38,7 +38,17 @@ export default {
   handleArray (value) {
     this.cleanup(value)
     for (var i = 0, l = value.length; i < l; i++) {
-      if (value[i]) {
+      if (value[i] && isPlainObject(value[i])) {
+        var keys = Object.keys(value[i])
+        for (var j = 0; j < keys.length; j++) {
```
yyx990803: this part has a lot of duplicate with `handleObject`. we should extract them into a common function.

src/directives/internal/class.js
```diff
@@ -50,8 +60,19 @@ export default {
       var i = this.prevKeys.length
       while (i--) {
         var key = this.prevKeys[i]
-        if (key && (!value || !contains(value, key))) {
+        if (key && (!value || !contains(value, key)) && !isPlainObject(key)) {
```
yyx990803: I'd refactor the logic to:

```
if (!key) continue
if (isPlainObject(key)) {
  // ...
} else if (!value || !contains(value, key)) {
  // ...
}
```

src/directives/internal/class.js
```diff
@@ -50,8 +60,19 @@ export default {
...
           removeClass(this.el, key)
+        } else if (key && isPlainObject(key)) {
+          var keys = Object.keys(key)
+          for (var k = 0; k < keys.length; k++) {
+            for (var j = 0; j < value.length; j++) {
```
yyx990803: `value` is not guaranteed to be an Array - it can be an Object too. I think it's better to improve `contains` so that it also checks objects inside arrays.

FadyMak:
Thanks a lot for the great pointers! I know the code is messy and far from perfect, I thought I'd get it in a PR and iterate over it.

FadyMak:
Refactored the code and added some more tests.

src/directives/internal/class.js
```diff
@@ -24,21 +24,16 @@ export default {
...
-    }
+    this.prevKeys = Object.keys(value)
+    setObjectClasses(this.el, value)
   },
 
   handleArray (value) {
     this.cleanup(value)
     for (var i = 0, l = value.length; i < l; i++) {
-      if (value[i]) {
+      if (value[i] && isPlainObject(value[i])) {
```
yyx990803: let's cache `value[i]` in a variable here

yyx990803:
Nice! Just a small thing (see line note) and LGTM.

FadyMak:
Good catch, I've updated it to cache `value[i]`. Once the PR is merged I'll updated the related docs.

src/directives/internal/class.js
```diff
@@ -69,7 +85,15 @@ function stringToObject (value) {
 }
 
 function contains (value, key) {
-  return isArray(value)
-    ? value.indexOf(key) > -1
-    : hasOwn(value, key)
+  if (isArray(value)) {
+    for (var i = 0; i < value.length; i++) {
+      if (isPlainObject(value[i])) {
+        return hasOwn(value[i], key)
```
yyx990803: returning here is wrong: it should be `if (hasOwn(value[i], key)) { return true }`

src/directives/internal/class.js
```diff
@@ -69,7 +85,15 @@ function stringToObject (value) {
...
+      } else {
+        return value[i] === key
```
yyx990803: same here: should be a conditional return true.

We can also cache `value[i]` here.

FadyMak:
Cached `value[i]` and fixed prematurely returning.

test/unit/specs/directives/internal/class_spec.js
```diff
@@ -53,5 +53,16 @@ describe(':class', function () {
     arr.push('f')
     dir.update(arr)
     expect(el.className).toBe('a f')
+    // test array with objects
+    dir.update(['x', {y: true, z: true}])
+    expect(el.className).toBe('a x y z')
+    dir.update(['x', {y: true, z: false}])
+    expect(el.className).toBe('a x y')
+    dir.update(['x', {z: true}])
```
yyx990803: The test was passing previously with the bug in `contains`, so we need a bit more variation in the tests here - e.g. switching the order of the items in the array / changing `'x'` to something else.

FadyMak:
Removed contains and all prevKeys are cleared out.

yyx990803:
Nice! LGTM now - can you squash into one commit?

FadyMak:
Squashed commits!

yyx990803:
woot!

FadyMak:
Thanks for all your help @yyx990803

Issue #2337, opened 2016-02-09
Allow component prop type to be an array of accepted types


saul:
At the minute, prop types can only be custom/native constructors. It would be helpful if the type could be an array of custom/native constructors, and if the passed in prop value matches any of these functions, it would be deemed valid.

For example, a prop with name `value` that can accept either an `Array` or `Number` as its type:
```js
props: {
	value: {
		type: [Array, Number],
		default: 5 // or: default: [1, 2, 3]
	}
}
```

ealves-pt:
@yyx990803 is this up for grabs? I want to give an hand. :smiley:

yyx990803:
@ealves-pt sure thing! Just make sure to include tests in your PR

ealves-pt:
Perfect. :+1:

ealves-pt:
Should I update the documentation to reflect this changes?

phanan:
@ealves-pt Sure, please do. 
@Everyone: Let's continue the discussion (if need be) in the PR. I'm closing this issue to keep the list clean.

PR #2545, opened 2016-03-23
feature: allow component prop type to be an array of types

ealves-pt:
Solves #2337

### DIFF
--- a/src/compiler/compile-props.js
+++ b/src/compiler/compile-props.js
@@ -278,36 +278,23 @@ export function assertProp (prop, value) {
   }
   var options = prop.options
   var type = options.type
-  var valid = true
-  var expectedType
+  var valid
+  var expectedTypes = []
   if (type) {
-    if (type === String) {
-      expectedType = 'string'
-      valid = typeof value === expectedType
-    } else if (type === Number) {
-      expectedType = 'number'
-      valid = typeof value === 'number'
-    } else if (type === Boolean) {
-      expectedType = 'boolean'
-      valid = typeof value === 'boolean'
-    } else if (type === Function) {
-      expectedType = 'function'
-      valid = typeof value === 'function'
-    } else if (type === Object) {
-      expectedType = 'object'
-      valid = isPlainObject(value)
-    } else if (type === Array) {
-      expectedType = 'array'
-      valid = isArray(value)
-    } else {
-      valid = value instanceof type
+    if (!isArray(type)) {
+      type = [ type ]
+    }
+    for (var i = 0; i < type.length && !valid; i++) {
+      var assertedType = assertType(value, type[i])
+      expectedTypes.push(assertedType.expectedType)
+      valid = assertedType.valid
     }
   }
-  if (!valid) {
+  if (typeof valid !== 'undefined' && !valid) {
     process.env.NODE_ENV !== 'production' && warn(
       'Invalid prop: type check failed for ' +
       prop.path + '="' + prop.raw + '".' +
-      ' Expected ' + formatType(expectedType) +
+      ' Expected ' + formatTypes(expectedTypes) +
       ', got ' + formatValue(value) + '.'
     )
     return false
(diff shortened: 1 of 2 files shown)


src/compiler/compile-props.js
```diff
@@ -344,6 +331,36 @@ export function coerceProp (prop, value) {
   return coerce(value)
 }
 
+function assertType (value, type) {
```
blake-newman: I think this function could be moved to a utilities file for reusability?

blake-newman:
Code looks good at this stage, however don't forget to add the tests.

src/compiler/compile-props.js
ealves-pt: Good idea. I'll probably move it to `src/util/component.js`

ealves-pt:
I'll be working on the test coverage today.

src/compiler/compile-props.js
yyx990803: Let's keep it here for now since it's not yet used in any other context. We can move it when we do need to use it elsewhere.

ealves-pt:
@blake-newman and @yyx990803 I think this PR is ready to be considered.

yyx990803:
Thanks for the great work!

src/compiler/compile-props.js
```diff
@@ -342,9 +329,46 @@ export function coerceProp (prop, value) {
...
+  }
+  return {
+    valid,
+    expectedType
+  }
+}
+
+function formatTypes (types) {
+  return types.reduce(
+    (p, c, i) => { return i === 0 ? formatType(c) : `${p}, ${formatType(c)}` },
```
saul: Can this be simplified to: `types.map(formatType).join(', ')`?

src/compiler/compile-props.js
```diff
@@ -278,36 +278,23 @@ export function assertProp (prop, value) {
...
+      type = [ type ]
+    }
+    for (var i = 0; i < type.length && !valid; i++) {
+      var assertedType = assertType(value, type[i])
+      expectedTypes.push(assertedType.expectedType)
+      valid = assertedType.valid
     }
   }
-  if (!valid) {
+  if (typeof valid !== 'undefined' && !valid) {
```
saul: condition could be replaced with `valid === false`

src/compiler/compile-props.js
```diff
@@ -278,36 +278,23 @@ export function assertProp (prop, value) {
   }
   var options = prop.options
   var type = options.type
-  var valid = true
-  var expectedType
+  var valid
```
saul: Can this not be initialised to `false`?

src/compiler/compile-props.js
ealves-pt: Yeah the code is more readable the way you say. @yyx990803 already did that in this [commit](https://github.com/vuejs/vue/commit/daab3c5f58c97bcb7c528ae199f0419456703e8f). You are right, perhaps using ` or ` instead of `, ` is a better approach.

src/compiler/compile-props.js
ealves-pt: I've just noticed that this was also changed with this [merge](https://github.com/vuejs/vue/commit/e831412f2a4fbf89f66c13aab0988976adc8ff72)

src/compiler/compile-props.js
fullfs: They say, `let` should be used instead of `var` in ES6.

src/compiler/compile-props.js
```diff
@@ -342,9 +329,46 @@ export function coerceProp (prop, value) {
   return coerce(value)
 }
 
-function formatType (val) {
-  return val
-    ? val.charAt(0).toUpperCase() + val.slice(1)
+function assertType (value, type) {
+  var valid
+  var expectedType
```
fullfs: By the way, checking the `constructor` or `constructor.name` is not reliable enough?

```js
let ok = 'test'
ok.constructor === String // true
```

hanai:
If there some default type such as "Array", "String" and check them with internal method would better?

``` js
function checkType(prop, type) {
  return Object.prototype.toString.call(prop) === ('[object ' + type +']');
}

type: ["Array", "Object"]
```

saul:
Is it possible to support `null` as a possible type? For example, if I want a prop to either be null or a string, I could set the type to `[String, null]`

ealves-pt:
@saul it doesn't support that.

This change maintains the current behaviour. To achieve that  you can create a custom validator.

yyx990803:
@saul as of 1.0.20 I believe `null` and `undefined` no longer raises type check warnings.

Issue #1469, opened 2015-10-25
beforeDetach hook


zenozeng:
I am now trying to write a list view component for Vue.
And I had a problem saving state of current items before detach.

I want to collect boundingClientRect of all list items on beforeDetach, so that I can restore the ListView when re-attach <list-view>.

So, I wonder, if there is a way to exec certain code right before detach, or is there any plan for it?

yyx990803:
Currently no, but it is easy to add. For now you can use a transition's `leave` hook...

zenozeng:
Thanks a lot.

fnlctrl:
+1

smolinari:
If I may ask, what is a list view component @zenozeng ?

Scott

zenozeng:
See https://github.com/CatchLabs/vue-list-view

smolinari:
Thanks @zenozeng. 

Scott

yyx990803:
Dropping this because in most cases you can do this before you manipulate the data instead.

For items with transitions, do it in the `beforeLeave` hook.

Also in 2.0, attach/detach hooks are being deprecated. If a component is being removed then it is being destroyed.

PR #2526, opened 2016-03-21
beforeDetach hook

rowanwins:
Hi,

I've made a start on implementing a fix for  #1469  . Im not convinced I've nailed it 100% but I thought I'd submit what I've got for feedback and direction.

I also need to prepare a test although I've not worked with jasmine or e2e so that'll take a little bit of digging but I'll attempt to get that up and running tomorrow evening.

Cheers
Rowan

PR #2547, opened 2016-03-23
Adding beforeAttach and beforeDetach event handlers including tests

rowanwins:
This is a followup pull request on #2526 which aims to implement the beforeAttach and beforeDetach features discussed in #1469 .

Let me know if anything further is required.


yyx990803:
Looking good! Will take one last pass later today to make sure everything is covered :+1:

test/unit/specs/api/lifecycle_spec.js
```diff
@@ -112,7 +112,7 @@ describe('Lifecycle API', function () {
     })
 
     it('hooks', function () {
-      var hooks = ['created', 'beforeCompile', 'compiled', 'attached', 'ready']
+      var hooks = ['created', 'beforeCompile', 'compiled', 'beforeAttach', 'attached', 'ready']
```
vuchl: Is `beforeDettach` missing here?

posva:
Can you please add to the commit description `Fix #1469` so it also closes the issue?

src/fragment/fragment.js
```diff
@@ -138,6 +145,7 @@ function multiRemove () {
   var self = this
   var shouldCallRemove = inDoc(this.node)
   this.beforeRemove()
+  self.callHook(beforeDetach)
```
yyx990803: should check `shouldCallRemove` here too

src/fragment/fragment.js
```diff
@@ -99,6 +102,7 @@ function singleRemove () {
   this.inserted = false
   var shouldCallRemove = inDoc(this.node)
   var self = this
+  self.callHook(beforeDetach)
```
yyx990803: need to check `shouldCallRemove`

test/unit/specs/api/lifecycle_spec.js
rowanwins: @vuchl it looked like this series of tests was for monitoring adding stuff to the DOM, rather the the removal aspect but I could be wrong...

rowanwins:
Gday @yyx990803 just checking that you saw I updated this as per your suggestion, hopefully it's close now :)

yyx990803:
@rowanwins it is - it's just this PR has a bit more moving pieces than others so I want to be extra careful. It will probably be part of 1.1 instead of the next patch release, so please be patient :)

rowanwins:
Gday @yyx990803 , so Im thinking this pull request may not be required if the [Vue 2.0 write up](https://github.com/vuejs/vue/wiki/2.0-features) is anything to go by... Anyway no biggie if that's the case, was good to poke around in the source and I'm looking forward to seeing what v2 brings :)

Cheers
Rowan

yyx990803:
Hey @rowanwins - apologies for letting this PR hang for so long.

In v2, `attached` and `detached` are indeed no longer supported. This is mainly due to their usefulness not justifying the amount of complexity in ensuring correct behavior for these hooks. Given that, `beforeAttach` and `beforeDetach` obviously are not needed as well.

For v1, it would be less than ideal to introduce two hooks that are soon to be gone in v2, so unfortunately that means we will probably have to let this PR go. I know you've put a lot of work into it, so I do want to thank you for that and say sorry ;'(

PR #2553, opened 2016-03-24
Proof of concept: Detect writing to non-reactive properties

skyronic:
This is very very early proof of concept right now. If a property in a data object is written to while it's not reactive, a warning is shown. This will later be used only in debug mode, but I haven't added a debug mode check yet. The code will be re-written cleanly once it's confirmed working.

Note: this uses no dirty checking and instead uses ES6 proxies, which is supported only on latest chrome and firefox. 

Need help with this:

* Pull my branch `reactive-warn` into your repo.
* edit `changewatch/index.html` with updates 
* find false positives, etc. Things when it should show warnings but doesn't. Or other edge cases.

Example: 
screen shot 2016-03-24 at 9 49 58 am

### DIFF
--- a/examples/changewatch/index.html
+++ b/examples/changewatch/index.html
@@ -1,0 +1,64 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Vue.js tree-view demo</title>
+  <script src="../../dist/vue.js"></script>
+</head>
+<body>
+  <div id="demo">
+    <div>
+      <button @click="doIt">Do It</button>
+    </div>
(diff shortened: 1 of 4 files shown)


yyx990803:
I like the idea - but one problem with proxies is that they become different references and would break cases where the user is doing equality comparisons...

skyronic:
Sigh :( you're right. i confirmed that reference equality isn't preserved.

On the bright side i did learn quite a bit about vue internals from this exercise. I really think this would be a great feature. I am currently out of ideas and am stashing this. Please do see if you can get a nice solution for this problem :) 

Closing PR.

Issue #2518, opened 2016-03-20
orderBy mutlitple parameters [fetaure request]


dcrystalj:
it would be really nice to have option for sorting like this ```user in users | orderBy 'name' 'lastName'```. if multiple users have same name they get sorted by lastname

posva:
Hi

Thanks for the feature request!
IMO the multi-field sorting is something usual and should be implemented. The impact on size is minimal and is not a breaking change
I'm implementing the feature asap tomorrow :smile:

rhyzx:
@posva @yyx990803 how about supporting both key and custom sort fn like `items | sortBy myComparator`, it should be more flexiable in my opinion :smiley:

posva:
@rhyzx  I'd like to but we need to add it as a last parameter in order to be safe from a breaking change :cry: 
We can also add it as an optional first parameter but it may become harder to understand (as a filter)

Here's what I've done so far: https://github.com/posva/vue/commit/3a55f3c1e571af3deaafd1368cff02cd4744b190
At the moment I'm breaking the grid e2e example and removing the ability to use the filter with null as a parameter, although imo there's no reason to call the filter manually with `null`
I'll try to fix it this evening

posva:
@rhyzx About the comparator function I think this should be another filter called `sort`. `orderBy` orders by keys while sort should be simplier. I'm not sure if it should be added to the main repo

dcrystalj:
Thank you for not ignoring this request and all the effort you have put into it! 
Passing custom sorting comparator like @rhyzx said makes sense as well. I hope both features will be included in the release one day.

posva:
@dcrystalj I appreciate you support :smile: . I've been a bit slow and I'm sorry, next time I'll handle a request faster :muscle: . As for the custom comparator I think this should be included in another filter named `sort` because making the `orderBy` filter more complex isn't a good idea. Here is the code for the sort filter: https://jsfiddle.net/posva/5ky7ec1t. By default it compares by string though as that is the default behaviour for `Array::sort`. The code have been done in about 5 minutes :sweat_smile: and it can be improved.

349989153:
It's very nice to have multiple fields to be sorted by the ``orderBy`` in the latest version.
Can we go further, I mean, multiple fields with multiple sort order like:
``user in users | orderBy 'name' -1 'lastName' 1``

Or maybe change the ``orderBy`` into ``orderBy chain``:
[...]

PR #2562, opened 2016-03-25
Multiple field order by

posva:
This is refers to the feature request #2518

I replaced the deconstructing array with a simple array because it makes it easier to use an array from data or computed properties: `files | orderBy sortKeys` instead of `files | orderBy 'name' 'size'` with `sortKeys = ['name', 'size']`. Tell me your thoughts about it

### DIFF
--- a/src/filters/array-filters.js
+++ b/src/filters/array-filters.js
@@ -74,18 +74,22 @@ export function filterBy (arr, search, delimiter) {
 /**
  * Filter filter for arrays
  *
- * @param {String} sortKey
+ * @param {String|Array<String>} sortKeys
  * @param {String} reverse
  */
 
-export function orderBy (arr, sortKey, reverse) {
+export function orderBy (arr, sortKeys, reverse) {
   arr = convertArray(arr)
-  if (!sortKey) {
+  let order = (reverse && reverse < 0) ? -1 : 1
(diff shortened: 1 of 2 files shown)


posva:
Ping @yyx990803

src/filters/array-filters.js
```diff
@@ -93,6 +97,16 @@ export function orderBy (arr, sortKey, reverse) {
...
+  }
+
+  function recursiveCompare (a, b, i) {
+    if (i === sortKeys.length - 1) return compare(a, b, i)
+    return compare(a, b, i) || recursiveCompare(a, b, i + 1)
+  }
+
+  // sort on a copy to avoid mutating original array
+  return arr.slice().sort(function (a, b) {
+    return recursiveCompare(a, b, 0)
```
yyx990803: we can jsut `sort(recursiveCompare)` if we do `i = i || 0` in it right?

posva:
@yyx990803 Thanks for the update, I didn't think about the unnecessary function call :smile: . I fixed that and also squashed everything into a single commit. I made a backup in case you prefer it the other way, just tell me if you want me to switch back to the 4 commits version :wink:

yyx990803:
:+1: good job!

PR #2581, opened 2016-03-29
Support for mutiple class names as key (close #2571)

### LINKED ISSUE #2571: [feature request] :class object support multiple class names as key
just like this
`:class="{'a b c': flag, d: flag2}"`
as well as this
`:class="[flag?'a b c':'']"`

phanan:
This commit adds support for using multiple class names as key when specifying a `:class` object, for example `<div :class="{ 'a b c': flag }">`. Previously, doing so will trigger an uncaught InvalidCharacterError, as spaces are not allowed in DOM tokens. Feature request #2571 should also be satisfied by this commit.

### DIFF
--- a/src/util/dom.js
+++ b/src/util/dom.js
@@ -206,6 +206,19 @@ export function setClass (el, cls) {
  */
 
 export function addClass (el, cls) {
+  // Space " " is not a valid character for a DOM token.
+  // As such, if cls includes multiple class names (e.g.
+  // "a b c"), we split it into mutiple child-tokens and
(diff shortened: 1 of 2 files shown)


yyx990803:
We also need to handle `removeClass` for the same. I think it's better to handle this inside `directives/internal/class.js` and add related test cases for the directive.

phanan:
Oh right. I'll look into that.

phanan:
Let me close this and use a proper, updated, branch instead.

PR #2585, opened 2016-03-29
Support for mutiple class names as key

phanan:
This is to replace #2581.

### DIFF
--- a/src/directives/internal/class.js
+++ b/src/directives/internal/class.js
@@ -41,19 +41,16 @@ export default {
   },
 
   cleanup (value) {
-    if (this.prevKeys) {
-      var i = this.prevKeys.length
-      while (i--) {
-        var key = this.prevKeys[i]
-        if (!key) continue
-        if (isPlainObject(key)) {
-          var keys = Object.keys(key)
-          for (var k = 0; k < keys.length; k++) {
-            removeClass(this.el, keys[k])
-          }
-        } else {
-          removeClass(this.el, key)
-        }
+    if (!this.prevKeys) return
+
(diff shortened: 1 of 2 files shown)


src/directives/internal/class.js
```diff
@@ -63,9 +63,11 @@ function setObjectClasses (el, obj) {
   var keys = Object.keys(obj)
   for (var i = 0, l = keys.length; i < l; i++) {
     var key = keys[i]
-    if (obj[key]) {
-      addClass(el, key)
-    }
+    if (!obj[key]) continue
+
+    key.split(' ').forEach(function (k) {
```
yyx990803: Unconditional split + forEach can be slow, and class bindings can be a hot spot so performance is important.

Let's do an `key.indexOf(' ') > 1` check first, then use a normal for loop.

phanan: That's what concerned me too, though I don't have an exact idea about the speed comparison between the check and a blind `split`.

src/directives/internal/class.js
```diff
@@ -77,3 +73,39 @@ function stringToObject (value) {
...
+
+/**
+ * Add or remove a class/classes on an element
+ *
+ * @param {Element} el
+ * @param {String} key The class name. This may or may not
+ *                     contain a space character, in such a
+ *                     case we'll deal with multiple class
+ *                     names at once.
+ * @param {Boolean} add
```
yyx990803: instead of passing a boolean, we can just pass in `addClass` or `removeClass` and call it directly.

phanan: Great idea. Will do.

PR #2591, opened 2016-03-30
orderBy can take a Function as its first argument

posva:
Hello,

This feature was asked in #2518. The point of this generic filter is to support any kind of sorting by using a custom comparator. Tell me your thoughts on the feature and the code :smile: 

PS: I did the PR instead of the issue because it took little to no time to do it and I didn't have internet when I coded it :stuck_out_tongue:

### DIFF
--- a/src/filters/array-filters.js
+++ b/src/filters/array-filters.js
@@ -74,16 +74,21 @@ export function filterBy (arr, search, delimiter) {
 /**
  * Filter filter for arrays
  *
- * @param {String|Array<String>} sortKeys
+ * @param {String|Array<String>|Function} sortKeys
  * @param {Number} [order]
  */
 
 export function orderBy (arr, sortKeys, order) {
   arr = convertArray(arr)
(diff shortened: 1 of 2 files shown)


src/filters/array-filters.js
```diff
@@ -110,6 +110,27 @@ export function orderBy (arr, sortKeys, reverse) {
 }
 
 /**
+ * Filter filter for arrays
+ *
+ * @param {Function} [compare]
```
fullfs: `@param {Array} [arr]` is missing

yyx990803:
Forgot to comment in the feature request issue, but I think this should just be part of `orderBy`. The first argument can either be a string or a function, this is how `filterBy` works currently.

posva:
Ok, so instead of giving a String or an Array of string you can give your own Function.

posva:
Let me review this a last time tomorrow in the morning (I want to be fresh :smile: )
I'll also update the doc

posva:
ping @yyx990803

Issue #2606, opened 2016-04-01
Component Props: coerce not working with default option in v1.0.20


JosephusPaye:
Hi there, first I would like to say thanks for Vue, it is truly awesome.

I think I may have found a slight regression in Vue 1.0.20. It seems like the `default` option for component props is not passed through the `coerce` function in Vue 1.0.20. I had this working in 1.0.17 before I upgraded. See the Fiddles below for details.

Thanks

### Vue.js version
1.0.20

### Reproduction Link
Not working Fiddle (Vue 1.0.20): http://jsfiddle.net/5sH6A/292/
Working Fiddle (Vue 1.0.17): http://jsfiddle.net/0w62a9xx/1/

### Steps to reproduce
See the Fiddles above.

### What is Expected?
The prop default should be passed through the coerce function.

### What is actually happening?
Only props that are specified in the template (and not defaults) are passed through the coerce function.

simplesmiler:
This looks like unintentional breaking change, but I don't think the behavior was ever documented.
As a workaround you can move default into the coerce function: http://jsfiddle.net/5sH6A/297/

I'm pinpointing a commit where the change has happened.

simplesmiler:
Fixed by https://github.com/vuejs/vue/pull/2610

simplesmiler:
Landed in 1.0.21.

PR #2610, opened 2016-04-01
Default a prop before coercion

simplesmiler:
Proposed fix for https://github.com/vuejs/vue/issues/2606

### DIFF
--- a/src/compiler/compile-props.js
+++ b/src/compiler/compile-props.js
@@ -229,10 +229,10 @@ function makePropsLinkFn (props) {
 
 function processPropValue (vm, prop, rawValue, fn) {
   const isSimple = prop.dynamic && isSimplePath(prop.parentPath)
-  let value = coerceProp(prop, rawValue)
-  if (value === undefined) {
-    value = getPropDefaultValue(vm, prop)
+  if (rawValue === undefined) {
+    rawValue = getPropDefaultValue(vm, prop)
   }
+  let value = coerceProp(prop, rawValue)
   const coerced = value !== rawValue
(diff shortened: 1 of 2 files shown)


simplesmiler:
This is what I get for making commits from github onsite editor.
Looking into why it fails.

src/compiler/compile-props.js
```diff
@@ -229,10 +229,10 @@ function makePropsLinkFn (props) {
 
 function processPropValue (vm, prop, rawValue, fn) {
   const isSimple = prop.dynamic && isSimplePath(prop.parentPath)
-  let value = coerceProp(prop, rawValue)
-  if (value === undefined) {
-    value = getPropDefaultValue(vm, prop)
+  if (rawValue === undefined) {
+    rawValue = getPropDefaultValue(vm, prop)
   }
+  let value = coerceProp(prop, rawValue)
```
posva: If you want to keep compatibility with the old version you'll need to check again if the value is undefined, althought I don't see why your coerce fn would return undefined...

posva: I don't see a reason to support returning undefined in coerce fn to set the value to the default. Just wanted to raise the point, we never know :smiley:

simplesmiler:
@yyx990803 what do you think?

src/compiler/compile-props.js
yyx990803: in most cases you wouldn't use the two together actually... coerce can already take care of default values.

JosephusPaye:
Thanks for the fix. I will use the workaround @simplesmiler  provided in #2606 until a new version is released.

PR #2615, opened 2016-04-02
Update deprecated phantomjs version.

mohamedhayibor:
The current version has been deprecated. Use `phantomjs-prebuilt` instead.


yyx990803:
Thanks, we are intentionally using PhantomJS 1.9 for now due to breaking changes in 2.x. Will upgrade when it's appropriate.

blake-newman:
Not to day if you can get the 2.x working we won't accept.

PR #2617, opened 2016-04-03
Use more serious strings in tests

phanan:
As discussed, this commit aims to make some tests a bit more serious by replacing `'hi hi ho ho'` with `'foo bar'` :)

### DIFF
--- a/test/e2e/markdown.js
+++ b/test/e2e/markdown.js
@@ -13,9 +13,9 @@ casper.test.begin('markdown', 5, function (test) {
   .then(function () {
     this.sendKeys(
       'textarea',
-      '## yo\n\n' +
-      '- test\n' +
-      '- hi\n\n',
+      '## foo\n\n' +
+      '- bar\n' +
+      '- baz\n\n',
       { keepFocus: true }
     )
     // keyUp(13)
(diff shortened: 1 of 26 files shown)


azamat-sharapov:
I liked previous version soooo much better than this "serious" one.. This is open source.. why so serious? :)

yyx990803:
Well, it's good to have some consistency for dummy text usage :) The e2e tests seem to be failing though.

phanan:
Ok that's weird:

image

Let me check again.

phanan:
@yyx990803 I think this is not a problem with my code per-se. @chrisvfritz is having the same problem with the two `e2e`.

chrisvfritz:
@phanan I'm only experiencing the issue with the updated phantomjs binary though (in phantomjs-prebuilt). :confused:

azamat-sharapov:
Let's just keep sourcecode fun, as it is :smile:

yyx990803:
@azamat-sharapov I'd like to keep it fun, but also consider Vue's adoption in the more "serious" world :)

That said, @phanan is pretty good at writing humorous comments, so maybe we can squeeze some of those in in the future.

Issue #2507, opened 2016-03-18
Suggestion: Plugin directive for slot


paulpflug:
Hi,
Fairly often ([1](https://github.com/paulpflug/vue-clusterize), [2](https://github.com/vue-comps/vue-position-after-transform), [3](https://github.com/paulpflug/vue-materialize)) when building reusable components I use the same pattern of nested components.
The pattern consists of a parent component using the slot-directive and a child component which must be nested to work correctly. The child component also has a slot-directive to contain the custom content.

The aim of the pattern is to have a defined parent-child behavior with the support for custom content.
(Would be useful for all sorts of popular ui-components, like wizard, accordion, etc.)

Example from [vue-position-after-transform](https://github.com/vue-comps/vue-position-after-transform):
```html
<vc-pat> //parent component
  <vc-patc> //child component
    Text 1  //custom content
  </vc-patc>
  <vc-patc> // parent:child - 1:n
    Text 2
  </vc-patc>
</vc-pat>
```

There are several problems with this approach in vue. Most important, components build with this pattern can't be nested and data passing to the child is verbose (using `$children` in the parent and filtering only the "real" child components).

example from [vue-position-after-transform](https://github.com/vue-comps/vue-position-after-transform):
```coffee
for child in @$children
  # reason why I can't nest the components, 
  # $children will contain ALL somewhere nested patc-components
  # not only the direct children
  if child.isPatc 
    process(child)
```

I was thinking about it for a while and I found a good solution:
An additional directive 'plugin' on the `slot` component with the following features:
- slot only accepts components (should throw If has plain content)
- `v-bind` to pass data to components
- deeper nested components are put in childs `$children` instead of parents
- child should be evaluated in the parent's parent scope, like normal slot
- the directive optionally takes a string which must match the `name` of the child component (should throw if a direct child has no or the wrong name) or an expression which is evaluated in the childrens scope and must be true.

The last point would allow different children with different behavior on the same parent,
for example when building an accordion, there could be a child component which collapses when another one is opened and one which remains opened.

In the template of the parent component `slot` with `plugin` would be used and in the child `slot` without `plugin`

The downside: I looked into the slot component and found no easy way of achieving this.

What do you think?

prog-rajkamal:
man, i got the same idea and was planning to raise an issue tonight.

my idea was:
1)  add a directive `v-default-tag`, with default value empty("") for current behavior.
2) on values like `v-default-tag="div"`, `<div>` tag will be used to contain fragments.

[...]

PR #2619, opened 2016-04-03
Plugin directive for slot

paulpflug:
This is WIP !
There are several TODOs remaining

I just wanted to show, where I was aiming at with #2507.

### DIFF
--- a/src/directives/element/slot.js
+++ b/src/directives/element/slot.js
@@ -9,7 +9,7 @@ import {
 export default {
 
   priority: SLOT,
-  params: ['name','plugin',''],
+  params: ['name','plugin'],
 
   bind () {
     // this was resolved during component transclusion
@@ -78,6 +78,10 @@ export default {
         }
         // child relations test: how to get the slot content to be comp11 child?
         this.unlink = context.$compile(content, host, scope, this._frag)
+        if (this.params.plugin !== true) {
+          // should check on the newly created vm if it is valid
+        }
+
       } else {
         this.unlink = context.$compile(content, host, scope, this._frag)
       }
--- a/src/directives/element/slot.js
+++ b/src/directives/element/slot.js
@@ -2,13 +2,14 @@ import { SLOT } from '../priorities'
 import {
   extractContent,
   replace,
-  remove
+  remove,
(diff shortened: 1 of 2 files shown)


src/directives/element/slot.js
```diff
@@ -38,9 +39,47 @@ export default {
         elseBlock._context = this.vm
         content.appendChild(elseBlock)
       }
-      const scope = host
+      let scope = host
         ? host._scope
         : this._scope
+
+      if (this.params.plugin) {
+        console.log(this)
```
skyronic: stray console log.

src/directives/element/slot.js
```diff
@@ -38,12 +39,48 @@ export default {
...
+          if (scope[data] == null) {
+            defineReactive(scope,data,context._data[data])
+          }
+        }
+        // add all data from host to scope with prefixed names
+        for(let data in host._data) {
+          defineReactive(scope,'_'+data,host._data[data])
+        }
+        // child relations test: how to get the slot content to be comp11 child?
+        this.unlink = context.$compile(content, host, scope, this._frag)
```
paulpflug: I can't find `vm.$compile` in the source, where does it come from?

I don't understand it, yet..

src/directives/element/slot.js
```diff
@@ -38,12 +39,48 @@ export default {
...
+
+        // add all data from context to scope
+        for(let data in context._data) {
+          if (scope[data] == null) {
+            defineReactive(scope,data,context._data[data])
+          }
+        }
+        // add all data from host to scope with prefixed names
+        for(let data in host._data) {
+          defineReactive(scope,'_'+data,host._data[data])
```
paulpflug: best way I can think of to avoid naming conflict of context and host.
Forces the user of `plugin` to prefix the variables in his markup when using slot
````html
<slot plugin :text="_text"></slot>
```
But the functionality outweighs the inconsistency (and it is hidden from the user, as it should be)

yyx990803:
Hi! Closing this because I don't think we will introduce any major changes to slot in 1.x, and in 2.x slots will become a lot more flexible with programmatic render functions.

PR #2620, opened 2016-04-04
[fix] Don't remove newlines during parseText

sirlancelot:
Since `text` is being used as the key in `cache.get(text)` and the
cache is checked *before* newlines were removed, the cache would always
miss if the input `text` had *any* newlines.

I updated `tagRe` to match tags which contain newlines, and added an
extra test case to ensure newlines outside of tags are preserved


sirlancelot:
Updated with passing tests

sirlancelot:
Thanks!

PR #2627, opened 2016-04-06
Make TodoMVC a self contained example

hugofqueiros:
I am just starting to learn this great framework, and I came across the standard TODO example. It is quite simple to put it to work, but I feel it should be a self-contained project.
This also allows the removal of code from dependencies that were added to vue repo, which I don't agree that should be there.

Keep up your great work.


yyx990803:
Thanks for the PR, but this example uses the local build for e2e testing purposes. You can, however, find a self-contained version in the official TodoMVC repo: https://github.com/tastejs/todomvc/tree/gh-pages/examples/vue

PR #2647, opened 2016-04-09
Fix #2639 v-show + v-else on components

### LINKED ISSUE #2639: v-else not work when together with v-show on custom component
### Vue.js version
1.0.21

### Reproduction Link
http://www.webpackbin.com/412B9bgJb

### Steps to reproduce
click Toggle button

### What is Expected?
when MyComponent is showing, the msg should be hidden

the msg always shows

posva:
Hi,
The actual problem is that `this.el.nextElementSibling` is null in the `bind` call and in the first `update` call.

I'd like some help on this issue @yyx990803 :stuck_out_tongue: because this solutions feels dirty and I think I'm missing something.

### DIFF
--- a/src/directives/public/show.js
+++ b/src/directives/public/show.js
@@ -5,16 +5,25 @@ export default {
 
   bind () {
     // check else block
-    var next = this.el.nextElementSibling
+    let next = this.el.nextElementSibling
     if (next && getAttr(next, 'v-else') !== null) {
       this.elseEl = next
     }
+    // we need a second check for components using this directive
+    this.checkCounter = 0
   },
 
   update (value) {
     this.apply(this.el, value)
     if (this.elseEl) {
       this.apply(this.elseEl, !value)
+    } else if (this.checkCounter < 2) {
+      this.checkCounter++
+      let next = this.el.nextElementSibling
+      if (next && getAttr(next, 'v-else') !== null) {
+        this.elseEl = next
+        this.apply(this.elseEl, !value)
+      }
     }
   },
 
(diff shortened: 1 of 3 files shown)


posva:
After further analyse, I think the method `checkComponent` can also add directives with higher priority so they get executed before the component itself and then we could set the priority of the `show` directive to 1500. This way the `show` directive will get binded before the `component` directive and get access to sibling elements.

@yyx990803 Tell me your thoughts on this. I'm willing to keep working on the issue because it's interesting :smile:

Twiknight:
`v-show` and `v-if` both check its `elseEl` separately, that might be something not dry...

How about making `v-else` a separate directive and assign it the duty to check for previous `v-show` and `v-if`

posva:
@Twiknight hmm, in that case `v-else` would have to reintegrate the behaviour of `v-if` and `v-show`. The problem would however be the same (about the component directive not respecting priority)

Twiknight:
@posva Yes,you're right, `v-if` and `v-show` acts somehow differently..

Maybe we can simply replace `v-else` with its previous tag at compile time?  
Thus we don't need to consider which tag `v-else` is following at run time.

posva:
@Twiknight Yes, although it won't solve the issue about `v-show` either :sweat_smile:

yyx990803:
Unfortunately this fix doesn't fix the issue when the `v-show`'s value is `true` on first render (both branches are shown). I'll take a look to see if there's a better way to handle this.

posva:
I thought that optimisation was only written for the `if` directive.
Tell me if you find a better way. I'm still interested in coding it because I may learn a lot from it :smile:

Issue #2658, opened 2016-04-12
capture modifier not work in safari


eric6356:
Only repro with safari, both iOS and Mac.
Vue added some `keyFilter` function as the event listener mistakenly.
But nothing will happen with chrome.

PR #2659, opened 2016-04-12
fix key modifier filter #2658

eric6356:
this line may be too long
don't know if you have any 80 max-len lint rules


yyx990803:
Maybe we should extract non-key-modifiers into an array and just do a indexOf check here.

Issue #2657, opened 2016-04-12
template-tag, v-for, v-if together, not as expect


hisland:
### Vue.js version
1.0.21

### Reproduction Link
https://jsfiddle.net/hisland/f9gdg0d1/

### Steps to reproduce
i want a condition v-for

```
template(v-for="a in 6" v-if="a%2")
  p {{a}}
```
i got 0-5, i don't want this
p 0
p 1
p 2
p 3
p 4
p 5

```
  template(v-for="a in 6")
    template(v-if="a%2")
      p {{a}}
```

this is ok
p 1
p 3
p 5

### What is Expected?
can i use template(v-for="a in 6" v-if="a%2")?

simplesmiler:
Closed with https://github.com/vuejs/vue/issues/2663

PR #2663, opened 2016-04-12
Fix handling v-if along with v-for on a template tag

simplesmiler:
Proposed fix for https://github.com/vuejs/vue/issues/2657

### DIFF
--- a/src/fragment/factory.js
+++ b/src/fragment/factory.js
@@ -18,7 +18,7 @@ export default function FragmentFactory (vm, el) {
   this.vm = vm
   var template
   var isString = typeof el === 'string'
-  if (isString || isTemplate(el) && !hasDirectives(el)) {
+  if (isString || isTemplate(el) && !el.hasAttribute('v-if')) {
     template = parseTemplate(el, true)
   } else {
     template = document.createDocumentFragment()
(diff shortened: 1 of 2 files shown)


src/fragment/factory.js
```diff
@@ -41,6 +41,15 @@ export default function FragmentFactory (vm, el) {
   this.linker = linker
 }
 
+function hasDirectives (el) {
```
yyx990803: this could be simpler, since `v-for` and `v-if` are the only ones that can work on `<template>`, and `v-for` has higher priority than `v-if`, it seems `v-if` is the only thing that we need to check.

paulpflug:
I'm unsure if this relates: while working on #2619, I noticed, that `v-for` is evaluated before `slot`, besides the priority says otherwise. 
Could have the same origin? (although it is handy to have `v-for` before `slot`)

simplesmiler:
@paulpflug are you talking  about something like `<slot v-for="...">`? I don't think it's related, no.

paulpflug:
@simplesmiler fine. It just attracted my attention that in both cases priorities don't work (the way I understand them :smile:)

simplesmiler:
@paulpflug this one has to do with the fact that `<template>` losses all attributes when being turned into a fragment factory.

PR #2676, opened 2016-04-15
Add support for decimal places in currency filter

phanan:
I know @yyx990803 has little intention to extend these core functionality, but IMHO you either support something fully or you don't at all. So…

Currently the "currency" filter only accepts one argument: the currency symbol. The number of decimal places is fixed as 2. This can cause troubles with several currencies out there whose number of decimal places can be 0 (e.g. Japanese Yen or Vietnamese Dong) or 3 (e.g. Jordanian Dinar).

This commit modifies the filter to accept an extra optional argument: decimal places, which defaults to 2. With this, one can write `{{ 1234 | currency '¥' 0 }}` to properly display a Yen value. Backward compatibility is maintained.

(Just realized there's a [StackOverflow question](http://stackoverflow.com/questions/35725935/need-a-flexible-currency-filter-in-vuejs) about this very issue, where the OP was recommended to create a new filter, which is anti-DRY IMO).

### DIFF
--- a/src/filters/index.js
+++ b/src/filters/index.js
@@ -64,19 +64,25 @@ export default {
    * 12345 => $12,345.00
    *
    * @param {String} sign
+   * @param {Number} decimals Decimal places
    */
 
-  currency (value, currency) {
+  currency (value, currency, decimals) {
     value = parseFloat(value)
     if (!isFinite(value) || (!value && value !== 0)) return ''
     currency = currency != null ? currency : '$'
-    var stringified = Math.abs(value).toFixed(2)
-    var _int = stringified.slice(0, -3)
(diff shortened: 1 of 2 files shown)


blake-newman:
I still think this should be a separate filter,  called precision.  It had way more use cases,  and can be chained with this.

phanan:
I personally don't think `precision` needs to be a filter – it's a mere `toFixed()` call after all – or if the users want to, they can create it easily. This commit on the other hand extends what Vue currently has.

phanan:
Also, decimal places is certainly a currency property and should be implemented so.

blake-newman:
I agree to some extent with your argument,  I suppose this is down to how @yyx990803 sees it best implemented.

PR #2680, opened 2016-04-16
use absolute exclude path in `build/webpack.test.config.js` and fix error when recursively traverse an object

flytreeleft:
* use absolute path to make sure running tests is OK even if it is in node_modules of other project
* fix 'Maximum call stack size exceeded' when recursively traverse an object

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -334,7 +334,7 @@ Watcher.prototype.teardown = function () {
 function traverse (val, walkedObjs) {
   var i, keys
 
-  !walkedObjs && (walkedObjs = {})
+  walkedObjs = walkedObjs || {}
   if (isArray(val)) {
     i = val.length
     while (i--) traverse(val[i], walkedObjs)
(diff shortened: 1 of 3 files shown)


src/watcher.js
```diff
@@ -331,14 +331,25 @@ Watcher.prototype.teardown = function () {
  * @param {*} val
  */
 
-function traverse (val) {
+function traverse (val, walkedObjs) {
   var i, keys
+
+  !walkedObjs && (walkedObjs = {})
```
posva: I'm curious about this notation. How is it different from `walkedObjs = walkedObjs || {}` ?

flytreeleft: Uh, it's just personal habits, there is no difference between them, do not be feeling trouble with it.

posva: I wasn't complaining, just curious :smile:

yyx990803:
Please do not mix multiple fixes in the same PR, since this makes it harder to review and merge. Submit separate PRs please.

flytreeleft:
Splitting to #2686 and #2685

PR #2685, opened 2016-04-17
Using absolute exclude path in `build/webpack.test.config.js` 

flytreeleft:
Changing the relative `exclude` path to absolute to make sure tests can be running OK even if Vue is in `node_modules` of other project.


flytreeleft:
@yyx990803 sorry, it seems that there is no parallel edit checking. Hope no trouble happened. :smile:

PR #2686, opened 2016-04-17
Fix error when recursively traverse an object

flytreeleft:
Fixing 'Maximum call stack size exceeded' when recursively traverse an object which contains some circular references.


yyx990803:
Thanks!

Issue #2397, opened 2016-02-25
Initial prop value in child component


azamat-sharapov:
### Vue.js version
1.0.16

### Reproduction Link

https://jsfiddle.net/azamatsharapov/w6nry3ms/

### Steps to reproduce
See console for self-explanatory logs.

### What is Expected?
Parent component's data property has been changed in route `activate` hook, so the new value is expected in child component, when passed as `prop`.

### What is actually happening?
Child component has old value of `prop` on initialization.

Please let me know if anything not clear. Thank you.

Mat-Moo:
I've note used router yet, but that looks correct to me? The indexComponent is initialised that kicks the sub component (null) then the route active kicks in and set the message and the watcher kicks in? I might be wrong, but because of the way Vue batches updates, I don't believe the order you see the console messages reflects 100% the way the data is being batched and processed.

azamat-sharapov:
1. `indexComponent` has router `activate` hook, which is fired first
2. In that `activate` hook, we change `vm.message`
3. Then sub-component gets `ready`, but `msg` prop (which is `$parent.message`) has old value
4. Finally sub-component `prop` is updated to newest value of `vm.message`.

I'm starting to think, that this is more related to vue-router. I expected router hooks to process before component lifecycle hooks..

yyx990803:
This is because prop propagation is also async using the internal batcher queue. The child's `ready` is called synchronously after the parent's `activate`, so the prop hasn't been propagated yet. I think I should probably mention this in the doc.

azamat-sharapov:
Hmm, one solution would be to use `$parent.prop`. Anything else you could advice?

yyx990803:
Similar to waiting for DOM updates, you can do a `nextTick` in the child's `ready` hook.

azamat-sharapov:
Interesting! Works, thanks :)

PR #2689, opened 2016-04-18
Fix data functions being called twice. (fix #2687)

### LINKED ISSUE #2687: Data function is called twice not once.
### Vue.js version
1.0.21

### Reproduction Link
http://jsbin.com/doxonuviwe/edit?html

### Steps to reproduce

- Register a Vue component with data function.
```javascript

Vue.component('my-hello', {
	data: function() {
		console.log('hello')
		return {};
	}
});

new Vue({
	el: 'body'
});
```
- Instantiate the component.
```html
<my-hello></my-hello>
<my-hello></my-hello>
```
- You will see in the console (DevTool):
```
hello
hello
hello
hello
```

### What is Expected?
Data function should be called twice since there is only two instances of Vue component.

Data function called four times.

### Quick Fix
Set the data in the `created` hook.
```
Vue.component('my-component', {
	created: function() {
		this.$data  = {}; // Set your data here.
	},
});
```

Hope this helpful.

blake-newman:
- Added test for checking prop to data clash
- Added test to ensure data function is only called once per strut
- Removed this._runtimeData uses 'data' for prop to data clash check

Fixes: [ISSUE-2687](https://github.com/vuejs/vue/issues/2687)

### DIFF
--- a/src/instance/internal/init.js
+++ b/src/instance/internal/init.js
@@ -94,11 +94,6 @@ export default function (Vue) {
     // it will be filled up in _initScope().
     this._data = {}
 
-    // save raw constructor data before merge
-    // so that we know which properties are provided at
-    // instantiation.
-    this._runtimeData = options.data
-
     // call init hook
     this._callHook('init')
 
--- a/src/instance/internal/state.js
+++ b/src/instance/internal/state.js
@@ -87,11 +87,6 @@ export default function (Vue) {
       )
     }
     var props = this._props
-    var runtimeData = this._runtimeData
-      ? typeof this._runtimeData === 'function'
-        ? this._runtimeData()
-        : this._runtimeData
-      : null
     // proxy data on instance
     var keys = Object.keys(data)
     var i, key
@@ -104,7 +99,7 @@ export default function (Vue) {
       //    template prop present
       if (
         !props || !hasOwn(props, key) ||
-        (runtimeData && hasOwn(runtimeData, key) && props[key].raw === null)
+        (data && hasOwn(data, key) && props[key].raw === null)
       ) {
         this._proxy(key)
       } else if (process.env.NODE_ENV !== 'production') {
--- a/test/unit/specs/instance/state_spec.js
+++ b/test/unit/specs/instance/state_spec.js
@@ -9,6 +9,27 @@ describe('Instance state initialization', function () {
     expect('should return an object').toHaveBeenWarned()
   })
 
+  it('should initialize data once per strat', function () {
+    var spyOncePerStrat = jasmine.createSpy('called once per strat')
+    const VM = Vue.extend({
+      data: function () {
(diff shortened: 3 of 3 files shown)


blake-newman:
@vuejs/collaborators Anyone with any opinions why `this._runtimeData` is used...

kazupon:
@blake-newman 
I think that this [commit](https://github.com/vuejs/vue/commit/93c1a50b89e1c263f8c324a36d9292a39131c798) will probably be a good reference for that.  😺

blake-newman:
@kazupon so i guess i can remove my test for the warnings... However tests pass fine if not using `_runtimeData`

dargue3:
Hey, I have been asking about this question on various sites trying to find why my ```data()``` hook is being duplicated... so according to this it's a bug introduced starting at 1.0.18 and the newest commit is fixing? Is ```data()``` supposed to fire twice? Just checking for my own sanity :)

posva:
@dargue3 No, it's not. It should only be fired once 😄

dargue3:
@posva well at least I know I'm not crazy. Interesting how it has taken this long for anyone to notice... Or am I just the only one sending AJAX requests in my ```data()``` hook?

Like I said, through trial-and-error I found that this bug was introduced in ```v1.0.18```

posva:
@dargue3 You might not be the only one but you shouldn't do that 😖 
This may help you http://vuejs.org/guide/components.html#activate-Hook

Thanks for reporting that @dargue3

yyx990803:
This was partially a dirty workaround to allow passing values for props in the `data` option. But now I think it was a bad idea to overload `data` for that purpose.

So now I've introduced a dedicated option `propsData` for passing props at instantiation calls.

YerkoPalma:
I don't understand the purpose of the new `propsData` hook. The `props` can be defined with a default value, and data can be initialized in the `data()` hook right?

azamat-sharapov:
@YerkoPalma `props` are resolved async, [as described here](https://github.com/vuejs/vue/issues/2397#issuecomment-188925841) and I think `propsData` is sync, so it's ready in child component init. But let's wait until Evan updates docs about new `propsData`, he will make it clear.

yyx990803:
@YerkoPalma default values are the same for every instance created. `propsData` are a way to provide different prop values at instance creation time without having to wrap it in a parent instance.

e.g. instead of first registering and rendering a parent instance with:

``` js
<test :data="123"></test>
```

You can just

``` js
new Test({
  propsData: {
    data: 123
  }
})
```

This is mostly about facilitating easier unit tests for components, not something you'd use very often in application code.

YerkoPalma:
I see. Everything that help on tests is fine :smile: Thanks for the answer.

blake-newman:
:+1: makes sense

audinue:
That was fast :+1:

PR #2701, opened 2016-04-21
Add an option to extend a component

pespantelis:
#### Related to: vuejs/vue-loader#207

This works as expected for the `<template>` and the `<script>`, but not for the `<style>`.
I noticed that all the `styles` are loaded, but with the wrong priority.

Maybe some changes are needed to the `vue-loader` to figure out that, right?

### DIFF
--- a/src/util/options.js
+++ b/src/util/options.js
@@ -332,6 +332,9 @@ export function mergeOptions (parent, child, vm) {
   }
   var options = {}
   var key
+  if (child.extends) {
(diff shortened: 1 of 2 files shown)


yyx990803:
Can you add a test case for it?

pespantelis:
Yeah, of course!

yyx990803:
I just realized `extends` should also support targeting an already created constructor function, so it should check if `typeof child.extends === 'function'`, and if true, use `child.extends.options` instead. Up for another PR?

PR #2712, opened 2016-04-23
Vue treats `null` expression as `undefined`?

phanan:
Not sure if it's just me misunderstanding how things work, but Vue is apparently treating `null` expression as `undefined`. Here is a failed test case to demonstrate the issue. 

If this is confirmed to be a bug, I can try to work on a fix.


phanan:
@yyx990803 Any idea?

Issue #2707, opened 2016-04-22
"literal" modifier on custom directive causes v-if to be ignored


decademoon:
### Vue.js version
1.0.21

### Reproduction Link
http://jsfiddle.net/5sH6A/317/

blake-newman:
Quick fix: `<div v-test.literal="" v-if="false" >1. This should not be visible</div>`

Placing the literal before v-if resolves this bug for a quick fix. This is indeed a bug.

@yyx990803 Assigning to myself to look at.

blake-newman:
@vuejs/collaborators FYI

This is due to the modifiers of directives being parsed to directives. so `.literal` is being fed to the previous directive even though it's not applied to that directive.

simplesmiler:
Wowsie.
Does not happen with every directive/modifier though, e.g. `@click="doSomething" @mouseover.prevent` does not prevent the click default.

blake-newman:
@simplesmiler Happens with terminable directives

Found the issue and creating PR now

PR #2714, opened 2016-04-23
[HOTFIX] [ISSUE-2707] Fix modifiers being incorrectly passed to terminal directives

blake-newman:
- Fixes [ISSUE-2707](https://github.com/vuejs/vue/issues/2707)
- Terminal directives where incorrectly revieving modifiers from other inline directives


azamat-sharapov:
Oh those caps.. :D JIRA style

blake-newman:
@azamat-sharapov Always done it even before i ever used JIRA, i have a little OCD when it comes to being able to easily filter issues and pr's

yyx990803:
Just for convention sake - can we use (fix #2707) at the end of the title/commit message and don't use CAPS :)

PR #2724, opened 2016-04-25
Fixed the fragment instance link in the console warning

sebastiandedeyne:
Title describes it!


blake-newman:
Thanks for the fix 👍

PR #2740, opened 2016-04-26
added v-once directive unit tests

blake-newman:


test/unit/features/directives/once.spec.js
```diff
@@ -0,0 +1,110 @@
+import Vue from 'vue'
+
+describe('Directive v-once', () => {
+  it('should not rerender component', done => {
+    const vm = new Vue({
+      el: '#app',
+      template: '<div v-once>{{ a }}</div>',
+      data: { a: 'hello' }
+    })
+    waitForUpdate(() => {
```
yyx990803: the initial rendering is sync so you don't have to wait to assert the initial render.

blake-newman: I was wondering this after i submitted, I will quickly ammend :)

PR #2757, opened 2016-04-28
Replace var with proper let/const

phanan:
I guess 2.0 is a perfect opportunity to adapt ES6 fully. Using `const` and `let` instead of `var` is a good start.

### DIFF
--- a/src/compiler/parser/html-parser.js
+++ b/src/compiler/parser/html-parser.js
@@ -63,11 +63,11 @@ export function parseHTML (html, handler) {
     last = html
     // Make sure we're not in a script or style element
     if (!lastTag || !special(lastTag)) {
-      var textEnd = html.indexOf('<')
+      const textEnd = html.indexOf('<')
       if (textEnd === 0) {
         // Comment:
         if (/^<!--/.test(html)) {
-          var commentEnd = html.indexOf('-->')
+          const commentEnd = html.indexOf('-->')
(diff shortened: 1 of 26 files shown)


azamat-sharapov:
Interesting.. what does Evan use to automatically add github labels to PRs? :)

phanan:
@azamat-sharapov I'm not sure, but a quick search for "Github auto labels" yields several results for me ;)

egoist:
Whatever, he must be using the github webhook 😂

azamat-sharapov:
@phanan well those results don't give anything obvious, so I thought maybe it's custom script running on some server.. sorry for offtop though :)

phanan:
@azamat-sharapov True that… Let's hope @yyx990803 unveils his secret.

nervgh:
@phanan 
> I guess 2.0 is a perfect opportunity to adapt ES6 fully. Using const and let instead of var is a good start.

The constants should be in UPPERCASE (a best practice / naming conventions)
Example https://github.com/nervgh/recursive-iterator/blob/master/src/RecursiveIterator.js#L8

azamat-sharapov:
@nervgh can't agree with that, it isn't really constant.. `const` in JS is different, you probably already know how const works, compared to many other languages.

phanan:
What @azamat-sharapov said.

PR #2768, opened 2016-04-28
fix dynamic Style's position should be located in before static Style

okoala:


yyx990803:
why does this matter?

okoala:
Same as class, Look neat.

yyx990803:
I intentionally put it at the end so that it doesn't matter if the inline style has an ending `;` or not.

Putting it first you may end up with two `;;` so you will have to check for an ending `;` - this can be avoided by just placing static styles last.

PR #2772, opened 2016-04-29
Preserve cursor position when changing input from Vuex.

zigomir:
Fixes https://github.com/vuejs/vuex/issues/162

textarea works different and we need to set it as before.

Same issue is also present on v2.0. @yyx990803 should I open PR for 2.0 too?

### DIFF
--- a/src/directives/public/bind.js
+++ b/src/directives/public/bind.js
@@ -110,11 +110,15 @@ export default {
       attrWithPropsRE.test(attr) &&
       attr in el
     ) {
-      el[attr] = attr === 'value'
+      var attrValue = attr === 'value'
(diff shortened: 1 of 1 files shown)


yyx990803:
`setAttribute` actually doesn't do anything, as the `value` attribute is only used to indicate the initial value. The proper fix I think is checking if the input's value is already the same as the desired value and skip it if yes.

zigomir:
Thanks for quick and useful feedback. I changed it now and it fixes problem with vuex too.
There seem to be some lint issues though.

yyx990803:
I fixed the lint issue earlier - can you try rebasing your branch?

zigomir:
Fix is now in one commit on top of `dev` branch.

PR #2780, opened 2016-04-30
updated dependencies to latest, with better control ranges

blake-newman:
- Updated dependencies to latest version
- Modified semver to use x-ranges
 - Gives much better control over dependencies
 - Helps with tracking issues that could be caused by changes to dependencies


simplesmiler:
I think it's totally reasonable to pin `devDependencies` that way, if none of patch versions were set to target specific bug fixes in tools.

PR #2784, opened 2016-04-30
Preserve cursor position

zigomir:
Same issue as https://github.com/vuejs/vue/pull/2772 but for 2.0.

Not sure if it's correct but fixes same issue.


yyx990803:
This turns out to be a bit more complex than it seems. I'm also looking at alternatives to two-way filters to `v-model` so this part may need to be overhauled.

PR #2796, opened 2016-05-03
improved SSR rendering

blake-newman:
- Reduced duplicated code
- Added more test coverage
- Added component option resolving (async to sync conversion)
- Allows resolvable components
- Fixed component lifecycle
- Added raw render method, to allow custom streaming possibilities


yyx990803:
Let's avoid checking in `dist` files for now. (Only check-in dist on releases)

blake-newman:
Oops didn't mean to..

src/server/resolve-options.js
```diff
@@ -0,0 +1,43 @@
...
+      const total = keys.length
+      if (!total) {
+        return
+      }
+      totalResolvable += total
+      for (let i = 0; i < total; i++) {
+        const name = keys[i]
+        resolve(component.components[name], component.components, name)
+      }
+      return
```
phanan: Isn't `return` here redundant?

blake-newman: Most likely both `return` statements here are not needed.  :)

src/server/resolve-options.js
```diff
@@ -0,0 +1,39 @@
+export function resolveOptions (options, done) {
```
yyx990803: Unfortunately we don't actually need this. I didn't explain it clear enough, but the proper usage would be bundling the root component via Webpack with `target: 'node'`, which already automatically shims `require.ensure` and `require(['...'], cb)` into synchronous requires. So as long as you stick to Webpack's supported code splitting syntax, you don't need to do anything.

src/server/init-component.js
```diff
@@ -0,0 +1,17 @@
+export function initComponent (component) {
```
yyx990803: While this fixes the hook invocation order, this is mounting the entire component tree synchronously. I'm wondering if it's possible to delay the mounting to play nicer with streaming mode...

src/server/resolve-options.js
blake-newman: Ah, did not know this was a feature of webpack. Will remove this functionality!

src/server/init-component.js
blake-newman: Agreed, would need to be able to initialise the component's without mounting.

src/server/resolve-options.js
blake-newman: Only other way I could invision, is to put the whole render on a setTimeout. This will allow any asynchronous flow to be resolved. Would also be more friendly to vue-router ect. With much less code, will mean `renderToString` and `renderToStream` need to output result to a callback.

blake-newman: Resolved: Ensured that synchronous components are able to be resolved.

blake-newman:
@yyx990803 As discussed I have ammended the PR to work with resolving components, this is now done internally to prevent blocking the UI. 

As mentioned we still need to ignore reactivity, and manually call hooks.

src/core/instance/lifecycle.js
```diff
@@ -105,7 +105,7 @@ export function lifecycleMixin (Vue) {
   }
 
   Vue.prototype.$forceUpdate = function () {
-    this._watcher.update()
+    this._update(this._render())
```
blake-newman: Directly call `_update` with `_render` result. In some scenarios `_watcher` was not yet defined. This is also faster anyway.

src/server/render.js
```diff
@@ -1,16 +1,28 @@
...
+        _parentVnode: node,
+        _parentListeners: listeners,
+        _renderChildren: children
+      }
+      // check inline-template render functions
+      const inlineTemplate = node.data.inlineTemplate
+      if (inlineTemplate) {
+        options.render = inlineTemplate.render
+        options.staticRenderFns = inlineTemplate.staticRenderFns
+      }
```
blake-newman: Above could probably be provided by an internal function to reduce duplicated code.

PR #2797, opened 2016-05-03
simplified resolving of components

blake-newman:


src/core/vdom/create-component.js
```diff
@@ -108,25 +104,26 @@ function destroy (vnode) {
...
       // cache resolved
       factory.resolved = res
-      // invoke callbacks
-      for (let i = 0, l = cbs.length; i < l; i++) {
-        cbs[i](res)
+      // invoke forceUpdate
+      while (factory.requested) {
+        // it's ok to queue this on every render because
+        // $forceUpdate is buffered.
+        parent.$forceUpdate()
```
yyx990803: I don't think this is right - this would invoke multiple updates on the same `parent` (the last one). In the previous version different callbacks may contain different parents in their closures.

blake-newman: Ah yes, you are true. Now i have revisited the internals this much more sense

PR #2810, opened 2016-05-06
[Vue 2.0] Fix BooleanAttr & EnumeratedAttr serialization

tejitak:
Values of BooleanAttr & EnumeratedAttr are added even when their values are `false`. (The attrs are not added when the value is 'null')

For example, the following ':selected' values currently result in `selected="selected"` for both options.
```js
{
      data: {
        value: 'bar'
      },
      template: `
        <div>
          <select>
            <option value="foo" :selected="value === 'foo'">Foo</option>
            <option value="bar" :selected="value === 'bar'">Bar</option>
          </select>
        </div>
      `
}
```
Only second option should have selected attribute.

F.Y.I. As a current work around, we can set the prop like `:selected="value === 'foo' ? true : null`.

### DIFF
--- a/src/platforms/web/runtime/modules/attrs.js
+++ b/src/platforms/web/runtime/modules/attrs.js
@@ -1,4 +1,4 @@
-import { isBooleanAttr, isEnumeratedAttr, isXlink, xlinkNS } from 'web/util/index'
+import { isBooleanAttr, isEnumeratedAttr, isXlink, xlinkNS, getXlinkProp } from 'web/util/index'
 
 function updateAttrs (oldVnode, vnode) {
   if (!oldVnode.data.attrs && !vnode.data.attrs) {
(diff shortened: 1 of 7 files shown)


posva:
@tejitak Don't forget about the test 😄

yyx990803:
👍 if you can add add a test case that'd be great

tejitak:
Okay! I'll try to add a test case for that soon.

tejitak:
@posva @yyx990803 
While preparing the test cases, I found some attribute serialization are different than v1.0 output like the following.

| Attr | current v2.0 output | v1.0 output |
|:-----------|:------------|:------------|
:test="null" | attr value: null | attr removed
:test="false" | attr value: 'false' | attr removed
:test="true" | attr value: 'true' | attr value: ''
:xlink:special="null" | attr value: null | attr removed
:draggable="false" | attr value: 'true' | attr value: 'false'
:disabled="false" | attr value: 'disabled' | attr removed

I modified the 2.0 serialization code to be matched with v1.0 output.

Note that current code of `el.data.attrs['server-rendered'] = true` will have blank attribute by applying the spec. I changed this into String `el.data.attrs['server-rendered'] = 'true'`

More detail, please refer the test cases.

yyx990803:
@tejitak nice job!

yyx990803:
FYI I made some changes in 5fb1a54 to make the behavior more consistent (and different from 1.x):

- Only `null`, `undefined` and `false` are considered to be falsy values in all cases, including for enumerated attributes. This means `:draggable="''"` will result in `draggable="true"`.

- For enumerated attributes, in addition to the falsy values above, the string value of `"false"` will also render as `attr="false"`.

PR #2817, opened 2016-05-09
Simplify proxy() and defineReactive() using es6 method definitions

phanan:


yyx990803:
This is actually intentional so that the functions have distinctive names in stack traces during debugging.

phanan:
Oh, I see.

PR #2840, opened 2016-05-10
Fix issue with renderToString stack length overflowing

blake-newman:


src/server/create-renderer.js
```diff
@@ -10,9 +10,18 @@ export function createRenderer ({
   return {
     renderToString (component) {
       let result = ''
+      let stackDepth = 0
       render(component, (str, next) => {
         result += str
-        next && next()
+        if (next) {
+          if (stackDepth >= 1000) {
```
yyx990803: Let's export `MAX_STACK_DEPTH` as a const here and import it in `render-stream.js`.

blake-newman:
Needs revision... Need to either synchronously avoid stack overflow, or provide result in callback.

yyx990803:
We can change the `renderToString` interface to be a standard Node async function:

``` js
```

PR #2841, opened 2016-05-10
SSR no data observations

blake-newman:
- `$isServerRenderer` added to Vue prototype
- Turn off data observations with `$isServerRenderer`
- `server-renderer` exports Vue runtime without compiler for convenience
- Updated SSR test imports
- Added test coverage for data observation

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -10,6 +10,7 @@ import {
   hasOwn,
   isReserved
 } from '../util/index'
+import Vue from '../instance/index'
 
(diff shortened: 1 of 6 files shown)


yyx990803:
I don't think this is how we want to approach it. This basically forces users to use the Vue from `server-renderer.js` for SSR, but the desired UX is you'd be using the same component codebase in which you would do `import Vue from 'vue'`.

I'll take another look at this later.

blake-newman:
Fair enough makes sense, didn't think of that usage.

PR #2842, opened 2016-05-10
Added benchmark for SSR

blake-newman:
- Includes `renderToStream` and`renderToString`
- Added SSR benchmark README.md
- Renders table with 10000 rows with 100 columns (1 million cells)

### DIFF
--- a/benchmarks/ssr/README.md
+++ b/benchmarks/ssr/README.md
@@ -1,0 +1,13 @@
+# Vue.js SSR benchmark
+
+This benchmark renders a table of 1000 rows with 10 columns (10 thousand components). This benchmark is to demonstrate the overall speeds of Vue.js SSR and time to content comparison between `renderToString` and `renderToStream`.
+
+To view the results follow the run section. Note that the overall completion time for the results are variable, this is due to other system related variants at run time (available memory, processing ect). In ideal circumstances both should finish within equal times.
+
+`renderToStream` pipes the content through a stream which gives massive performance benefits over renderToString. This can be observed through the benchmark.
+
+### run
+
+``` bash
+npm run bench:ssr
+```
--- a/benchmarks/ssr/common.js
+++ b/benchmarks/ssr/common.js
@@ -1,0 +1,86 @@
+'use strict'
+
+const compiler = require('../../dist/compiler.js')
+const self = (global || root)
+
+self.performance = {
+  now: function () {
+    var hrtime = process.hrtime()
+    return ((hrtime[0] * 1000000 + hrtime[1] / 1000) / 1000)
+  }
+}
+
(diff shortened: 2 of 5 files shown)


yyx990803:
Hold this (will need to amend after data observation skipping is properly implemented)

blake-newman:
Yup, put up so others can inspect code in the mean time.

benchmarks/ssr/common.js
```diff
@@ -0,0 +1,87 @@
...
+  template: '<div><h1>{{ performance }}ms</h1><my-table></my-table></div>',
+  mixins: [perfMixin],
+  components: {
+    myTable: {
+      data: function () {
+        return {
+          grid: gridData
+        }
+      },
+      template: '<table width="100%" cellspacing="2"><row v-for="row in grid"></row></table>',
```
yyx990803: actually the benchmark is broken because you are not passing in the `:row` prop at all :/

blake-newman: I was wondering why all of a sudden node was able to render 10 million cells without running into memory issues.

Fixed.

benchmarks/ssr/common.js
```diff
@@ -0,0 +1,86 @@
...
+        return {
+          grid: gridData
+        }
+      },
+      template: '<table width="100%" cellspacing="2"><row v-for="row in grid" :row="row"></row></table>',
+      components: {
+        row: {
+          props: ['row'],
+          mixins: [perfMixin],
+          template: '<tr><th>{{ performance }}ms</th><column v-for="item in row"></column></tr>',
```
yyx990803: Here it should be `v-for="item in row.items"`... previous versions were ignoring the columnCount all together. This explains why numbers were off - we should probably just test 1000 x 10 which is a more practical scenario (Note it's 10k components, not just cells)

benchmarks/ssr/README.md
```diff
@@ -0,0 +1,14 @@
...
+This benchmark renders a table of 10000 rows with 100 columns (1 million cells). It is extremely likely that you will never render 1 million table cells, this benchmark is to demonstrate the overall speeds of Vue.js SSR and time to content comparison between `renderToString` and `renderToStream`.
+
+To view the results follow the run section. Note that the overall completion time for the results are variable, this is due to other system related variants at run time (available memory, processing ect). In ideal circumstances both should finish within equal times.
+
+`renderToStream` pipes the content through a stream which gives massive performance benefits over renderToString. This can be observed through the benchmark.
+
+### run
+
+``` bash
+VUE_ENV=server node renderToString.js
```
yyx990803: Also add `NODE_ENV=production` here - probably make it an npm script to run both.

PR #2844, opened 2016-05-10
SSR warning for no VUE_ENV and added tests

blake-newman:
- Added test coverage for data observation and VUE environment variable
- Moved Test helpers into own folder for reusability


test/ssr/ssr-env.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+describe('SSR: VUE_ENV=server', () => {
+  it('_isServer set as "server" on Vue config', () => {
+    const vm = new Vue({
+      data: {
+        foo: 'server',
+        bar: 'rendering'
+      }
+    })
+    expect(vm.$isServer).toBe(true)
```
yyx990803: doesn't seem to be testing what is described ;)

test/ssr/ssr-env.spec.js
```diff
@@ -0,0 +1,52 @@
...
+          return this.foo + this.bar
+        }
+      }
+    })
+
+    vm.foo = ''
+
+    expect(vm.foo).toBe('')
+    expect(vm.combined).toBe('rendering')
+    expect(vm.__ob__).toBe(undefined)
```
yyx990803: Should be asserting `vm.$data.__ob__` here since Vue instances themselves are never observed.

PR #2849, opened 2016-05-11
Fix eslint error

kazupon:
In `dev` branch, I upgrade eslint and fixed eslint error because eslint is too old.


blake-newman:
LGTM :)

PR #2854, opened 2016-05-11
[next][advice]new render mechanism

Jinjiang:
In web the best practise of first time rendering is just append the whole dom tree once into the body. But in some other virtual dom situation like mobile web or native app, the first time rendering is not as fast as users' expectation. So I tried to make "a little" change to implement "stream rendering" to optimize the first-screen perf. And let it be able to append the whole dom tree either once or node-by-node:

1. changed `__h__(tag, data, children)` into `__r__(__s__(tag, data), children)` so the parent node could be calculate->generate->append first and the children later.
2. supported `stream: true` option and added a new `__stream_patch__` method besides `__patch__` to optimize the first-time rendering logic. If no `stream: true` option set, just render as normal before.
3. supported `atom: true` in `vnode.data` to describe whether this node will be appended with its children together as a atom or child-by-child in "stream rendering" mode.

Additionally, I changed the `class`/`staticClass` from string to array and added an object `staticStyle` for better virtual dom control.

### DIFF
--- a/src/compiler/codegen.js
+++ b/src/compiler/codegen.js
@@ -23,7 +23,7 @@ export function generate (ast, options) {
   warn = options.warn || baseWarn
   platformDirectives = options.directives || {}
   isPlatformReservedTag = options.isReservedTag || (() => false)
-  const code = ast ? genElement(ast) : '__h__("div")'
+  const code = ast ? genElement(ast) : '__r__(__s__("div"))'
   staticRenderFns = prevStaticRenderFns
   return {
     render: `with (this) { return ${code}}`,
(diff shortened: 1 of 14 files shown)


yyx990803:
There are multiple changes (streaming, class changes, style changes) mixed in this big PR, making it harder to review and merge. We should split them into smaller PRs, each addressing a specific issue.

Also re streaming mode: since this feature is very platform-specific, I wonder if it's possible to implement it in a less intrusive / more decoupled fashion. The concern is that these features will mostly be used for weex, but weex itself doesn't live in this repo, so coordinating changes between the two projects will become problematic in the long run.

Ideally, weex-specific feature requirements should live in weex and should be able to iterated on without relying on Vue core to pull in the changes. One possible approach to achieve that is implementing the minimal changes required in core to expose an interface that allows a platform to customize the patching/codegen process. That might be difficult in the short term, but this patch as it is just seems to be coupling the two projects too much with serious maintenance cost.

Jinjiang:
Yes your concern is true. My point is that there is no way to get streaming mode based on `__h__`. I can try to make another smaller pr to use `__r__` & `__s__` only but no streaming mode and atom option.

Another thing I think the `class` & `style` process is too tricky for web, not general enough. So another small discuss or pr will be made later.

blake-newman:
We have recently introduced a VUE_ENV environment variable. You may be able
to aid support for this by using this.

This way we can opitimize the build to use cases. As Evan said this is
aimed mainly to Weex or other native code implementations, so being able to
target the functionality may be of favour.

Jinjiang:
Thx. I will have a check about VUE_ENV

PR #2855, opened 2016-05-12
add observer tests

kazupon:
NOTE: bring vue 1.x observer test

### DIFF
--- a/test/unit/modules/observer/observer.spec.js
+++ b/test/unit/modules/observer/observer.spec.js
@@ -4,9 +4,9 @@ import {
   observe,
   set as setProp,
   del as delProp
-} from '../../../../src/core/observer/index'
-import Dep from '../../../../src/core/observer/dep'
-import { hasOwn } from '../../../../src/core/util/index'
+} from 'core/observer/index'
+import Dep from 'core/observer/dep'
+import { hasOwn } from 'core/util/index'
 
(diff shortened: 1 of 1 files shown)


test/unit/modules/observer/observer.spec.js
```diff
@@ -0,0 +1,320 @@
+import Vue from 'vue'
+import {
+  Observer,
+  observe,
+  set as setProp,
+  del as delProp
+} from '../../../../src/core/observer/index'
```
yyx990803: can just use `core/observer/index` because the build has [these aliases](https://github.com/vuejs/vue/blob/next/build/alias.js)

test/unit/modules/observer/observer.spec.js
```diff
@@ -0,0 +1,320 @@
...
+    expect(obj.__ob__).toBe(ob1)
+    // should've walked children
+    expect(obj.a.__ob__ instanceof Observer).toBe(true)
+    expect(obj.b.__ob__ instanceof Observer).toBe(true)
+    // should return existing ob on already observed objects
+    const ob2 = observe(obj)
+    expect(ob2).toBe(ob1)
+  })
+
+  it('create on null', function () {
```
yyx990803: missed arrow function change

kazupon:
Done

PR #2857, opened 2016-05-12
new render mechanism

Jinjiang:
1. Modified `__h__(tag, data, children, namespace)` into `__r__(__s__(tag, data, namespace), children)` to separate node-self process and children process.
2. Supported getters for text node (`__t__`) and static root (`__m__`) to make general detail control possible.


src/core/instance/render.js
```diff
@@ -18,7 +18,11 @@ export function initRender (vm: Vue) {
   vm.$slots = {}
   // bind the public createElement fn to this instance
   // so that we get proper render context inside it.
-  vm.$createElement = bind(createElement, vm)
+  vm.$createElement = function (tag) {
+    const __r__ = bind(renderElement, vm)
```
yyx990803: 这里的绑定不应该在函数体内做（等于每一次调用都创建两个新函数）
直接 `this.__r__(this.__s__())` 即可

Jinjiang: Fixed it and passed all flow check & ssr cases.

yyx990803:
Made some simplifications based on your changes in https://github.com/vuejs/vue/commit/2dc5b1782871d87a7c795361eac0d49fd396c05f:

1. renaming to make function names for self-apparent
  - `renderSelf` -> `renderElement`
  - `renderElement` -> `renderElementWithChildren`

2. removed the need for the `ComponentDef` - `renderElement` always return `VNode | void`, instead we do different set children ops depending on whether the vnode is a component or not.

PR #2858, opened 2016-05-12
Add support for v-bind:content shorthand (fixes #2843)

### LINKED ISSUE #2843: Feature request: `<div v-bind:content>` as a shorthand for `<div v-bind:content="content">`
This pattern is quite common. The intent of the shorthand seems clear to me, and it currently has no effect I'm aware of.

phanan:
I personally quite like the feature request #2843, so I went ahead and try to support it. 

With this commit, `v-bind` with an empty expression will imply the expression to be the directive name, i.e. `v-bind :content` will have the same effect as `v-bind:content="content"`, `:href` will have the same effect as `:href="href"` and so on so forth. Note that only attribute bindings are affected by this change. Event, transition, class/style bindings etc. remain untouched. 

Not sure if I'm missing anything here – all unit tests passed fine, though. Let me know if there's anything to fix or improve, or just close this PR if you don't intend to support the feature.

### DIFF
--- a/src/compiler/compile.js
+++ b/src/compiler/compile.js
@@ -14,7 +14,8 @@ import {
   checkComponentAttr,
   findRef,
   defineReactive,
-  getAttr
+  getAttr,
+  camelize
 } from '../util/index'
 
 // special binding prefixes
@@ -716,6 +717,11 @@ function compileDirectives (attrs, options) {
         pushDir(dirName, internalDirectives[dirName])
       } else {
         arg = dirName
+        // support for shorthand expression (#2843)
+        // <div v-bind:content> => <div v-bind:content="content">
+        if (!value) {
+          value = camelize(dirName)
+        }
         pushDir('bind', publicDirectives.bind)
       }
     } else
--- a/test/unit/specs/compiler/compile_spec.js
+++ b/test/unit/specs/compiler/compile_spec.js
@@ -112,6 +112,8 @@ describe('Compile', function () {
     el.setAttribute(':class', 'a')
     el.setAttribute(':style', 'b')
     el.setAttribute(':title', 'c')
+    el.setAttribute(':content', '')
+    el.setAttribute(':data-content', '')
 
     // The order of setAttribute is not guaranteed to be the same with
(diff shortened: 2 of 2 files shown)


bopjesvla:
Nice! Something I just thought of: `v-bind:data-content` and `v-bind:ns:attr` should be equivalent to `v-bind:data-content="dataContent"` and `v-bind:ns:attr="nsAttr"`, respectively. I can submit a PR to your PR for that.

A second thing, which may be out of the scope of this PR: it'd be nice if custom directives had the ability to enable this behaviour.

phanan:
> `v-bind:data-content` and `v-bind:ns:attr` should be equivalent to `v-bind:data-content="dataContent"` and `v-bind:ns:attr="nsAttr"`

You're right, I overlooked these (especially the first part). Happy to receive your PR. @yyx990803 may or may not approve this feature btw.

src/compiler/compile.js
```diff
@@ -716,6 +717,11 @@ function compileDirectives (attrs, options) {
         pushDir(dirName, internalDirectives[dirName])
       } else {
         arg = dirName
+        // support for shorthand expression (#2843)
+        // <div v-bind:content> => <div v-bind:content="content">
+        if (!value) {
```
posva: Are you sure about this? We don't want this to happen when the value is `null` or `undefined`?

bopjesvla: I'm not sure if this is what you mean, but `value` is the expression, not the return value of the expression. So you can still do `v-bind:content="null"`.

posva: You're right, that's the rawValue. My bad!

rpkilby:
> v-bind:ns:attr="nsAttr"

Is namespacing with colons an accepted syntax with Vue? Not sure if that would be accepted.

LinusBorg:
@phanan Should we close this PR in favour of #2877 ?

phanan:
Sure. I don't think @yyx990803 is a fan of this change, though.

LinusBorg:
All the more reason to not have two PR's about it :-P

PR #2868, opened 2016-05-13
add watcher tests

kazupon:


yyx990803:
Looks like both PRs have some minor issues - let me know after the fix!

kazupon:
I fixed both PRs!

PR #2869, opened 2016-05-13
add scheduler tests

kazupon:


test/unit/modules/observer/scheduler.spec.js
```diff
@@ -0,0 +1,117 @@
...
+          id: 2,
+          run () {
+            callOrder.push(3)
+          }
+        })
+        callOrder.push(2)
+      }
+    })
+    waitForUpdate(() => {
+      expect(callOrder.join()).toBe('1,2,3')
```
yyx990803: missing `done()` here!

kazupon: oh, sorry ...

kazupon:
fixed!

PR #2871, opened 2016-05-13
add ssr example

dennisreimann:
this is not finished yet, as I could not get the hydration part to work. the server-side rendering works fine, but somehow the radio button bindings for the `currentBranch` do not seem to work after hydration.

I'd like to get you opinion on whether or not it makes sense to add this example at the current pre-alpha stage. I think it would be great to have a resource for ssr, as there is not much documentation about it yet – I grabbed most of what is in my example from the tests.

If you think this would be a nice addition, can you provide some info on how to make it work?


yyx990803:
Thanks! I think it is probably still to early to have an example, since some of the details and API may still be subject to change. (also need tests for hydration, yes) I think it's better to do this after we enter beta, but I'd appreciate any information on what's not working with hydration at this moment.

dennisreimann:
Thanks for your answer :)

Well, I'm pretty sure it's me not understanding this correctly or missing something. The [created hook](https://github.com/vuejs/vue/pull/2871/commits/1706ad3b7d8442be781883afae0142ab5d3054c0#diff-16cdbc0c23582da9088930a897ce0e93R18) fires on the client-side, which means the basic setup should be working. I expected `$data` to be populated with [what I set on the server](https://github.com/vuejs/vue/pull/2871/commits/1706ad3b7d8442be781883afae0142ab5d3054c0#diff-a3310005786536aaf72e324eef49cee7R72), but it is `undefined`. (in the current state of the code I tried to set `data` in the app.js, but you can ignore that)

The currentBranch binding also does not seem to work, as changes do not trigger the watch – but as I said, I guess t is just me doing something wrong here. That's why I'd be really thankful if you could provide some guidance or point me to the right places in the vue code or test, so I can dig around for myself... even though this is early stage I'd like to get this to work and provide an example others could use to figure out, how ssr and hydration will work :)

Thanks again!

P.S.: In case you would like to try the example, start the server with `node run.js` in the commits-ssr example directory.

dennisreimann:
Thanks to @moerssl I got it working now - as I guessed it was just me not understanding fully how it is supposed to work. I'll close this PR for now as we are extracting the example into a separate repository. I'll point to it once this repo exists :)

Issue #2872, opened 2016-05-13
Tests


yyx990803:
[![Coverage](https://img.shields.io/codecov/c/github/vuejs/vue/next.svg)](https://codecov.io/github/vuejs/vue?branch=next)

When contributing, please ensure the `npm run test:cover` task passes. During development you can also run the tests in continuous mode with `npm run dev:test` (requires Chrome to be installed on system).

## E2E Tests

- [x] commits
- [x] grid
- [x] markdown
- [x] modal
- [x] select2
- [x] svg
- [x] todomvc
- [x] tree

### Global config

- [x] Vue.config.preserveWhitespace
- [x] Vue.config.silent
- [x] Vue.config.errorHandler

### Global API

- [x] Vue.extend
- [x] Vue.nextTick
- [x] Vue.set
- [x] Vue.delete
- [x] Vue.directive
- [x] Vue.component
- [x] Vue.use
- [x] Vue.mixin
- [x] Vue.compile
- [x] Vue.filter
- [x] Vue.transition

##### data

- [x] data
- [x] props
  - [x] prop validation
    - [x] type
    - [x] required
    - [x] custom validator
  - [x] default value
- [x] propsData
- [x] computed
- [x] methods
- [x] watch

##### DOM

- [x] el
- [x] template
- [x] render

##### Lifecycle Hooks

- [x] init
- [x] created
- [x] beforeMount
- [x] mounted
- [x] beforeUpdate
- [x] updated
- [x] beforeDestroy
- [x] destroyed

##### Assets

- [x] directives
- [x] components
- [x] filters
- transitions (tested in dedicated transition tests)

##### Misc

- [x] parent
- [x] mixins
- [x] name
- [x] extends
- [x] delimiters

### Instance Properties

- [x] vm.$data (tested in options - data)
- [x] vm.$el (tested in options - el and $mount)
- [x] vm.$options
- [x] vm.$parent (tested in options - parent)
- [x] vm.$root
- [x] vm.$children
- [x] vm.$refs (tested in v-ref)
- [x] vm.$slots (tested in component slots)
- [x] vm.$isServer (tested in SSR)

##### data

- [x] vm.$watch

##### events

- [x] vm.$on
- [x] vm.$once
- [x] vm.$off
- [x] vm.$emit

##### Lifecycle

- [x] vm.$mount
- [x] vm.$destroy
- [x] vm.$forceUpdate
- [x] vm.$nextTick

### Directives

- [x] v-text
- [x] v-html
[...]

PR #2882, opened 2016-05-16
add optimizer tests

kazupon:
ref: #2872


yyx990803:
I think we should add test cases with some more complex template structures (e.g. with nesting like `static -> dynamic -> static`)

kazupon:
thanks review!!
I'll try to add more test cases.

kazupon:
Done

PR #2885, opened 2016-05-16
extract style/class/transition as web platform compiler modules

Jinjiang:
Currently they are written and treated as special attributes. This pr made them possible decoupled from web platform implementations.

And each platform will custom more attributes as platform compiler modules in the future.

Updates:

* added `modules?: Array<ModuleOptions>` in `CompilerOptions` type
* added `src/platforms/web/compiler/modules/*`

Changes:

* `parser`
  * `processClassBinding(element)`
  * `processStyleBinding(element)`
  * `processTransition(element)`
* `optimizer`
  * `isStaticKey`
* `codegen`
  * `genData(el)`

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -12,7 +12,7 @@ declare type CompilerOptions = {
 }
 
 declare type ModuleOptions = {
-  staticKeys: Array<string>,
+  staticKeys?: Array<string>,
   parse: Function,
   genData: Function
 }
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -2,12 +2,19 @@ declare type CompilerOptions = {
   warn?: Function,
   expectHTML?: boolean,
   preserveWhitespace?: boolean,
+  modules?: Array<ModuleOptions>,
   directives?: { [key: string]: Function },
   isUnaryTag?: (tag: string) => ?boolean,
   isReservedTag?: (tag: string) => ?boolean,
   mustUseProp?: (attr: string) => ?boolean,
   getTagNamespace?: (tag: string) => ?string,
   delimiters?: [string, string]
+}
+
+declare type ModuleOptions = {
+  staticKeys: Array<string>,
+  parse: Function,
+  genData: Function
 }
 
 declare type ASTElementHandler = {
--- a/src/compiler/codegen.js
+++ b/src/compiler/codegen.js
@@ -121,7 +121,7 @@ function genData (el: ASTElement): string {
   }
   // platform modules
   platformModules.forEach(module => {
-    data = module.genData(el, data)
+    data += module.genData(el) || ''
   })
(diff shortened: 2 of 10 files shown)


flow/compiler.js
```diff
@@ -10,6 +11,12 @@ declare type CompilerOptions = {
   delimiters?: [string, string]
 }
 
+declare type ModuleOptions = {
+  staticKeys: Array<string>,
```
yyx990803: can make `staticKeys` optional here.

src/compiler/codegen.js
```diff
@@ -117,21 +119,10 @@ function genData (el: ASTElement): string {
...
-  if (el.styleBinding) {
-    data += `style:${el.styleBinding},`
-  }
-  // transition
-  if (el.transition) {
-    data += `transition:{definition:(${el.transition}),appear:${el.transitionOnAppear}},`
-  }
+  // platform modules
+  platformModules.forEach(module => {
+    data = module.genData(el, data)
```
yyx990803: I think it's better to do `data += module.genData(el)` here. `genData` for each module doesn't need existing `data` as an argument.

flow/compiler.js
Jinjiang: fixed

src/compiler/optimizer.js
```diff
@@ -17,13 +18,27 @@ let isPlatformReservedTag
  */
 export function optimize (root: ?ASTElement, options: CompilerOptions) {
   if (!root) return
+  isStaticKey = genStaticKeys(options.modules)
```
yyx990803: We are generating a new map every time `optimize` is called. Wondering if there's a way to cache this since most of the time it's going to be the same map.

src/compiler/codegen.js
Jinjiang: OK, done.

src/compiler/optimizer.js
Jinjiang: How about generate a string option `staticKeys` directly in `CompilerOptions` in `web-compiler.js`?

src/entries/web-compiler.js
```diff
@@ -16,6 +17,8 @@ const cache2: { [key: string]: CompiledFunctions } = Object.create(null)
 const baseOptions: CompilerOptions = {
   expectHTML: true,
   preserveWhitespace: true,
+  modules,
+  staticKeys: genStaticKeys(modules),
```
yyx990803: actually this option can be a function called `isStaticKey` directly... no need to cache it!

Jinjiang: If we do this the basic static keys `'type,tag,attrsList,attrsMap,plain,parent,children,staticAttrs'` also put into the entry file. That's not good from my opinion.

PR #2886, opened 2016-05-17
revert _init method for Vuex

Jinjiang:


src/core/instance/init.js
```diff
@@ -9,35 +9,38 @@ import { mergeOptions } from '../util/index'
...
-    initInternalComponent(vm, options)
-  } else {
-    vm.$options = mergeOptions(
-      vm.constructor.options,
-      options || {},
-      vm
-    )
+export function initMixin (Vue: Class<Component>) {
+  Vue.prototype._init = function (options?: Object) {
+    const vm = this
```
Jinjiang: Oops I forgot the type `const vm: Component = this`...

yyx990803: I'll add it

PR #2889, opened 2016-05-17
Unit tests for Data

brendanbowidas:
Please let me know if I missed any possible test cases, this is my first open source PR
:)

### DIFF
--- a/dist/compiler.js
+++ b/dist/compiler.js
@@ -1,58 +1,6 @@
 'use strict';
 
-var config = {
-
-  /**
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks for the PR - can you avoid checking in the dist files? They should only be checked in during releases.

yyx990803:
Also there are a few things for testing options that need to be covered, I'd appreciate if you can hold off until there are some examples for option testing. Thanks!

PR #2892, opened 2016-05-17
added ref directive tests

blake-newman:


test/unit/features/directives/ref.spec.js
```diff
@@ -0,0 +1,93 @@
...
+    vm.$mount()
+    expect(vm.$refs.test.$options.id).toBe('test')
+    vm.test = 'test2'
+    waitForUpdate(() => {
+      expect(vm.$refs.test.$options.id).toBe('test2')
+      vm.test = ''
+    }).then(() => {
+      expect(vm.$refs.test).toBeUndefined()
+      done()
+    })
```
kazupon: `}).catch(done)`

blake-newman: Good spot. I'll amend

yyx990803: ping @blake-newman 

Note: use the new `.then(done)` syntax.

blake-newman: Removed keep-alive test due to it currently being depreciated. 

Changed the structure to use `.then(done)` syntax. 

All tests are now receiving this error once rebasing from `next`.

```
Chrome 50.0.2661 (Mac OS X 10.11.4) Directive v-ref should accept dynamic component FAILED
	SyntaxError: missing ) after argument list
	    at compileToFunctions (/Users/blake/work/vue/vue/test/unit/index.js:5251:17 <- webpack:///src/entries/web-compiler.js:68:15)
	    at Vue._webRuntime2.default.$mount (/Users/blake/work/vue/vue/test/unit/index.js:117:70 <- webpack:///src/entries/web-runtime-with-compiler.js:35:61)
	    at Object.<anonymous> (/Users/blake/work/vue/vue/test/unit/index.js:8457:9 <- webpack:///test/unit/features/directives/ref.spec.js:62:7)
```

posva: Shouldn't it actually be `then(done, done)` in case there's an error in tests? (first done is called with the return value as a normal then and second done is called with the error as an argument if there's an exception)

PR #2894, opened 2016-05-18
add codegen tests

kazupon:
ref #2872

I think that is not the best about codegen checking because it is hard coding.
Are there any good idea ?

I think that there are following about codegen checking:
- helper method for codegen test (e.g. genElement)


yyx990803:
Seems it would be a good idea to extract a helper function to reduce the duplicate code for each test. e.g. `assertCodegen(template, generatedCode, [staticRenderFnCode, ...])`

kazupon:
Thanks you advice!

kazupon:
refactoring done

PR #2935, opened 2016-05-22
add patch tests

kazupon:
ref: #2872

### DIFF
--- a/test/unit/modules/vdom/modules/transition.spec.js
+++ b/test/unit/modules/vdom/modules/transition.spec.js
@@ -1,0 +1,23 @@
+import Vue from 'vue'
+import { basePatch as patch } from 'web/runtime/patch'
+import VNode from 'core/vdom/vnode'
+import { inBrowser } from 'core/util/index'
+
+if (inBrowser) {
+  describe('trasition module', () => {
+    it('should transit with basic transition', () => {
+      const vm = new Vue()
+      // create
+      const vnode1 = new VNode('div', {}, [
+        new VNode('p', { transition: { definition: 'expand', appear: true }},
+          undefined, undefined, undefined, undefined, vm
+        )
+      ])
+      let elm = patch(null, vnode1)
+      expect(elm.childNodes[0].classList.contains('expand-enter')).toBe(true)
+      // remove
+      const vnode2 = new VNode('div', {})
+      elm = patch(vnode1, vnode2)
+    })
+  })
+}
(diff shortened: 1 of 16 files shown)


Twiknight:
Thanks for the work.

I prefer moving `module` specialized cases to separated files ( attrs, classes, etc).

All `patch` need to do is to call their hooks at the right time.
A case suit for `patch` is not responsible for their correctness.

test/unit/modules/vdom/patch.spec.js
```diff
@@ -0,0 +1,986 @@
...
+import { createComponent } from 'core/vdom/create-component'
+import { createPatchFunction } from 'core/vdom/patch'
+import baseModules from 'core/vdom/modules/index'
+import * as nodeOps from 'web/runtime/node-ops'
+import platformModules from 'web/runtime/modules/index'
+import VNode from 'core/vdom/vnode'
+
+function prop (name) {
+  return obj => { return obj[name] }
+}
```
yyx990803: blank lines between functions please

yyx990803:
Agree with @Twiknight that this file is a bit too big - let's split it into multiple files (similar to how snabbdom's tests are).

kazupon:
Thanks review!
I'll try to split test cases into multiple files.

kazupon:
Done

test/unit/modules/vdom/modules/transition.spec.js
```diff
@@ -0,0 +1,23 @@
...
+      const vnode1 = new VNode('div', {}, [
+        new VNode('p', { transition: { definition: 'expand', appear: true }},
+          undefined, undefined, undefined, undefined, vm
+        )
+      ])
+      let elm = patch(null, vnode1)
+      expect(elm.childNodes[0].classList.contains('expand-enter')).toBe(true)
+      // remove
+      const vnode2 = new VNode('div', {})
+      elm = patch(vnode1, vnode2)
```
yyx990803: is this complete?

test/unit/modules/vdom/patch/hydration.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    const node0 = createServerRenderedDOM()
+    const vnode1 = new VNode('div', {}, [
+      new VNode('span', {}),
+      new VNode('div', { hook: { init }}, [
+        new VNode('span', {}),
+        new VNode('span', {})
+      ])
+    ])
+    patch(node0, vnode1)
+    expect(result.length).toBe(1)
```
yyx990803: should also assert that each `VNode`'s `elm` property is correctly set to the corresponding DOM element.

PR #2936, opened 2016-05-22
added test for style binding

Jinjiang:


yyx990803:
Thanks! Merged #2932 instead of this because:

- #2392 was submitted first
- @brendanbowidas is a first time contributor and we want to be more inclusive!
- #2392 has better coverage.

Hope you don't mind!

PR #2948, opened 2016-05-24
Delimiters tests

bartlomieju:
ref #2872

I think it should be somehow emphasized that if we want delimiters to be changed globally we should define Vue.options.delimiters. Doing so in a constructor of our root instance will not have effect (check out 'delimiters should only affect vm' spec). I guess this is desirable, for most of users rely on default delimiters, and we want to be able to use 3rd party plugins.


yyx990803:
Great! Thanks for the thorough tests.

PR #2950, opened 2016-05-25
Add halt event modifier.

knpwrs:
In my experience with JavaScript it is *very* common to call `stopPropagation` and `preventDefault` in tandem. This pull request adds a modifier which is a shortcut for both. Just a few notes:

1.) I am not attached to the name `halt`.
2.) If this looks good I can go ahead and add docs.

Possible alternative implementation:

```js
    if (this.modifiers.stop || this.modifiers.halt) {
      handler = stopFilter(handler)
    }
    if (this.modifiers.prevent || this.modifiers.halt) {
      handler = preventFilter(handler)
    }
```

Or, since `this.modifiers` is called a whole bunch:

```js
    var mods = this.modifiers
    if (mods.stop || mods.halt) {
      handler = stopFilter(handler)
    }
    if (mods.prevent || mods.halt) {
      handler = preventFilter(handler)
    }
```

### DIFF
--- a/src/directives/public/on.js
+++ b/src/directives/public/on.js
@@ -107,6 +107,9 @@ export default {
     if (this.modifiers.prevent) {
       handler = preventFilter(handler)
     }
+    if (this.modifiers.halt) {
+      handler = stopFilter(preventFilter(handler))
+    }
     if (this.modifiers.self) {
       handler = selfFilter(handler)
     }
--- a/test/unit/specs/directives/public/on_spec.js
+++ b/test/unit/specs/directives/public/on_spec.js
@@ -214,6 +214,27 @@ describe('v-on', function () {
     expect(window.location.hash).toBe(hash)
   })
 
+  it('halt modifier', function () {
+    var prevented
+    var outer = jasmine.createSpy('outer')
+    new Vue({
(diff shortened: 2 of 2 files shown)


LinusBorg:
Hi @knpwrs, 

thank you for oyur contribution. It certainly seems to add a useful shortcut, albeight a minor one. Anyone else have an opinion on this?

Anyways, please give us some time to consider this, as we are currently working mostly on shipping 2.0

simplesmiler:
Give https://davidwalsh.name/javascript-events and https://css-tricks.com/dangers-stopping-event-propagation/ a good read or two before stopping propagation.

knpwrs:
That David Walsh article indicates that should this feature be accepted that there should probably be documentation stating to be careful when handling events in libraries intended for distribution.

yyx990803:
Thanks for the PR, but I am somehow not in favor of this, because there is no corresponding semantics in the native DOM event API. When I see `.stop`, `.prevent` or `.capture`, I can relate them to what is actually happening, but `.halt` just feels less explicit than `.stop.prevent`. I think this is the case where I'd prefer explicitness over typing a few less character.

PR #2959, opened 2016-05-26
tests for parent, name and mixins

bartlomieju:
ref #2872

### DIFF
--- a/test/unit/features/options/mixins.spec.js
+++ b/test/unit/features/options/mixins.spec.js
@@ -1,0 +1,77 @@
+import Vue from 'vue'
+const mergeOptions = Vue.util.mergeOptions
+
+describe('Mixins', () => {
+  it('vm should have options from mixin', () => {
+    const mixin = {
+      directives: {
+        c: {}
+      },
+      methods: {
+        a: function () {}
+      }
+    }
+
+    const vm = new Vue({
+      mixins: [mixin],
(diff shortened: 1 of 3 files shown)


test/unit/features/options/mixins.spec.js
```diff
@@ -0,0 +1,77 @@
...
+        b: function () {}
+      }
+    }).$mount()
+
+    expect(vm.a).toBeDefined()
+    expect(vm.b).toBeDefined()
+    expect(vm.$options.directives.c).toBeDefined()
+  })
+
+  it('should call hooks from mixins first, mixin methods should not override defined method', () => {
```
yyx990803: Can we split this into two separate cases? One for hook order and the other for method overriding.

test/unit/features/options/mixins.spec.js
```diff
@@ -0,0 +1,77 @@
+import Vue from 'vue'
+const mergeOptions = Vue.util.mergeOptions
+
+describe('Mixins', () => {
```
yyx990803: Let's prefix option test suites with Options. So it would be `describe('Options mixin', ...)`

yyx990803:
Thanks for the PR! Just a few small things to be adjusted and LGTM after that.

bartlomieju:
Adjusted to your guidelines 👍

PR #2965, opened 2016-05-27
add tests for lifecycle hooks

brendanbowidas:
ref #2872 
Please let me know if I missed any test cases and I will revise. I was't quite sure how you wanted these structured within the larger "lifecycle" test suite.

Not sure why CircleCi failed, looks like there are some eslint errors elsewhere in the codebase

### DIFF
--- a/test/unit/features/options/lifecycle.spec.js
+++ b/test/unit/features/options/lifecycle.spec.js
@@ -1,0 +1,37 @@
+import Vue from 'vue'
+
+describe('Lifecycle API', () => {
+  it('hooks', () => {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR - but I'm afraid we need a much more thorough test suite than this. We need to assert that each hook works properly for its intended purpose - e.g. `init` should allow you to customize the options, `created` should have set up data observation, etc. It could be a bit complicated to explain since that involves the internal design (compared to some other options where their behavior is more obvious), so I think it's probably better to let me write tests.

zzz6519003:
@brendanbowidas  test can you see my reaction as a notification? XD

PR #2966, opened 2016-05-27
add options render tests

kazupon:
ref: #2872

### DIFF
--- a/test/unit/features/options/render.spec.js
+++ b/test/unit/features/options/render.spec.js
@@ -1,0 +1,30 @@
+import Vue from 'entries/web-runtime'
+
+describe('Options render', () => {
+  it('basic usage', () => {
+    const vm = new Vue({
+      render () {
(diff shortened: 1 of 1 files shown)


test/unit/features/options/render.spec.js
```diff
@@ -0,0 +1,31 @@
...
+      }
+    }).$mount()
+    expect(vm.$el.tagName).toBe('UL')
+    for (let i = 0; i < vm.$el.children.length; i++) {
+      const li = vm.$el.children[i]
+      expect(li.tagName).toBe('LI')
+      expect(li.textContent).toBe(vm.items[i].name)
+    }
+  })
+
```
yyx990803: Let's also add a case where the render function is using `this.$createElement`:

``` js
render () {
  const h = this.$createElement
  return h('ul', ...)
}
```

PR #2967, opened 2016-05-27
add options el tests

kazupon:
ref: #2872


yyx990803:
Looks like there's one case failing in PhantomJS

PR #2968, opened 2016-05-27
add options template tests

kazupon:
ref: #2872

### DIFF
--- a/src/entries/web-runtime-with-compiler.js
+++ b/src/entries/web-runtime-with-compiler.js
@@ -26,7 +26,10 @@ Vue.prototype.$mount = function (el: string | Element): Component {
       } else if (template.nodeType) {
         template = template.innerHTML
       } else {
-        warn('invalid template option:' + template, this)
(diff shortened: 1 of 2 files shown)


src/entries/web-runtime-with-compiler.js
```diff
@@ -26,7 +26,10 @@ Vue.prototype.$mount = function (el: string | Element): Component {
       } else if (template.nodeType) {
         template = template.innerHTML
       } else {
-        warn('invalid template option:' + template, this)
+        if (process.env.NODE_ENV !== 'production') {
+          warn('invalid template option:' + template, this)
+        }
+        return mount.call(this, el)
```
kazupon: if specify the invalid template, there is a possibility that runtime error in render process.
I don't know whether this approach is correct.

yyx990803: I think just `return this` is fine.

kazupon: thanks advice :)
I updated.

PR #2971, opened 2016-05-29
do not call shift if the key already exists

leemotive:
if the cache size is equal to cache limit and now we put a new entry with an existed key (not the head key), the result is we removed the head entry and updated an existed entry. the cache size will be limit-1. so we need to move the if-statement inside

### DIFF
--- a/src/cache.js
+++ b/src/cache.js
@@ -33,12 +33,12 @@ var p = Cache.prototype
 
 p.put = function (key, value) {
   var removed
-  if (this.size === this.limit) {
-    removed = this.shift()
-  }
 
   var entry = this.get(key, true)
   if (!entry) {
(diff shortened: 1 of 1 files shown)


src/cache.js
```diff
@@ -33,12 +33,12 @@ var p = Cache.prototype
...
   var removed
-  if (this.size === this.limit) {
-    removed = this.shift()
-  }
 
   var entry = this.get(key, true)
   if (!entry) {
+    if (this.size === this.limit) {
+      removed = this.shift()
+    }
```
LinusBorg: I the outer if could be avoided if we moved the inner after the entry was added/updated.

then:
* if entry did alread exist, size will still be the same
* if entry is a new one, size will be over the limit, and we can remove the head.

PR #2981, opened 2016-05-31
Fix package.json files section

lepture:
Just a small fix.


yyx990803:
Thanks, this is actually intended - the current build script outputs `compiler.js` but the dist folder has not been updated yet.

PR #2989, opened 2016-06-01
Fixed #2988

### LINKED ISSUE #2988: [BUG] one time binding doesn't deal with null like value properly
### Vue.js version
1.0.24

### Reproduction Link
http://jsbin.com/yesozewezu/1/edit?html,js,output

### Steps to reproduce
For one time binding, undefined value would display as "undefined", it's different from two-way binding.
### What is Expected?
Should display as empty.
Display undefined.

dsonet:
Make rendering of undefined and null in one-time binding mode as same as reactive binding.

### DIFF
--- a/src/compiler/compile.js
+++ b/src/compiler/compile.js
@@ -5,6 +5,7 @@ import { parseText, tokensToExp } from '../parsers/text'
 import { parseDirective } from '../parsers/directive'
 import { parseTemplate } from '../parsers/template'
 import {
+  _toString,
   resolveAsset,
   toArray,
   warn,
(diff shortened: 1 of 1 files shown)


LinusBorg:
Hello dsonet, 

Thank you very much for your bug report and this pull request with a suggested fix. 

We will review it and come back to you shortly..

src/compiler/compile.js
```diff
@@ -446,7 +446,7 @@ function makeTextNodeLinkFn (tokens, frag) {
           if (token.html) {
             replace(node, parseTemplate(value, true))
           } else {
-            node.data = value
+            node.data = value === undefined ? '' : value
```
yyx990803: Let's use [this function](https://github.com/vuejs/vue/blob/dev/src/util/lang.js#L114) for consistency with normal bindings.

dsonet: Sure.

PR #3030, opened 2016-06-03
add ssr custom directive test

kazupon:
ref: #2872

### DIFF
--- a/test/ssr/ssr-string.spec.js
+++ b/test/ssr/ssr-string.spec.js
@@ -3,8 +3,6 @@ import { compileToFunctions } from '../../dist/compiler.js'
 import createRenderer from '../../dist/server-renderer.js'
 const { renderToString } = createRenderer()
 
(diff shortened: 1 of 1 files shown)


test/ssr/ssr-string.spec.js
```diff
@@ -360,6 +358,39 @@ describe('SSR: renderToString', () => {
     })
   })
 
+  it('custom directives', done => {
+    const custom = jasmine.createSpy()
```
yyx990803: We should implement some mock behavior of manipulating the VNode and assert the result. (see https://github.com/vuejs/vue/tree/next/src/platforms/web/server/directives)

kazupon: thanks!
I'll try to fix it. :)

kazupon: Done

yyx990803:
👍

PR #3034, opened 2016-06-04
remove a needless if.

dsonet:
Since always remove the old class.


posva:
Thanks, I overlooked that!

Issue #3054, opened 2016-06-11
[2.0.0-alpha] A bug with inline templates


acasar:
### Vue.js version
2.0.0.-alpha

### Reproduction Link
https://jsfiddle.net/bd1wfuht/

### Steps to reproduce
Click toggle

### What is Expected?
"true" should be displayed in bold

### What is actually happening?
Uncaught TypeError: Cannot read property '0' of null.

Note: If you remove ``<strong>`` element, it all works correctly. So there is a problem with nested nodes inside v-if in inline templates.

simplesmiler:
Looks like the problem is with `v-else`: https://jsfiddle.net/bd1wfuht/2/

simplesmiler:
Should be fixed with https://github.com/vuejs/vue/pull/3063
Feel free to reopen if the issue will persist in the next alpha release.

PR #3063, opened 2016-06-12
Fix options for inline-template instances

simplesmiler:
Should fix https://github.com/vuejs/vue/issues/3054


yyx990803:
Ouch, good catch!

PR #3064, opened 2016-06-13
refactor: `.test` is faster than `.match` for regex bools

maraisr:
Please refer to this [stack](http://stackoverflow.com/questions/10940137/regex-test-v-s-string-match-to-know-if-a-string-matches-a-regular-expression) for my claim that .test is faster than .match :)

Love your work Evan!

### DIFF
--- a/src/core/util/env.js
+++ b/src/core/util/env.js
@@ -15,7 +15,7 @@ export const devtools = inBrowser && window.__VUE_DEVTOOLS_GLOBAL_HOOK__
 // UA sniffing for working around browser-specific quirks
 const UA = inBrowser && window.navigator.userAgent.toLowerCase()
 const isIos = UA && /(iphone|ipad|ipod|ios)/i.test(UA)
-const iosVersionMatch = UA && isIos && UA.match(/os ([\d_]+)/)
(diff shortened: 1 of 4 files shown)


src/core/util/env.js
```diff
@@ -15,7 +15,7 @@ export const devtools = inBrowser && window.__VUE_DEVTOOLS_GLOBAL_HOOK__
 // UA sniffing for working around browser-specific quirks
 const UA = inBrowser && window.navigator.userAgent.toLowerCase()
 const isIos = UA && /(iphone|ipad|ipod|ios)/i.test(UA)
-const iosVersionMatch = UA && isIos && UA.match(/os ([\d_]+)/)
+const iosVersionMatch = UA && isIos && /os ([\d_]+)/.test(UA)
```
yyx990803: this line must use `.match` because the next line depends on it for version sniffing.

maraisr: Ahh sorry! Totally misread it's context! All the tests ran fine though? Anyhow! I try! Thanks again for this amazing library!

yyx990803: Because the CI only runs the tests in PhantomJS but this line is targeted at iOS browsers. The other two changes are fine though.

PR #3101, opened 2016-06-18
Fixed: the "pluralize" filter with multi arguments, when selected arg…

nervgh:
In Vue.js `1.0.25` the "pluralize" filter returns "three"
```js
Vue.filter('pluralize')(1, "", "two", "three") // three
```
instead of the empty string. After this fix, the "pluralize" filter returns the empty string:
```js
Vue.filter('pluralize')(1, "", "two", "three") // ""
```


yyx990803:
Thanks!

Issue #3133, opened 2016-06-22
Multiple vue instance bug


douzi8:
### Vue.js version
 Vue.js v2.0.0-alpha.5

### Reproduction Link
https://jsbin.com/xedizivigi/edit?html,js,console,output

### Steps to reproduce
First read  'bar.car.log', then assign  bar.car.log = 'baoma'

### What is Expected?
foo object will not update

### What is actually happening?
foo object  will also be updated

pespantelis:
Hi @douzi8,

I have checked your bin and the `foo` object remains the same as expected.

simplesmiler:
Interesting, looks like a bug to me.

ealves-pt:
It's working as expected for me, `bar.car.log` gets a new value assigned and `foo.car.log` is not changed.

simplesmiler:
@ealves-pt but `update` is still called for `foo`, which should not be happening.

ealves-pt:
@simplesmiler hmmm indeed didn't noticed that. Interesting...

PR #3136, opened 2016-06-22
Fix watcher target not resetting properly

simplesmiler:
Possible fix for https://github.com/vuejs/vue/issues/3133.

The problem appears to be rooted in [`prevTarget`](https://github.com/vuejs/vue/blob/v2.0.0-alpha.5/src/core/observer/watcher.js#L15) getting funky [during instance mounting](https://github.com/vuejs/vue/blob/v2.0.0-alpha.5/src/core/instance/lifecycle.js#L53). This causes `Dep.target` to never be reset to `null`.

This is most likely not the best solution. Maybe a better place for solution would be around [here](https://github.com/vuejs/vue/blob/v2.0.0-alpha.5/src/core/instance/lifecycle.js#L54) or [here](https://github.com/vuejs/vue/blob/v2.0.0-alpha.5/src/core/observer/watcher.js#L77).

Also I can't come up with a better unit test than the original repro:
- Create two instances, first with a computed property (easiest way to reproduce), make sure computed field is used in the template,
- Change the data of the second instance,
- Make sure that first instance didn't update.

Should I include it anyway?

/cc @yyx990803

### DIFF
--- a/src/core/observer/watcher.js
+++ b/src/core/observer/watcher.js
@@ -204,10 +204,12 @@ export default class Watcher {
   evaluate () {
     // avoid overwriting another watcher that is being
     // collected.
-    const current = Dep.target
+    const currentPrevTarget = prevTarget
+    prevTarget = Dep.target
     this.value = this.get()
     this.dirty = false
-    Dep.target = current
+    Dep.target = prevTarget
+    prevTarget = currentPrevTarget
   }
 
   /**


simplesmiler:
Checking why e2e is failing.

simplesmiler:
Can not run `node test/e2e/runner.js --env phantomjs` on my windows machine.
Running `node test/e2e/runner.js --env chrome` yields no failures.

Would appreciate a hint.

simplesmiler:
Hm, maybe I should not be fiddling with prev target, and should be just saving it along with the current target.

yyx990803:
e2e is most likely due to CI unstableness. Thanks a lot for discovering the root cause tho! Indeed I think there's a better way to fix this and I've pushed it here: https://github.com/vuejs/vue/commit/2c63e94f692430d3ff48cec31457a27684b085a2

simplesmiler:
@yyx990803 :+1:

PR #3149, opened 2016-06-24
Fix bug: directive sort is not stable on Chrome

furybean:
I'm building a tabs component. It is used like this:

```
<tabs>
  <tab title="1">Tab1</tab>
  <tab title="2">Tab2</tab>
  <tab title="3">Tab3</tab>
  <tab title="4">Tab4</tab>
  <tab title="5">Tab5</tab>
  <tab title="6">Tab6</tab>
  <tab title="7">Tab7</tab>
  <tab title="8">Tab8</tab>
  <tab title="9">Tab9</tab>
  <tab title="10">Tab10</tab>
</tabs>
```

[Live demo(Works fine)](https://jsfiddle.net/sgd2sLqf/4/)

It works fine when there are less than 10 ```<tab>s```. But tabs will disorder when I have more than 10 ```<tab>s```.

[Live demo(Bug)](https://jsfiddle.net/sgd2sLqf/5/)

The main reason is that the directive sorting by priority is not stable on Chrome.

I fixed this bug using a sort method(time: O(n), space: O(n)),  It also can be fixed using a stable sort method such as merge sort.

### DIFF
--- a/src/compiler/compile.js
+++ b/src/compiler/compile.js
@@ -105,7 +105,7 @@ function linkAndCapture (linker, vm) {
   var originalDirCount = vm._directives.length
   linker()
   var dirs = vm._directives.slice(originalDirCount)
-  dirs.sort(directiveComparator)
+  sortDirectives(dirs)
   for (var i = 0, l = dirs.length; i < l; i++) {
     dirs[i]._bind()
   }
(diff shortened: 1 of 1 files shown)


yyx990803:
多谢，我发布 1.0.26 的时候忘记先 merge 这个了...

PR #3168, opened 2016-06-28
[2.0][suggestion] pass v-on event handlers on functional component

kazupon:
Sometimes, there are times when we pass v-on event handlers to `render` function arguments on functional component.

The below URL is implemented example with v-bind on functional component.
http://jsfiddle.net/kazupon/bjp40byw/6/

In example, we need to define handler in `props`. 
I think that come in useful by v-on event handers passing through.

### DIFF
--- a/src/core/vdom/create-component.js
+++ b/src/core/vdom/create-component.js
@@ -71,9 +71,10 @@ export function createComponent (
   if (Ctor.options.functional) {
     return Ctor.options.render.call(
       null,
-      parent.$createElement,      // h
-      propsData || {},            // props
-      normalizeChildren(children) // children
+      parent.$createElement,       // h
+      propsData || {},             // props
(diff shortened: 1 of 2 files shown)


dsonet:
```javascript
var on = { valid: props.onValid, invalid: props.onInvalid }
 var vnode = h('validate-input-control', { props: props, on: on })
```
why gather the properties from props since there is props passed already? Seems duplicated.

kazupon:
In virtual-dom system of vue 2.0, when we are necessary in order to communicate the parent-child components, need to use the `on`.

yyx990803:
I think it's more straightforward to have only `props` as the second argument. We can pass the raw `data` as the 4th argument, since it's going to be rarely used, only for advanced users.

kazupon:
That's sounds good.
I'll try to fix now.

PR #3177, opened 2016-06-29
fix compiler options

kazupon:
I found the compiler options issue.

The below the reproduction minimum codes.
https://jsfiddle.net/kazupon/dfok8du9/3/

### DIFF
--- a/src/platforms/web/compiler/index.js
+++ b/src/platforms/web/compiler/index.js
@@ -26,10 +26,18 @@ export function compile (
   template: string,
   options?: CompilerOptions
 ): CompiledResult {
-  options = options
+  options = options || {}
+  const newOptions = options
     ? extend(extend({}, baseOptions), options)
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks - this is intentional, because `Vue.compile` is an alias to `compileToFunctions`, which is only used at runtime. But runtimes already include pre-determined modules and directives, and you should not be able to compile the template in a different way.

The test for compile-time modules and directives should be in `test/modules/compiler/compiler-options.spec.js` and should be using the `compile` function exported by `src/entries/web-compiler`.

kazupon:
OK! I gotcha!

>The test for compile-time modules and directives should be in test/modules/compiler/compiler-options.spec.js and should be using the compile function exported by src/entries/web-compiler.

That make sense!
I think this test helpful when the user want to user the compiler.
I'll try to push the other PR. 😉

PR #3181, opened 2016-06-29
add compile options test

kazupon:
reference: #3177


test/unit/modules/compiler/compiler-options.spec.js
```diff
@@ -0,0 +1,117 @@
...
+            field:${JSON.stringify(el.validate.field)},
+            groups:${JSON.stringify(el.validate.groups)},
+            validators:${JSON.stringify(el.validators)},
+            result:${JSON.stringify(result)},
+            child:${code}}
+          })`
+        }
+      }]
+    })
+    expect(render).not.toBeUndefined()
```
zzz6519003: x.toBeDefined would be better

Issue #3183, opened 2016-06-29
Props validation fails (Vue warn) when component initialized with proper propsData


iagafonov:
I made a universal component (modal) that can be used as inline component
`<modal :entity="some">`
or via service
`this.$modal.open({propsData: {...}, component: ModalForm})`
it works fine, but warns about invalid props, because it runs in `compileProps` step, and propsData will be used later in `makePropsLinkFn` step (compile-props.js).
I think propsData should also be a source of validation along with element attributes

version 1.0.26 [jsFilddle](https://jsfiddle.net/oe7axeab/1296/) (see console)
version 2.0.0-alpha.8 does not have this problem

kazupon:
Close (Fixed in #3193)

PR #3193, opened 2016-06-30
fix props validation fails (#3183)

kazupon:


src/compiler/compile-props.js
```diff
@@ -122,6 +123,9 @@ export function compileProps (el, propOptions, vm) {
     } else if ((value = getAttr(el, attr)) !== null) {
       // has literal binding!
       prop.raw = value
+    } else if (propsData && (value = propsData[attr]) !== null) {
```
yyx990803: `attr` is hyphenated, but the user may use camelCase prop names in `propsData`. We need to check  both `propsData[attr]` and `propsData[name]`.

iagafonov: Maybe user can **only** use camelCase propsData[path]?

yyx990803: True, that makes more sense. then we should just replace `attr` with `name` here.

kazupon:
Thanks review!
I fixed.

PR #3202, opened 2016-07-01
textarea with v-pre attr should skip compile

### LINKED ISSUE #3184: v-pre on textarea breaks sometimes
### Vue.js version
1.0.26

### Reproduction Link
https://jsfiddle.net/simplesmiler/Lrk9sxjf/102/

### What is Expected?
Both textareas having text inside them.

First textarea has text. Second textarea is empty, and has literal attribute `:value`.

hylin:
skip compile before process it's :value when a textarea has v-pre attr
Fixed #3184


yyx990803:
Thanks!

Issue #3210, opened 2016-07-02
Parsing object as a filter argument fails with spaces inside


sqal:
### Vue.js version
1.0.26

### Reproduction Link
https://jsfiddle.net/0o853fzd/

Take a look at example. You can see that when i try to pass object to the filter as an argument it doesn't work when there is some space in the string.

![Err](http://image.prntscr.com/image/8d9b80adc23744cf966a4f983b6d6c52.png)

Isn't supposed to PR #834 fix this issue? or is it forbidden to use spaces when passing obejct to the filter like this?

yyx990803:
This is because the 1.x filter syntax uses space as the argument delimiter. If anyone wants to work on this, the relevant logic is here: https://github.com/vuejs/vue/blob/dev/src/parsers/directive.js#L20-L35

defcc:
I am working on a simple expression parser to parse the filter syntax rather then the regex way。Is it an acceptable method ? If so I will make a pull request soon.

yyx990803:
@defcc code size is also a concern here. We don't want to bloat the code for a relatively rare use case.

defcc:
ok, got it. @yyx990803

PR #3214, opened 2016-07-03
Allow white spaces inside object literal.

zigomir:
Fix for https://github.com/vuejs/vue/issues/3210.

### DIFF
--- a/src/parsers/directive.js
+++ b/src/parsers/directive.js
@@ -2,7 +2,7 @@ import { toNumber, stripQuotes } from '../util/index'
 import Cache from '../cache'
 
 const cache = new Cache(1000)
-const filterTokenRE = /[^\s'"]+|'[^']*'|"[^"]*"/g
+const filterTokenRE = /(({.*})|([^\s'"]+|'[^']*'|"[^"]*"))/g
 const reservedArgRE = /^in$|^-?\d+/
 
 /**
--- a/src/parsers/directive.js
+++ b/src/parsers/directive.js
@@ -1,5 +1,6 @@
 import { toNumber, stripQuotes } from '../util/index'
 import Cache from '../cache'
+import { warn } from '../util/index'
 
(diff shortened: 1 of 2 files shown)


posva:
LGTM. I guess it's worth mentioning in the docs

yyx990803:
pure regex based approach is a somewhat incomplete fix. For example it would fail for this:

``` js
abc | filter {a:1} {b:2}
```

It would match `{a:1} {b:2}` as a single argument.

Additional, if we support spaces inside object braces, they should probably work for arrays too:

``` js
abc | filter [1, { a: 2 }]
```

So we either support the full syntax, or not support it at all. A more proper parsing logic would have to do something similar to what the directive parser is doing: keeping track of braces/brackets/quotes state and only delimit an argument if we are not in any pairs.

yyx990803:
Fortunately in 2.0 we don't have this issue :o

posva:
So maybe we should just drop it and recommend to pass a computed property instead?

zigomir:
@yyx990803 I agree. This isn't proper fix then. I'll see what I can do about it :)

zigomir:
@yyx990803 I was playing around a bit more and come up with new syntax when filter has multiple arguments. If there are more filters they need to be wrapped in `[]` as you can see in spec. I'm not sure if this is the way we want to do it...especially with eval used 😊 . 

This also forces us to document this behavior and update the docs/guide. I can do it, if this is the way we want to handle this, otherwise I think it's better we just document you should extract filters to functions or use no spaces like @posva already mentioned :)

posva:
We forgot about this. I'll rather keep it the way it is and add some documentation that adding a new syntax to filters which already use a special (non-js) syntax

posva:
Can this be closed now? I haven't followed well enough @zigomir

zigomir:
Yeah, for sure, it was just a silly attempt of mine :)

PR #3250, opened 2016-07-09
Enhance extend function

bowen31337:
copy the values of all enumerable own properties from one or more source objects to a target object.

### DIFF
--- a/src/util/lang.js
+++ b/src/util/lang.js
@@ -258,13 +258,25 @@ export function toArray (list, start) {
  * @param {Object} from
  */
 
-export function extend (to, from) {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR, however This function is perf critical, it is intentionally kept simple.

bowen31337:
@yyx990803  thanks! where can i find your code of standard or contribution manual for vuejs project? If I could have those in hand, I would not make any arbitrary PRs, so it would not waste your time viewing the PRs.

bowen31337:
@yyx990803 Thanks, I have just already found the contribution guide, I will not make any arbitrary PRs.

PR #3251, opened 2016-07-09
Enhance extend function

bowen31337:
copy the values of all enumerable own properties from one or more source objects to a target object


phanan:
Closing as per #3250.

PR #3258, opened 2016-07-11
[2.0] [vuejs:next] add "aside" as reservedTag

tommyZZM:
fix "[Vue warn]: Unknown custom element: `<aside>` - did you register the component correctly? ..."

`<aside>` is acturally a html element
Reference:https://developer.mozilla.org/en/docs/Web/HTML/Element/aside


yyx990803:
Thanks, there was a conflict so I manually added it in, hope you don't mind.

PR #3292, opened 2016-07-19
Handle back-ticks in expression

### LINKED ISSUE #3273: Bad parsing when bound data contains a backtick
### Vue.js version
1.0.26

### Reproduction Link
http://codepen.io/sirlancelot/pen/zBPVBo?editors=1010#0

### Steps to reproduce
* Bind an inline object with a value containing the backtick character

### What is Expected?
* Object is reproduced properly in the data model

* `scope.` is being added to everything between the backtick and an escaped quote.

We're using props to curry an initial data model from the server to the browser. We're doing it in a manner similar to the example shown in the CodePen. Basically, our root Vue instance contains a list of props that will show up on the element and are populated by the server.

skyronic:
This fixes #3273 but might introduce other bugs. I'm not sure of it.

### DIFF
--- a/src/parsers/expression.js
+++ b/src/parsers/expression.js
@@ -23,7 +23,7 @@ const improperKeywordsRE =
 
 const wsRE = /\s/g
 const newlineRE = /\n/g
-const saveRE = /[\{,]\s*[\w\$_]+\s*:|('(?:[^'\\]|\\.)*'|"(?:[^"\\]|\\.)*"|`(?:[^`\\]|\\.)*\$\{|\}(?:[^`\\]|\\.)*`|`(?:[^`\\]|\\.)*`)|new |typeof |void /g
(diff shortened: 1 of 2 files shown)


src/parsers/expression.js
```diff
@@ -23,7 +23,7 @@ const improperKeywordsRE =
 
 const wsRE = /\s/g
 const newlineRE = /\n/g
-const saveRE = /[\{,]\s*[\w\$_]+\s*:|('(?:[^'\\]|\\.)*'|"(?:[^"\\]|\\.)*"|`(?:[^`\\]|\\.)*\$\{|\}(?:[^`\\]|\\.)*`|`(?:[^`\\]|\\.)*`)|new |typeof |void /g
+const saveRE = /[\{,]\s*[\w\$_]+\s*:|('(?:[^'\\]|\\.)*'|"(?:[^"\\]|\\.)*"|`(?:[^`\\]|\\.)*\$\{|\}(?:[^`\\"']|\\.|)*`|`(?:[^`\\]|\\.)*`)|new |typeof |void /g
```
skyronic: There's a stray "|" which should be removed here I guess.

yyx990803:
Can you add an accompanying test case?

skyronic:
@yyx990803: updated commit with test case.

sirlancelot:
Thanks!

PR #3303, opened 2016-07-20
Check the element exists

### LINKED ISSUE #3302: Removing data after removed from DOM
I am currently implementing drag and drop functionality into a project of mine and have run into an issue where the library i am using removes the node from the DOM but upon syncing that action in the model, i am getting the following error...

> Uncaught TypeError: Cannot read property 'removeChild' of null

This is because Vue is trying to remove a node that has already been removed by a third party plugin. I imagine it would be fixed by the inclusion of additional checks that the node exists which seems to be blocking the data from being removed.

**Edit**: I have added a gif mimicking the behavior of the third party library to showcase the error.

nblackburn:
Check the DOM node exists before attempting to remove it.

Closes #3302.

### DIFF
--- a/src/util/dom.js
+++ b/src/util/dom.js
@@ -124,7 +124,6 @@ export function after (el, target) {
 
 export function remove (el) {
   if (!inDoc(el)) return false
-  
   el.parentNode.removeChild(el)
 }
 
--- a/src/util/dom.js
+++ b/src/util/dom.js
@@ -123,6 +123,8 @@ export function after (el, target) {
  */
 
 export function remove (el) {
+  if (!inDoc(el)) return false;
+  
   el.parentNode.removeChild(el)
 }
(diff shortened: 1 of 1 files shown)


grafa:
Hey there,

was this fix already incorporated into Vue.js? I just ran into this and I don't know how to fix it.

Thank you.

nblackburn:
@grafa no because it is not recommended that the DOM be manipulated.

grafa:
Well I'm getting this error when I change data in my grid which is based on v-for. I do use track-by but without luck to get rid of this error.

My opinion is when I do call a method in given variable, the variable should be tested first.
I.e.

```
cleanup()  {
    if (myString)
        myString.trim()
   ...
}
```

Its jsut one line of code, one simple test.

PR #3329, opened 2016-07-27
Allow 2 root nodes with v-if and v-else

chrisvfritz:
Since 2 root nodes with `v-if` and `v-else` are guaranteed to return one vnode, they should be considered safe in the root of a template.

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -142,7 +142,11 @@ export function parse (
             )
           }
         }
-      } else if (process.env.NODE_ENV !== 'production' && !stack.length && !warned) {
+      } else if (
+        process.env.NODE_ENV !== 'production' && !stack.length && !warned &&
+        // allow 2 root elements with v-if and v-else
+        !(root.attrsMap.hasOwnProperty('v-if') && element.attrsMap.hasOwnProperty('v-else'))
+      ) {
         warned = true
         warn(
           `Component template should contain exactly one root element:\n\n${template}`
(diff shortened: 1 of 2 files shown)


LinusBorg:
You also have to check for the absence of `v-for` on the element mit `v-if`, as the combination of both would potentially also result in more than one root element.

chrisvfritz:
@LinusBorg There's a [separate warning](https://github.com/chrisvfritz/vue/blob/f1a16e5703c60131387e0b163a2b0dc40ae228f8/src/compiler/parser/index.js#L138-L143) for `v-for` on root, which is triggered whether or not a `v-if` exists.

LinusBorg:
I saw that but was unsure weither it only applies if there *is* one root item.

chrisvfritz:
@LinusBorg Just confirmed with another test locally that this:

```
<div v-if="1" v-for="i in [1]"></div><div v-else></div>
```

does indeed warn about `v-for`, but root constraints were _not_ being applied to the 2nd element with `v-else`. Added a test for it and fixed the issue. 👍

chrisvfritz:
It perhaps should be noted that the following will currently NOT trigger a warning:

``` vue
<p v-if="0"></p>
<p v-else></p>
<p v-else></p>
```

But I consider this a separate issue, since this doesn't warn in [_any_ context](https://jsfiddle.net/chrisvfritz/72eqv40t/), even though it should probably be considered invalid.  Also, the above code does still guarantee the return of a single vnode (either the first or last element - the 2nd element is seemingly ignored by the parser).

yyx990803:
Great job!

Issue #2702, opened 2016-04-22
[feature request] v-model add 'trim' param attribute


luwenjin:
As I remember, angular's `ng-model` will trim it's value by default.
I've tested that using vue's `v-model`, the binding value will leave trailing spaces.

How about adding a `trim` **param attribute** to `v-model` ?
So the tailing spaces will be trimmed automatically.

scenario: 
in register form, to deal with the 'tailing space' issue,
I have to either replace every `this.email` into `this.email.trim()`
or make an computed `email` property to do the trim

azamat-sharapov:
Why not create custom filter?

> angular's ng-model will trim it's value by default

It sucks. What if I want trailing spaces in my app?

luwenjin:
@azamat-sharapov 
ng-model + ng-trim='false' will leave the tailing space intact.

posva:
@luwenjin That implicit behaviour is unintuitive 😕 
Using a filter or triming when needed seems to be the right way to me.

luwenjin:
@posva 
Actually I'm proposing the explicit way

without trim (just v-model):
`<input v-model="email">`

with trim:
`<input v-model="email" trim>`

yyx990803:
It will be a modifier in the future:

``` html
<input v-model.trim="email">
```

tomsmithgroup:
Would it be better to have the modifier as
"notrim"   and make trimming default, at least for 2.0?
Most scenarios need trimming so having a lot of .trim in the form is a bit cluttering

farzher:
I need a trailing space at the end of my textarea using v-model. It looks like it's being trimmed automatically. What's the current workaround to stop trimming?

DannyFeliz:
I think this issue should be closed because in 2.0 now we have a modifier [trim](https://vuejs.org/v2/guide/forms.html#trim)

FFX01:
I like the idea of a trim modifier. I do not like the idea of `v-model` automatically trimming whitespace.

Generally, implicit and arbitrary modifications to data are always a bad thing.

PR #3330, opened 2016-07-27
support input text trim option

Eden-Sun:
according to  #2702 
This PR makes it support 
```html
<input v-model="some" trim>
```
I think it should warn when both params 'number' and 'trim' is selected.


NicolasParada:
There is already 
```html
<input v-model.trim="some">
```

Eden-Sun:
It is for v1 branch.

PR #3367, opened 2016-08-01
prevent nested thunk children from failing

chrisvfritz:
This prevents avoidable failures in some cases, for example in a functional component using JSX that produces a double thunk:

``` js
render: (h, { children }) => <MyComponent>{ children }</MyComponent>
```


posva:
As this may get executed often, is it worth changing the recursion with a while loop? even thought it is less readable?

chrisvfritz:
@posva Good idea! I don't think it's really less readable - just less functional, which is fine. 😃

PR #3393, opened 2016-08-05
made v-show respect display values in style attributes

chrisvfritz:
Addresses issue described [here](https://github.com/vuejs/vue/issues/2873#issuecomment-237921731).

### DIFF
--- a/src/platforms/web/runtime/directives/show.js
+++ b/src/platforms/web/runtime/directives/show.js
@@ -19,7 +19,7 @@ export default {
     }
     const originalDisplay = el.style.display
     el.style.display = value ? originalDisplay : 'none'
-    el.dataset.__vOriginalDisplay = originalDisplay
+    el.__vOriginalDisplay = originalDisplay
   },
   update (el: HTMLElement, { value, oldValue }: VNodeDirective, vnode: VNodeWithData) {
     /* istanbul ignore if */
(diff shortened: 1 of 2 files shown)


miljan-aleksic:
Thank you @chrisvfritz, I hope the PR will be accepted :)

src/platforms/web/runtime/directives/show.js
```diff
@@ -17,22 +17,24 @@ export default {
     if (value && transition && transition.appear && !isIE9) {
       enter(vnode)
     }
-    el.style.display = value ? '' : 'none'
+    const originalDisplay = el.style.display
+    el.style.display = value ? originalDisplay : 'none'
+    el.dataset.__vOriginalDisplay = originalDisplay
```
yyx990803: I don't think IE9~10 supports `dataset`, so we may have to store the value directly on the element itself.

chrisvfritz: @yyx990803 Wow, you're absolutely right - I can't believe I didn't know this! How should we handle this with the Flow types? Declare a custom type that extends `HTMLElement` and allows `__vOriginalDisplay` as a string?

src/platforms/web/runtime/directives/show.js
```diff
@@ -17,7 +17,9 @@ export default {
     if (value && transition && transition.appear && !isIE9) {
       enter(vnode)
     }
-    el.style.display = value ? '' : 'none'
+    const originalDisplay = el.style.display
+    el.style.display = value ? originalDisplay : 'none'
+    el.__vOriginalDisplay = originalDisplay
```
yyx990803: flow doesn't allow attaching properties to HTMLElements, let's just declare `el`'s type as `any` in the function signatures to make it pass.

PR #3411, opened 2016-08-08
Treat bound object with key named 'class' same as :class

### LINKED ISSUE #3382: Merging of HTML classes via class attribute and v-bind broken
### Vue.js version

1.0.26
### Reproduction Link

http://jsfiddle.net/d3xwtrqn/
### Steps to reproduce

As shown in the jsfiddle: The combination of a `class` attribute and the dynamic `:class` leads to a merged class attribute. However using v-bind does not.
### What is Expected?

Both ways should merge correctly.

zigomir:
This should fix #3382. I'm 100% sure we want to do it like this. Issue applies also to v2.0.


yyx990803:
Why is the check added in `style.js` not `bind.js`?

FYI this is already in 2.0 via https://github.com/vuejs/vue/commit/59b2a0eeb32126cd6a35c0c82cfdcddf216e0df8

zigomir:
@yyx990803 mostly because there is already a `for..in` loop where you can check for a key named `class`. I can pull this back to `bind.js` (where it actually belongs, I agree), but then we'll need to introduce a loop to check key names.

PR #3425, opened 2016-08-11
Allow rendering forbidden tags on the server

naumovs:
Currently if you try to render `<script>` tag on the server side you'll get an error, which is not right, because server is not restricted by the rules of the browser. It is useful if server-side component is rendering the `<html>` tag with `<head>` `<body>` and everything else

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -85,7 +85,7 @@ export function parse (
         element.ns = ns
       }
 
-      if (isForbiddenTag(element)) {
+      if (process.env.VUE_ENV !== 'server' && isForbiddenTag(element)) {
         element.forbidden = true
         process.env.NODE_ENV !== 'production' && warn(
           'Templates should only be responsbile for mapping the state to the ' +


yyx990803:
Thanks for the PR - however, you should not be using Vue for rendering the entire page - because in the browser Vue will need to "pick up" the server-rendered DOM, so they must be consistent on both ends. Allowing `<script>` on the server will cause hydration errors on the client.

In most cases you'd wrap the markup rendered from your app with surrounding head/tail markup to make it a complete HTML page.

naumovs:
Yes, I know, but with the appropriate setup it still works very well without any effort:

Please take a look on this [gist](https://gist.github.com/naumovs/6fe32db51d4ee81c7cc6a66270233241).

There are two entry points for server and for the client. The server renderer just sets the server-rendered attribute to the main component from the client-index.js and client picks up rendered content. I feel like it is amazing to drop off other thirdparty templating systems if everything can be done with Vue

yyx990803:
@naumovs ah, yeah that's a pretty neat idea!

Issue #1698, opened 2015-11-04
v-show with transition bug……


ccqgithub:
vue.js version: 1.0.4
template：
    
    <ul class="ul ul-b" v-show="0==1" transition="toggle">
        <li class="li li-b" v-for="sub in menu.subs">                            
            <a href="{{sub.href}}" class="link">{{sub.title}}</a>
        </li>
    </ul>  

    //or just like this:
    <div class="ul ul-b" v-show="0==1" transition="toggle">
        some string
    </div> 

css:

    .ul-b.toggle-transition
        transition all .5s  
        transform scale(1)
        
    .ul-b.toggle-enter,
    .ul-b.toggle-leave
        transform scale(.5)

after first rendered，the dom is not hide，and it‘s html like below：

    <ul class="ul ul-b toggle-transition toggle-leave">

if I make the ul empty like below, that's ok

    <ul class="ul ul-b" v-show="0==1" transition="toggle">
    </ul>

ccqgithub:
another bug：

if I make below：

    <ul class="ul ul-b" v-show="!menu.href && menu._open">
        <li class="li li-b" v-for="sub in menu.subs" v-show="!menu.href && menu._open" transition="toggle">
            <a href="{{sub.href}}" class="link">{{sub.title}}</a>
        </li>
    </ul>

has a error：`TypeError: this.op is not a function`
in code js:

    p.leaveDone = function () {
	  this.left = true
	  this.cancel = this.pendingJsCb = null
	  this.op()  // here
	  removeClass(this.el, this.leaveClass)
	  this.callHook('afterLeave')
	  if (this.cb) this.cb()
	  this.op = null
	}

yyx990803:
Can't reproduce. You will need to make an actual reproduction.

AlexandreBonaventure:
I have the error `TypeError: this.op is not a function` also in 1.0.migration. I will try to make a repro.

ccqgithub:
I will try to make a repro later.

ccqgithub:
@yyx990803 the demo for the first bug in jsfiddle:

http://jsfiddle.net/5b6x9rax/

ccqgithub:
@AlexandreBonaventure   TypeError: this.op is not a function, 

I find my reason(i use variable start with _):

> Properties that start with _ or $ will not be proxied

and you ?

yyx990803:
[...]

PR #3427, opened 2016-08-11
Fix Problem : TypeError: this.op is not a function

sirzxj:
Here is the problem  https://github.com/vuejs/vue/issues/1698

I come accross this problem today 
I found *leaveDone* function execute twice, while this.op is null at the end of first execution lead that problem;

```
p$1.leaveDone = function () {
  this.left = true;
  this.cancel = this.pendingJsCb = null;
  this.op();
  removeClass(this.el, this.leaveClass);
  this.callHook('afterLeave');
  if (this.cb) this.cb();
  this.op = null;
};
```

### DIFF
--- a/src/transition/transition.js
+++ b/src/transition/transition.js
@@ -246,7 +246,7 @@ p.leaveNextTick = function () {
 p.leaveDone = function () {
   this.left = true
   this.cancel = this.pendingJsCb = null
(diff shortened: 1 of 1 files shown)


sirzxj:
This is not my code problem  ?

Issue #3323, opened 2016-07-25
Interpolation fails on inline string containing HTML


chrisvfritz:
As demonstrated in [this fiddle](https://jsfiddle.net/chrisvfritz/dtvcthkh/), the following is interpreted literally:

``` html
<p>{{{ '<strong>hello</strong> world' }}}</p>
```

Instead, I would expect it to produce:

``` html
<p><strong>hello</strong> world</p>
```

This also occurs with normal interpolation (`{{ }}`).

Grafikart:
I create a failing test for it https://github.com/Grafikart/vue/blob/fix-html-interpolation/test/unit/specs/compiler/compile_spec.js#L719 

I try to look at the source to find where it comes from

Grafikart:
I'm working on a fix

## Idea for a fix 

This solution is pretty dirty but I see no way around: 
When compiling a node I check if I find a childNode containing an opening delimiter. If I encounter such a case every Node (until the next childNode containing a closing delimiter) should be treated as text.

Bilie:
Hi

I am not entirely convinced this should be considered a bug.
According to the [docs](https://vuejs.org/guide/syntax.html#Interpolations): The mustache tag will be replaced with the value of the msg property on the corresponding data object.
In your example there is no data object.

To be able to get an error message I modified your example from
```
<p>{{{ '<strong>hello</strong> world' }}}</p>
```
to
```
<p v-html="<strong>hello</strong> world"></p>
```
as the [docs](https://vuejs.org/api/#v-html) state the triple curly braces will ultimately be compiled as v-html directive.
Then you can see in the console the following warning:
> [Vue warn]: Invalid expression. Generated function body: `<scope.strong>scope.hello</scope.strong>scope.world`

What is happening here is vue is looking for the properties on the data object but cannot find them because they are not there.
Here is a working example: https://jsfiddle.net/bsyxkxy2/

chrisvfritz:
@Bilie Your example isn't quite right.

``` html
<p>{{{ '<strong>hello</strong> world' }}}</p>
```

actually compiles to:

``` html
<p v-html="'<strong>hello</strong> world'"></p>
```

And as you can [see here](https://jsfiddle.net/chrisvfritz/nh5vkqav/), the `v-html` version works fine while the interpolated version fails. So it is indeed a bug.

KingMario:
[...]

PR #3489, opened 2016-08-20
Fix html interpolation

Grafikart:
Partial Fix for issue #3323 

This fix allow `<p>{{ '<strong>hello</strong> world' }}</p>`
but `<p>{{{ '<strong>hello</strong> world' }}}</p>` won't work due to the way the compiler works :(. It would force a lots of changes for an edge case.


Grafikart:
Not needed anymore thanks to virtual DOM :)

PR #3496, opened 2016-08-22
Fix Spelling

BrandonSurowiec:


posva:
Thanks for pointing it out. It's actually on more places. If you could replace all the occurences and add it to the PR, that would be perfect 👌 

```
packages/vue-server-renderer/build.js
3462:        process.env.NODE_ENV !== 'production' && warn$1('Templates should only be responsbile for mapping the state to the ' + 'UI. Avoid placing tags with side-effects in your templates, such as ' + ('<' + tag + '>.'));

packages/vue-template-compiler/build.js
3244:        process.env.NODE_ENV !== 'production' && warn$1('Templates should only be responsbile for mapping the state to the ' + 'UI. Avoid placing tags with side-effects in your templates, such as ' + ('<' + tag + '>.'));

src/compiler/parser/index.js
93:          'Templates should only be responsbile for mapping the state to the ' +

test/unit/modules/compiler/parser.spec.js
64:    expect('Templates should only be responsbile for mapping the state').toHaveBeenWarned()
71:    expect('Templates should only be responsbile for mapping the state').toHaveBeenWarned()
```

BrandonSurowiec:
@posva Done!

yyx990803:
@BrandonSurowiec @posva in the future please make sure to only modify `src` files - the `build.js` in packages are auto-generated and only updated when the package is published during releases.

posva:
@yyx990803 My bad. I'll not forget next time 👍

BrandonSurowiec:
@yyx990803 Will do.

PR #3563, opened 2016-08-30
childval can not be array

njleonzhang:
I am not sure, but according to my understanding, childval can not be array, it should be a function().


yyx990803:
It's possible... please make sure the tests pass before submitting PRs.

njleonzhang:
sorry for this, will make test next time

PR #3566, opened 2016-08-30
Add empty-to-undefined modifier to v-model

### LINKED ISSUE #3541: Add a v-model modifier to treat empty input values as null
The straightforward reason for this request is that there's currently no way to create an optional numeric field using v-model since empty inputs are reset to zero.

The fuzzier reason is that in many if not most models an empty input signifies the absence of its value, hence the error message "Please enter a username" instead of "Please enter a username that isn't empty," and the fact that, in contrast to data validation tools like JSON schema, HTML5 doesn't allow `required` fields to be empty.

I have no idea what this modifier should be called, though.

bopjesvla:
Fixes #3541. Still haven't come up with a better name.


bopjesvla:
I have no idea what's up with those TodoMVC tests.

bopjesvla:
this branch is now identical to base but checks still fail

bopjesvla:
ok I'm done

LinusBorg:
make sure to clear your npm cache, then run the tests again

bopjesvla:
the CI runs every test on a brand new linux installation, right?

LinusBorg:
CircleCI seems to cache dependencies, this hit me with a recent PR for vue-router: 

https://github.com/vuejs/vue-router/pull/641#issuecomment-245479682

clear your local cache and re-run them, if that works locally, I guess Evan can reset circleCI'S cache ?

bopjesvla:
Cleared my cache and reinstalled, which resulted in yet another random test failing on the base branch (v-for with transitions iirc). I guess I'll just rollback my rollback and wait for the CircleCI reset.

bopjesvla:
@yyx990803 Could you clear the CircleCI cache?

PR #3579, opened 2016-09-01
Update package.json

kaorun343:
For TypeScript and Typings


package.json
```diff
@@ -3,11 +3,17 @@
   "version": "2.0.0-rc.4",
   "description": "Reactive, component-oriented view layer for modern web interfaces.",
   "main": "dist/vue.common.js",
+  "typings": "index.d.ts",
```
ktsn: This field should be `types/index.d.ts`, shouldn't it?

PR #3591, opened 2016-09-02
Fix definitons

kaorun343:
* add “FunctionalComponentOptions” and “ContextObject”
* fix and update types

#3509 and #3582

### DIFF
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -1,9 +1,11 @@
 import { Vue } from "./vue.d";
-import { VNode, VNodeDirective } from "./vnode.d";
+import { VNode, VNodeData, VNodeDirective } from "./vnode.d";
 
 type Constructor = {
   new (...args: any[]): any;
 }
+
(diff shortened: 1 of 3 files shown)


types/options.d.ts
```diff
@@ -39,6 +41,21 @@ export interface ComponentOptions {
...
 }
 
+export interface FunctionalComponentOptions {
+  props?: string[] | { [key: string]: PropOptions | Constructor | Constructor[] };
+  functional: boolean;
+  render(this: never, createElement: $createElement, context: ContextObject): VNode;
+  name?: string;
+}
+
+export interface ContextObject {
```
ktsn: I think it would be better to rename this interface to more concrete name, like `RenderContext`?

ktsn:
@kaorun343 Thanks for your quick fix 😄

yyx990803:
Seems that there are conflicts after merging #3582 - can we resolve that?

kaorun343:
I will resolve that.

PR #3643, opened 2016-09-08
support short prop declaration as prop: true

iagafonov:
Vue 1 support this form of prop declaration:
`props: {
  someProp: true
}`
[https://jsfiddle.net/89a15uby/](https://jsfiddle.net/89a15uby/)
Vue 2 throw an error `Right-hand side of 'instanceof' is not an object`
[https://jsfiddle.net/jbovp90x/](https://jsfiddle.net/jbovp90x/)

### DIFF
--- a/test/unit/features/options/props.spec.js
+++ b/test/unit/features/options/props.spec.js
@@ -244,6 +244,19 @@ describe('Options props', () => {
       expect(console.error.calls.count()).toBe(2)
       expect('Expected Boolean').toHaveBeenWarned()
     })
+
+    it('optional prop of any type (type: true or prop: true)', () => {
+      makeInstance(1, true)
+      expect(console.error.calls.count()).toBe(0)
+      makeInstance('any', true)
+      expect(console.error.calls.count()).toBe(0)
+      makeInstance({}, true)
+      expect(console.error.calls.count()).toBe(0)
+      makeInstance(undefined, true)
+      expect(console.error.calls.count()).toBe(0)
+      makeInstance(null, true)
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,4 @@
 <!--
 Please make sure to read the Pull Request Guidelines:
-https://github.com/vuejs/vue/blob/dev/.github/CONTRIBUTING.md#pull-request-guidelines
+https://github.com/vuejs/vue/blob/dev/CONTRIBUTING.md#pull-request-guidelines
```
yyx990803: This line is using the url for the next branch files, no need to touch it. Also it's usually a good idea to keep a PR focused on one thing only.

test/unit/features/options/props.spec.js
```diff
@@ -244,6 +244,19 @@ describe('Options props', () => {
       expect(console.error.calls.count()).toBe(2)
       expect('Expected Boolean').toHaveBeenWarned()
     })
+
+    it('optional of any type + null/undefined', () => {
```
yyx990803: probably should change the case description here

iagafonov: changed to "optional prop of any type (type: true or prop: true)"

.github/PULL_REQUEST_TEMPLATE.md
iagafonov: Sorry. Reverted

PR #3651, opened 2016-09-09
Update ComponentOptions

kaorun343:
* Make ComponentOptions as a Generic Interface
* Update some interfaces and types related to ComponentOptions
* Update tests

1. When defining `ComponentOptions.data` as Function, `this` is specified, but when it defined as union types, `this` isn't specified. So the former one is better, I think.
2. These changes enable us to do something like below. Adapting decorators to the `ComponentOptions<V>`, we don't have to define some methods in a class statement that is not a instance method.

```typescript

@VueComponent<MyComponent>({
  data () {
    return {
      message: this.method()
    }
  },
  beforeCreate () {
    this.message = ""
  },
  render (h) {
    return h("div", {}, this.message)
  }
})
class MyComponent extends Vue {
  message: string;

  method(): string {
    return "";
  }
}
```

### DIFF
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -8,7 +8,7 @@ type Constructor = {
 type $createElement = typeof Vue.prototype.$createElement;
 
 export interface ComponentOptions<V extends Vue> {
-  data?(this: V): Object;
+  data?: Object | ((this: V) => Object);
   props?: string[] | { [key: string]: PropOptions | Constructor | Constructor[] };
   propsData?: Object;
   computed?: { [key: string]: ((this: V) => any) | ComputedOptions<V> };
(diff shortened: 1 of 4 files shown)


ktsn:
> When defining ComponentOptions.data as Function, this is specified, but when it defined as union types, this isn't specified. So the former one is better, I think.

I believe it is better not to restrict usage of the library by convenience of its type declaration.
By the way, when I use the vue's typings via node_modules, the `this` type is successfully inferred even if `data` is union type. It's wired...

types/options.d.ts
```diff
@@ -7,37 +7,37 @@ type Constructor = {
...
+  beforeDestroy?(this: V): void;
+  destroyed?(this: V): void;
+  beforeMount?(this: V): void;
+  mounted?(this: V): void;
+  beforeUpdate?(this: V): void;
+  updated?(this: V): void;
 
   directives?: { [key: string]: DirectiveOptions | DirectiveFunction };
-  components?: { [key: string]: ComponentOptions | FunctionalComponentOptions | typeof Vue };
+  components?: { [key: string]: ComponentOptions<V> | FunctionalComponentOptions | typeof Vue };
```
ktsn: I think this should be `ComponentOptions<Vue>` because the `ComponentOptions` passed to this option indicate different components from `V`.

yyx990803:
Is this ready for merge?

kaorun343:
Yes.

ktsn:
LGTM

PR #3656, opened 2016-09-10
Add undefined check for instance methods

Kingdaro:
Adding an undefined property to `methods` on vue components leads to a very ambiguous error message: `Uncaught TypeError: Cannot read property 'length' of undefined`

Example:
```js
import stuff from './stuff'

export default {
  template: '...',
  methods: {
    someMethod: stuff.someMetod // typo!!!
  }
}
```
It's an easy mistake to make. This checks for undefined properties and warns if it finds one.

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -142,7 +142,9 @@ function initMethods (vm: Component) {
   const methods = vm.$options.methods
   if (methods) {
     for (const key in methods) {
-      vm[key] = bind(methods[key], vm)
+      if (methods[key] != null) {
+        vm[key] = bind(methods[key], vm)
+      }
     }
   }
 }
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -144,7 +144,7 @@ function initMethods (vm: Component) {
     for (const key in methods) {
       if (methods[key] != null) {
         vm[key] = bind(methods[key], vm)
-      } else {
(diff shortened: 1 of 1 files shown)


src/core/instance/state.js
```diff
@@ -142,7 +142,11 @@ function initMethods (vm: Component) {
   const methods = vm.$options.methods
   if (methods) {
     for (const key in methods) {
-      vm[key] = bind(methods[key], vm)
+      if (methods[key] != null) {
+        vm[key] = bind(methods[key], vm)
+      } else {
+        warn(`The method ${key} on vue instance is undefined.`, vm)
```
yyx990803: can we add `else if (process.env.NODE_ENV !== 'production')` here so that the warning can be stripped?

Kingdaro: Done!

src/core/instance/state.js
```diff
@@ -142,7 +142,11 @@ function initMethods (vm: Component) {
   const methods = vm.$options.methods
   if (methods) {
     for (const key in methods) {
-      vm[key] = bind(methods[key], vm)
+      if (methods[key] != null) {
+        vm[key] = bind(methods[key], vm)
+      } else if (process.env.NODE_ENV !== 'production') {
+        warn(`The method ${key} on vue instance is undefined.`, vm)
```
zzz6519003: Ahh I get it, You saved me!!!!!!!!!!

PR #3709, opened 2016-09-20
Allow setting of compiled render functions via template option

fergaldoyle:
In scenarios where we have pre-compiled templates, `compiler.compileToFunctions()`, you need to set both `render` and `staticRenderFns` in component options or use a mixin.

This PR creates a shortcut, by allowing the `template` option receive an object of complied functions.

In conjunction with something like https://www.npmjs.com/package/vue-template-compiler-loader, the following would be possible (where the imported template is pre-compiled and not a string):

```javascript
new Vue({
    el: '#app',
    template: require('./template.html')
})
```

or

```javascript
import template from './component-a.html'
new Vue({
    el: '#app',
    template
})
```

### DIFF
--- a/src/core/instance/lifecycle.js
+++ b/src/core/instance/lifecycle.js
@@ -40,6 +40,12 @@ export function lifecycleMixin (Vue: Class<Component>) {
   ): Component {
     const vm: Component = this
     vm.$el = el
+    const template = vm.$options.template
+    if (template && template.render && template.staticRenderFns) {
+      vm.$options.render = template.render
+      vm.$options.staticRenderFns = template.staticRenderFns
+      vm.$options.template = null
+    }
     if (!vm.$options.render) {
       vm.$options.render = emptyVNode
       if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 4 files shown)


yyx990803:
I'm not sure if I like the idea of overloading `template` for this. Actually you can just do:

``` js
import render from './component-a.html'

new Vue({
    el: '#app',
    ...render
})
```

fergaldoyle:
I found that just setting the `render` function without `staticRenderFns` lead to errors...

I assumed you needed both?

fergaldoyle:
e.g. a template of `<div></div>` while generate this object:
```
{
  render: function(){with(this){return _m(0)}},
  staticRenderFns: [function(){with(this){return _h('div')}}]
}
```

Setting just `render` will give error: `TypeError: this._staticTrees is null`

fnlctrl:
@fergaldoyle You may have missed the object spread operator `...`  ;)
```js
new Vue({
    el: '#app',
    // note the ... here, it will spread the `render` object and add its properties to the outer object
    ...render 
})
```

So Evan's idea was that you can make `import render from './component-a.html'` give you an object that is 
```js
{
  render: function(){ ... },
  staticRenderFns: [ ... ]
}
```
So after the spread you will have
```js
new Vue({
    el: '#app',
    render: function(){ ... },
    staticRenderFns: [ ... ]
})
```

fergaldoyle:
Thanks, I did miss that. (Also needed to install `transform-object-rest-spread` plugin for Babel).

I'm happy with that approach and for this PR to be closed.

PR #3714, opened 2016-09-21
Graphql Component Element

alexbowers:
This is the beginning of the graphql implementation


alexbowers:
Awaits vueify compilation too

alexbowers:
Ref: https://github.com/vuejs/vueify/pull/134

yyx990803:
Hmm, I'd wait until we figure out the design. I think it's a cool idea but not entire sure whether it should be done as a tag yet.

Issue #3689, opened 2016-09-17
[2.0] Internet Explorer select elements display first option by default


fergaldoyle:
Similar to issue https://github.com/vuejs/vue/issues/2983 which was logged for 1.x 

### Vue.js version
2.0 rc

### Reproduction Link
http://codepen.io/anon/pen/GjqwaA?editors=1010#0

### Steps to reproduce
Open the above codepen in IE

### What is Expected?
The visible select elements should be empty / have no option selected

### What is actually happening?
The visible select elements all have "AAA" selected. The underlying model is not in sync. Interestingly when you click the "show" button, the select elements display as expected.

PR #3719, opened 2016-09-22
Select fix for IE & Edge

fergaldoyle:
RE: https://github.com/vuejs/vue/issues/3689

Fixes select (and multiple) for IE9, IE10, IE11 and Edge

`setTimeout` isn't ideal but `nextTick` doesn't work once an option has been selected and page refreshed.

### DIFF
--- a/src/platforms/web/runtime/directives/model.js
+++ b/src/platforms/web/runtime/directives/model.js
@@ -95,12 +95,11 @@ function setSelected (el, binding, vm) {
   if (!isMultiple) {
     el.selectedIndex = -1
   }
-  if ((isIE || isEdge) && el.selectedIndex === -1) {
-    vm.$on('hook:mounted', () => {
-      setTimeout(() => {
-        setSelected(el, binding, vm)
-      }, 0)
-    })
+  if ((isIE || isEdge) && el.selectedIndex === -1 && !el._v_ieselectdone) {
+    el._v_ieselectdone = true;
+    setTimeout(() => {
+      setSelected(el, binding, vm)
+    }, 0)
   }
 }
 
--- a/src/platforms/web/runtime/directives/model.js
+++ b/src/platforms/web/runtime/directives/model.js
@@ -5,7 +5,7 @@
 
 import { looseEqual, looseIndexOf } from 'shared/util'
 import { warn } from 'core/util/index'
-import { isAndroid, isIE9 } from 'web/util/index'
+import { isAndroid, isIE9, isEdge, isIE } from 'web/util/index'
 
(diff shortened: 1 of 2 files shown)


fergaldoyle:
For cases where a select element is inside a `v-if`, `hook:mounted` would have already been called. Using just `setTimeout` works fine.

yyx990803:
Have you tried just changing the `bind` hook to `inserted` instead?

fergaldoyle:
Changing [#L24](https://github.com/vuejs/vue/blob/054b96bd7079bf54a079d1f17bea3f215970fdea/src/platforms/web/runtime/directives/model.js#L24) to `inserted` only doesn't fix the issue. It seems you have for a certain period after insertion before re-setting selection.

When `inserted` is called,  `selectedIndex` is already `0` when it should be `-1`.

Further playing round with various scenarios (select elements in root template, sub components, inside v-if, inside v-for) the following two options work:

### 1
Add the `inserted` hook and place IE workaround here. Can't check for `selectedIndex === -1` as it's already been reset by IE. Using `setTimeout` causes minor UI flicker (where select element value is changing) so `setSelected` instantly called then called against in `setTimeout`.
```javascript
  inserted (el, binding, vnode) {
    if((isIE || isEdge) && vnode.tag === 'select') {
      // IE/Edge will set selectedIndex to 0 in cases
      // where it should be -1, force the select to refresh.
      setSelected(el, binding, vnode.context)
      setTimeout(() => {
        setSelected(el, binding, vnode.context)
      }, 0)
    }
  },
```

### 2
Similar to the current PR, workaround is placed in the `setSelected` function. `selectedIndex === -1` so workaround will only run when this is the case. To prevent recursion `_v_ieselectdone` state is saved on the el. `setSelected` is called twice, once in `nextTick` and once in `setTimeout` to prevent UI flicker
```javascript
  if ((isIE || isEdge) && el.selectedIndex === -1 && !el._v_ieselectdone) {
    // IE/Edge will set selectedIndex to 0 in cases
    // where it should be -1, force the select to refresh.
    el._v_ieselectdone = true
    nextTick(() => {
      setSelected(el, binding, vm)
    })
    setTimeout(() => {
      setSelected(el, binding, vm)
    }, 0)
  }
```

yyx990803:
Hmm, using `inserted` seems to be working fine for me in all cases - can you specify where it's not working as intended?

yyx990803:
Thanks a lot of looking into this - I need to get a release out before getting off work today so I implemented your 1st suggestion (with your credit in the commit message) - hope you don't mind!

fergaldoyle:
I don't mind at all!

PR #3726, opened 2016-09-23
fixed the error when _scope and _context all null

Litor:
if _scope and _context all null ,  it will throw error when get $refs from null


yyx990803:
Thanks for the PR, do you have a reproduction of this error showing up? Just want to know what caused this because there's potentially a bug elsewhere.

Litor:
there is a axample here: https://github.com/Litor/ubase-vue-example
1、npm instal
2、npm run dev
3、http://localhost:8081/test1/#!/test2
4、switch right top tab test1 and test2, then get the error ubase-vue-test.js:8744 Uncaught TypeError: Cannot read property '$refs' of null

Litor:
@yyx990803 
there is a axample here: https://github.com/Litor/ubase-vue-example
1、npm instal
2、npm run dev
3、http://localhost:8081/test1/#!/test2
4、switch right top tab test1 and test2, then get the error ubase-vue-test.js:8744 Uncaught TypeError: Cannot read property '$refs' of null

posva:
@Litor Can minimise the code into a jsfiddle, please? That'd help us a lot!

Litor:
the error is caused by keep-alive, after remove keep-alive, it's worked fine

Litor:
I used the dynamic component in a vue component A, when the router changed vue component A destroyed failed and throw the error above.

after debug in chrome, i found the  error is cause by Vue.prototype._destroy function in instance/internal/lifecycle.js  

     // destroy all children.
      i = this.$children.length;
      while (i--) {
        this.$children[i].$destroy();
      }
      // teardown props
      if (this._propsUnlinkFn) {
        this._propsUnlinkFn();
      }
      // teardown all directives. this also tearsdown all
      // directive-owned watchers.
      if (this._unlinkFn) {
        this._unlinkFn();
      }

teardown all directives should execute before destroy all children, here maybe the root cause!

Litor:
@posva I used the dynamic component with keep-alive in a vue component A, when the router changed vue component A destroyed failed and throw the error above.

after debug in chrome, i found the error is cause by Vue.prototype._destroy function in instance/internal/lifecycle.js

origin source in instance/internal/lifecycle.js

 // destroy all children.
  i = this.$children.length;
  while (i--) {
    this.$children[i].$destroy();
  }
  // teardown props
  if (this._propsUnlinkFn) {
    this._propsUnlinkFn();
  }
  // teardown all directives. this also tearsdown all
  // directive-owned watchers.
  if (this._unlinkFn) {
    this._unlinkFn();
  }
teardown all directives should execute before destroy all children, here maybe the root cause!

PR #3734, opened 2016-09-23
use a simple expression parser instead of regexp to parse directive

### LINKED ISSUE #3210: Parsing object as a filter argument fails with spaces inside

defcc:
Use a state machine based expression parser to parse the directive.

 fix #3210 and #3214 #

### DIFF
--- a/src/parsers/directive.js
+++ b/src/parsers/directive.js
@@ -2,36 +2,173 @@ import { toNumber, stripQuotes } from '../util/index'
 import Cache from '../cache'
 
 const cache = new Cache(1000)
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks and nice job! 👍

libmw:
@defcc good job!

PR #3762, opened 2016-09-25
fix Flow lifecycle method name [Vue@2.x]

bluemsn:
fix lifecycle method name `init` to `beforeCreate`


LinusBorg:
good catch!

bluemsn:
@LinusBorg  ?

LinusBorg:
It means good that you found this. 

Evan will surely merge this soon.

PR #3776, opened 2016-09-26
fix directive sort compare

### LINKED ISSUE #3768: Directive params casted into strings
### Vue.js version
1.0.27

### Description
Parameter data type is illegal to chang

### Reproduction Link

[jsfiddle](https://jsfiddle.net/jawc4bpo/)

### What is Expected?

Here is same [example](https://jsfiddle.net/gn4aezn2/)  with version 1.0.26 where it works

defcc:
fix #3768 

Object.keys(groupedMap) returns an array whose elements are strings, so it breaks the sort function.

### DIFF
--- a/src/compiler/compile.js
+++ b/src/compiler/compile.js
@@ -122,20 +122,20 @@ function sortDirectives (dirs) {
 
   var groupedMap = {}
   var i, j, k, l
-  var keys = []
+  var index = 0
+  var priorities = []
   for (i = 0, j = dirs.length; i < j; i++) {
     var dir = dirs[i]
     var priority = dir.descriptor.def.priority || DEFAULT_PRIORITY
     var array = groupedMap[priority]
     if (!array) {
       array = groupedMap[priority] = []
-      keys.push(priority)
+      priorities.push(priority)
     }
     array.push(dir)
   }
 
-  var index = 0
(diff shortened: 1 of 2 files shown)


posva:
Thanks! LGTM
I'm not sure about the test though

defcc:
Ok, current test case seems not to fully cover the code. I will try to update the test case soon

src/compiler/compile.js
```diff
@@ -122,18 +122,20 @@ function sortDirectives (dirs) {
...
     if (!array) {
       array = groupedMap[priority] = []
+      keys.push(priority)
     }
     array.push(dir)
   }
 
   var index = 0
-  var priorities = Object.keys(groupedMap).sort(function (a, b) {
+  var priorities = keys.sort(function (a, b) {
```
yyx990803: sort is in-place, so we can just name this array `priorities` from the start, no need to assign a separate variable.

defcc:
forgot it. I will fix it later today

src/compiler/compile.js
defcc: thanks, I have updated the code :)

yyx990803:
Thanks! 👍

juukie:
Thanks, fixed the bug 👍

Issue #3798, opened 2016-09-28
v2.0.0-rc.8 IE/Edge select fix regression


fergaldoyle:
Related to: 
https://github.com/vuejs/vue/issues/3689
https://github.com/vuejs/vue/pull/3719

Unfortunately reverting `nextTick`  to use microtask semantics causes the IE/Edge select fix to only partially work. IE/Edge remembers the selected value after page refresh (but is OK after a hard refresh)

I was wondering why using just `nextTick` in my PR didn't work when I did for you, then I saw that `nextTick` had changed implementation so must have been taking that little bit longer before it executed.

Potential solutions:
Might be some flicker
````javascript
if (isIE || isEdge) {     
  setTimeout(function () {
    setSelected(el, binding, vnode.context)
  }, 0)        
}
```

No flicker
````javascript
if (isIE || isEdge) {
  nextTick(function () {
    setSelected(el, binding, vnode.context)
  })        
  setTimeout(function () {
    setSelected(el, binding, vnode.context)
  }, 0)        
}
```

yyx990803:
ugh, stupid IE... I guess we'll go with the no flicker version of it (they can share the same callback) - PR welcome.

PR #3803, opened 2016-09-28
fix IE/Edge select remembering selection

fergaldoyle:
Fix for https://github.com/vuejs/vue/issues/3798


src/platforms/web/runtime/directives/model.js
```diff
@@ -35,9 +35,11 @@ export default {
       setSelected(el, binding, vnode.context)
       /* istanbul ignore if */
       if (isIE || isEdge) {
-        nextTick(() => {
+        let cb = () => {
```
yyx990803: Note sure why CI is not running, but the linter setup will require using `const` here

fergaldoyle: Done

PR #3818, opened 2016-09-30
Fix $createElement and Vue.extend

kaorun343:
According to 06b4703, `namespace` has been removed, so I updated the types.
Also, I fixed the first argument of `$createElement`.
In addition, add `FunctionalComponentOptions` to `Vue.extend`

### DIFF
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -9,6 +9,8 @@ import {
 import { VNode, VNodeData, VNodeChildren } from "./vnode";
 import { PluginFunction, PluginObject } from "./plugin";
 
+type Component = typeof Vue | ComponentOptions<Vue> | FunctionalComponentOptions;
+
 export declare class Vue {
 
   constructor(options?: ComponentOptions<Vue>);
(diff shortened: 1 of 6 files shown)


types/vue.d.ts
```diff
@@ -39,10 +41,9 @@ export declare class Vue {
   $emit(event: string, ...args: any[]): this;
   $nextTick(callback?: (this: this) => void): void;
   $createElement(
-    tag?: string | Vue,
+    tag?: string | Component | (() => string | Component),
```
ktsn: Actually the function type of this argument is the factory of async component.
So I think following type is more appropriate.

```js
string |
Component |
((
  resolve: (component: Component) => void,
  reject: any
) => Promise<Component> | Component | void)
```

Maybe it would be better to define `AsyncComponent` for this function type.

ktsn: FYI: https://rc.vuejs.org/guide/components.html#Async-Components

kaorun343:
I've fixed.

types/vue.d.ts
```diff
@@ -39,10 +41,9 @@ export declare class Vue {
   $emit(event: string, ...args: any[]): this;
   $nextTick(callback?: (this: this) => void): void;
   $createElement(
-    tag?: string | Vue,
+    tag?: string | Component | AsyncComponent,
```
zigomir: That's exactly what was breaking things up for me 👍

PR #3826, opened 2016-10-01
Fix #3824 Allow text nodes on static templates in components

### LINKED ISSUE #3824: <template> unable to work, Cannot create property 'isStatic' on string ' '
### Vue.js version
2.0.1

### Reproduction Link
https://jsfiddle.net/3qpzv7v0/

template tag should be able to use for group multiple elements and inject into component slot. this used to work in rc8

posva:
It may be a better way to move the textnode out of the static render tree

### DIFF
--- a/test/unit/features/component/component-slot.spec.js
+++ b/test/unit/features/component/component-slot.spec.js
@@ -523,4 +523,17 @@ describe('Component slot', () => {
       expect(spy).toHaveBeenCalled()
     }).then(done)
   })
+
+  it('renders static tree with text', () => {
+    const vm = new Vue({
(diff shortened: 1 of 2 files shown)


test/unit/features/component/component-slot.spec.js
```diff
@@ -523,4 +523,17 @@ describe('Component slot', () => {
       expect(spy).toHaveBeenCalled()
     }).then(done)
   })
+
+  fit('renders static tree with text', () => {
```
DannyFeliz: fit?

posva: I forgot to change it back to `it`. Thank you!

PR #3845, opened 2016-10-04
update validateProp

### LINKED ISSUE #3839: Vue does not refuse invalid prop values
### Vue.js version
2.0.1

### Reproduction Link
https://jsfiddle.net/ncsh5yse/1/
https://jsfiddle.net/ncsh5yse/2/

### Steps to reproduce
Just open above link

### What is Expected?
No text is shown

Text `why this value is not rejected ?` is shown

### Description
Document says "When a prop validation fails, Vue will refuse to set the value", and v1.0.x behaves so.
But v2.0.1 does not refuse invalid prop values (only shows warning).

defcc:
fix #3839

when assertProp returns false, assign value to undefined

### DIFF
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -39,7 +39,9 @@ export function validateProp (
     observerState.shouldConvert = prevShouldConvert
   }
   if (process.env.NODE_ENV !== 'production') {
-    assertProp(prop, key, value, vm, absent)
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks for the PR! See https://github.com/vuejs/vue/issues/3839#issuecomment-251776877

PR #3864, opened 2016-10-07
Switch to https

vinkla:
Replaced all http vuejs.org links with https.


posva:
@phanan I cannot upvote that, so, 👍 
we should redirect to https
I didn't even know the site could work on https

phanan:
The cert is even wildcard apparently: https://jp.vuejs.org/.

PR #3866, opened 2016-10-07
use bind function from shared/util

defcc:
there is  a bind implementations in [shared/util.js](https://github.com/vuejs/vue/blob/dev/src/shared/util.js#L113), so just use that.


yyx990803:
No need for `bind` here because `nextTick` doesn't take any arguments. `bind` does some extra conditional checks.

PR #3868, opened 2016-10-07
setTimeout fallback update 

defcc:
update setTimeout fallback


LinusBorg:
What is the advantage of this change, in your opinion?

defcc:
As for the Promise and MutationObserver  branch，there is no need to take the two args when timerFunc invoked

PR #3876, opened 2016-10-08
Update typings

ktsn:
### Update methods option type

Apply `this` type in the functions of `methods` option like lifecycle hooks. (`Function` is alias of `(...args: any[]) => any`)
We can omit `this` type annotation for methods option due to this update (See options-test.ts).

### Improve $createElement type and expose it

1. Allow to create empty vnode (`h()`).
2. Allow to pass vnode children to 2nd argument.

Also, this patch enable us to use the type of `$createElement`. It would be useful when we annotate render option type manually.

```js
import { CreateElement } from 'vue'

export default {
  render (h: CreateElement) {
    // ...
  }
}
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-import {Vue as _Vue} from "./vue";
+import * as V from "./vue";
 import * as Options from "./options";
 import * as Plugin from "./plugin";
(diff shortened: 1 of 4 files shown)


kaorun343:
LGTM.

PR #3885, opened 2016-10-09
add getTypeList to process mutli-type prop

### LINKED ISSUE #3879: Multi-type prop validation is not consistent
### Vue.js version
2.0.1

### Reproduction Link
https://jsfiddle.net/alexman/f7sb0rd3/9/

### Steps to reproduce
Create a component with a prop of type `[String, Boolean]` and instantiate it with a boolean value or shorthand prop notation. (`<my-component my-prop></my-component>`)

### What is Expected?
Booleans are accepted and passed to the validator as a boolean value.

You'll see that it will not validate booleans / shorthand prop notation. 
Defining it reversed, `[Boolean, String]`, does seem to work as expected except for not being able to pass false as a value.

### Observations
I noticed `validateProp()` does the following "magic", which doesn't seem to take the option of an array of types into account?
```
  if (getType(prop.type) === 'Boolean') {
    if (absent && !hasOwn(prop, 'default')) {
      value = false
    } else if (value === '' || value === hyphenate(key)) {
      value = true
    }
  }
```

defcc:
the getType func only returns the first prop type, even if the prop type is an array.

Add getTypeList func to get prop types. 
fix #3879

### DIFF
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -21,7 +21,7 @@ export function validateProp (
   const absent = !hasOwn(propsData, key)
   let value = propsData[key]
   // handle boolean props
-  if (getTypeList(prop.type).indexOf('Boolean') >= 0) {
+  if (isBooleanType(prop.type)) {
     if (absent && !hasOwn(prop, 'default')) {
       value = false
     } else if (value === '' || value === hyphenate(key)) {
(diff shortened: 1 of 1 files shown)


src/core/util/props.js
```diff
@@ -21,7 +21,7 @@ export function validateProp (
   const absent = !hasOwn(propsData, key)
   let value = propsData[key]
   // handle boolean props
-  if (getType(prop.type) === 'Boolean') {
+  if (getTypeList(prop.type).indexOf('Boolean') >= 0) {
```
yyx990803: Since this function is pretty perf-sensitive, I'd suggest a `isBooleanType(prop.type)` function that:

- If `prop.type` is Array, iterate with `for` and return if any equals `'Boolean'`
- otherwise just check `=== 'Boolean'`.

defcc: thanks, got it

PR #3903, opened 2016-10-11
replace special terminators chars

### LINKED ISSUE #3895: Issue with rendering certain Unicode characters
### Vue.js version
2.0.1

### Reproduction Link
http://jsfiddle.net/wmmw143f/1/

### Steps to reproduce
Run the JSFiddle above and you will see the content isn't displayed and errors in the console.

### What is Expected?
I would expect the app to run regardless of the Unicode characters, without errors.

```
[Vue warn]: failed to compile template:

<div id="app">
  This is user-generated content, and unfortunately includes the Unicode character 2028 right here:  .
</div>

 (found in root instance)
[Vue warn]: Error when rendering root instance: 
Uncaught TypeError: Cannot set property 'isStatic' of undefined
```

### Background
I'm not sure if this is the right place for this or not, but I'm of the feeling this is a bug with Vue. I'll also point out that I'm not that knowledgable about character encodings, but here is the information I have.

We have been testing the Vue 2.0 upgrade on our website which has a lot of user-generated content. It appears that some of our user-generated contains uncommon Unicode characters like [U+2028 Line Separator](http://www.fileformat.info/info/unicode/char/2028/index.htm). These appear invisible by standard.

However, when they appear in the content of a Vue 2.0 page, Vue fails completely and then removes it's entire DOM node from the tree, leaving the page bare.

[Here is a JSFiddle that demonstrates the issue](http://jsfiddle.net/wmmw143f/1/). You'll notice that JSFiddle highlights the Unicode character in question. When you remove that character, the app runs correctly.

defcc:
replace terminator chars  \u2028 and \u2029 to avoid syntax error. fix #3895

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -22,6 +22,7 @@ const bindRE = /^:|^v-bind:/
 const onRE = /^@|^v-on:/
 const argRE = /:(.*)$/
 const modifierRE = /\.[^\.]+/g
+const specialNewlineRE = /\u2028|\u2029/g
 
 const decodeHTMLCached = cached(decodeHTML)
 
@@ -222,6 +223,8 @@ export function parse (
             text
           })
         } else {
+          // #3895 special character
+          text = text.replace(specialNewlineRE, '')
           currentParent.children.push({
             type: 3,
             text


yyx990803:
👍 good job!

PR #3904, opened 2016-10-11
Remove the License badge from README.md

limonte:
The License is already in interfaces:

**github.com:**

**npmjs.com:**

No need to duplicate this info.


limonte:
If you agree with this, we can consider removing this part of README.md in scope of this PR:

LinusBorg:
Sorry, but I don't like this. It's common practice to mention the Licence in the README.

The fact that it's visible in certain(!) interfaces doesn't matter, especially since the repository is also accessible from e.g. the git commandline, where such interface-information is not present.

limonte:
@LinusBorg 

> the git commandline, where such interface-information is not present.

In command line you can always do `cat LICENSE` :) But it's ok, I understand your reasoning.

PR #3922, opened 2016-10-12
select change event fix

### LINKED ISSUE #3917: Multiple select elements @change event problem
Multiple select box does not work in same Vue instance.

### Vue.js version
* Vue.js v2.0.2

### Reproduction Link
https://jsfiddle.net/hd0drbo0/3/

### Steps to reproduce
Change second select box value, method "testMethod" will be called infinite times, even it is not bonded to second select box.
Change first select box value, and later change second select box value, everything works as expected.

### What is Expected?
Update of different models of select.
Infinite call of not assigned @change method.

defcc:
If select binding not changed, then needRest should be set to false, and no **change** event should be emitted. fix #3917

### DIFF
--- a/src/platforms/web/runtime/directives/model.js
+++ b/src/platforms/web/runtime/directives/model.js
@@ -60,7 +60,7 @@ export default {
       // option in the DOM.
       const needReset = el.multiple
         ? binding.value.some(v => hasNoMatchingOption(v, el.options))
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks! In the future please include `(fix #xxxx)` in the commit message - this makes it easier to write / generate release logs.

defcc:
ok, got it :)

PR #3932, opened 2016-10-13
Fix using one delay to control all transitions.

guidobouman:
Vue transitions have a wrong timing when having CSS like the following:

```
.[transition-name]-enter-active, .[transition-name]-leave-active {
  transition: opacity 0.8s ease, transform 0.7s ease;
  transition-delay: 0.4s;
}
```

Which in turn bubbles errors to the console. This PR fixes the bug.

I have a question though: Is parameter re-assign allowed in Vue? It's not configured in the ESlint setup.

### DIFF
--- a/src/platforms/web/runtime/transition-util.js
+++ b/src/platforms/web/runtime/transition-util.js
@@ -131,10 +131,10 @@ export function getTransitionInfo (el: Element, expectedType?: ?string): {
 }
 
 function getTimeout (delays: Array<string>, durations: Array<string>): number {
-  while(delays.length < durations.length) {
-    delays = delays.concat(delays);
+  while (delays.length < durations.length) {
+    delays = delays.concat(delays)
   }
-  
+
   return Math.max.apply(null, durations.map((d, i) => {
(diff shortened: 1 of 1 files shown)


guidobouman:
Also, where'd I put a test for this? I assume after `custom-transition-classes` in `test/unit/features/transition/transition.spec.js`?

And should this actually be tested? As this seems like a niche thing to test.

yyx990803:
Yeah, looks quite a niche to test. Could you please provide a fiddle so that I can confirm it though?

guidobouman:
@yyx990803 https://jsfiddle.net/o8u35ywu/

guidobouman:
Also, delays is never an empty array, so concat can't end up in an infinite loop.

To be sure we could add a fallback like `delays = delays.concat(delays || '0s')` But that would mean we need to know the syntax of a delay inside the function, which could cause issues when a browser decides to change the syntax.

guidobouman:
@yyx990803 This should be mergeable, right?

yyx990803:
Thanks!

Issue #3979, opened 2016-10-18
Unable to watch nested data


allistera:
### Vue.js version
2.0.2

### Reproduction Link
http://jsfiddle.net/2vqkLkjL/1/

### Steps to reproduce
- Create a component with a nested data property: Object, of an object containing an array.
- Set up watcher for this property, specifying the deep flag.

### What is Expected?
When the value of an array changes, the watcher runs

### What is actually happening?
Watcher isn't getting triggered.

The code seemed to work fine in Vue 1.x.

Not 100% nested objects/arrays is the correct way for managing a view with lots of inputs.

defcc:
the same problem with #3958   I am resolving this problem

@allistera  You could use object as a workaround now.

allistera:
Ah I should have done a better job searching. Thanks deffc awesome work!

LinusBorg:
closing as it duplicates #3958

PR #3988, opened 2016-10-20
v-model binding with array.  (fix #3958,#3979)

### LINKED ISSUE #3958: vue 2.0.3 v-model not updating using an array
I'm in process migrating an application from Vue 1.x to 2.0.3. In this application I'm having several selects where each v-model binds to an individual index of an array . This is without problems in Vue 1.x but in version 2 it does not update.  I believe this is a potential bug.

I have reproduced the problem here:
https://jsfiddle.net/peterkorgaard/a7vvz753/18/

### Vue.js version
2.0.3

### Reproduction Link
https://jsfiddle.net/peterkorgaard/a7vvz753/18/

defcc:
fix #3958 #3979 
As for array type case, v-model="model[idx]" binding uses **model[idx]=** internally, so the value change could not be reactive.

the binding model expression maybe be as follows:

- test
- test[idx]
- test[test1[idx]]
- xxx.test[a[a].test1[idx]]
- test.xxx.a["asa"][test1[idx]]
- test["a"][idx]

So I use a modelParser to parse the model expression to two parts **exp** and **idx**, and if the exp is an Array type, then use $$exp.splice($$idx, 1, value)

### DIFF
--- a/src/platforms/web/compiler/directives/model.js
+++ b/src/platforms/web/compiler/directives/model.js
@@ -2,6 +2,7 @@
 
 import { isIE } from 'core/util/env'
 import { addHandler, addProp, getBindingAttr } from 'compiler/helpers'
+import parseModel from 'web/util/model'
 
 let warn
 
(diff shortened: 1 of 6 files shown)


yyx990803:
Beautifully done! 💯

blade254353074:
@defcc @yyx990803 
What about `test[index].key`?  
test: [{ key: 'value' }]

defcc:
already fixed

PR #4010, opened 2016-10-22
Add CDNJS version badge

PeterDaveHello:
This will show the latest vue version on CDNJS


yyx990803:
Thanks Peter, however it seems there's not much point in showing the badge since the assumption is that CDNs sync automatically after all. We'd also want to avoid cluttering the README and display only vital information.

PR #4043, opened 2016-10-27
fix problematic use of 'deprecated' in warnings

### LINKED ISSUE #4042: Interpolation inside attributes has been removed, not just deprecated
I brought this up in the forums first and @phanan agreed that "deprecated" is misleading in this warning message that is logged when trying to use interpolation inside attributes: 

'Interpolation inside attributes has been deprecated. Use v-bind or the colon shorthand instead. For example, instead of `<div id="{{ val }}">`, use `<div :id="val">`.'

This suggests that the feature is still present but will be removed in the future and so should be avoided. However, it has already been removed so it seems that "deprecated" should be replaced with "removed." I'm happy to create a PR if that is helpful.

Repro steps:
1. Use a mustache in any HTML attribute.
2. View messages logged to the console.

This fiddle doesn't show the console warning but demonstrates that the feature has been removed: http://jsfiddle.net/zcg2d825/1/

chrisvfritz:
Fixes #4042.

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -399,7 +399,7 @@ function processAttrs (el) {
         if (expression) {
           warn(
             `${name}="${value}": ` +
-            'Interpolation inside attributes has been deprecated. ' +
+            'Interpolation inside attributes has been removed. ' +
             'Use v-bind or the colon shorthand instead. For example, ' +
             'instead of <div id="{{ val }}">, use <div :id="val">.'
           )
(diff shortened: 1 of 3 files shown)


phanan:
Erm… The test failed.

chrisvfritz:
Yeah, just saw that I missed the matching tests - it's quite late where I am. 😅  They should pass now.

defcc:
haha，npm run test 👍

PR #4046, opened 2016-10-27
templates as a function

eco747:
It should be useful to be able to specify template as a function that return a template string.

    Vue.extend({
        template: function( ) { 
            return '<div>'+ this.my_func() + </div>';
        }
    });


fnlctrl:
No spams please.

PR #4051, opened 2016-10-28
Check and warn for falsy class name (fixes #4050)

### LINKED ISSUE #4050: Transition class bind null values has DOMException in addClass function.
### Vue.js version

2.0.1

```
<transition :enter-active-class="animatedIn" >
```

When 'animateIn' bind value is empty string or null,addClass function has exception:

```
vue.js:4540 Uncaught (in promise) DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty
```

Semantically,Binding a null value is intended to remove the transition effect.
Should a null check be made in addClass function or the transition should be ignored in the case of null values?

phanan:
This commit adds a check for falsy names (null or empty string) before attempting to add or remove them, to prevent a DOM exception. A warning will also be triggered if in development env.

### DIFF
--- a/src/platforms/web/runtime/class-util.js
+++ b/src/platforms/web/runtime/class-util.js
@@ -1,10 +1,17 @@
 /* @flow */
 
+import { warn } from 'core/util/index'
+
 /**
  * Add class with compatibility for SVG since classList is not supported on
  * SVG elements in IE
  */
-export function addClass (el: Element, cls: string) {
+export function addClass (el: Element, cls: ?string) {
+  if (!cls || cls.trim() === '') {
+    process.env.NODE_ENV !== 'production' && warn('Ignoring empty class name.')
+    return
+  }
+
   /* istanbul ignore else */
   if (el.classList) {
     if (cls.indexOf(' ') > -1) {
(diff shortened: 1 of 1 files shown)


src/platforms/web/runtime/class-util.js
```diff
@@ -25,6 +32,11 @@ export function addClass (el: Element, cls: string) {
  * SVG elements in IE
  */
 export function removeClass (el: Element, cls: string) {
```
phanan: You're right, I totally missed this. Adding it in, though maybe we don't need to cater for `null` or `undefined` at all (a "`.trim()` is not a function" error will be thrown).

posva:
@phanan Just to make it clear: this won't fix what the feature he was asking for. He was asking the class to be removed when the value is null. Or did I misunderstood something?

phanan:
No, it won't really "fix" what he asked for, but the DOM exception. The semantic idea (`null` removes the effect) is left undecided, though personally I'm not a fan of it.

sean220:
So,Could you provide a better way to cancel the transition effect dynamically? @phanan

PR #4059, opened 2016-10-29
caseInsensitive maybe more descriptive than expectsLowerCase

jingsam:

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -26,14 +26,14 @@ export function toNumber (val: string): number | string {
  */
 export function makeMap (
   str: string,
-  expectsLowerCase?: boolean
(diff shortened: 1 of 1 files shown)


fnlctrl:
Thanks, but `case-insensitive` is clear enough.
https://en.wikipedia.org/wiki/Case_sensitivity

jingsam:
@fnlctrl As case-insensitive is more clear, why not use caseInsensitive instead?

fnlctrl:
Oops.. Misread changes/original, sorry.
`expectsLowerCase` is also clear enough, and since it's just source code (not documentation), there's no practical to change it.

jingsam:
Understand, but I am reading the source code now 😄

PR #4067, opened 2016-10-30
Update create-bundle-renderer.js

ljanecek:


fnlctrl:
Hi, thanks for this PR, can you explain about it's purpose?

ljanecek:
I think, that it's the way how get the data of instance vue app in node.js.

```
stream.on('app', function(app){
	res.write('<!DOCTYPE html><html><head><title>'+app.$route.meta.title+'</title></head><body>')
});
```

yyx990803:
You can set the `app` on the render `context` and in your server entry file and access it in normal stream event handlers.

PR #4093, opened 2016-11-03
keep consistency in dev and prod env

jingsam:

### DIFF
--- a/src/core/global-api/extend.js
+++ b/src/core/global-api/extend.js
@@ -23,14 +23,13 @@ export function initExtend (Vue: GlobalAPI) {
       return extendOptions._Ctor
     }
     let name = extendOptions.name || Super.options.name
-    if (!/^[a-zA-Z][\w-]*$/.test(name)) {
-      if (process.env.NODE_ENV !== 'production') {
+    if (process.env.NODE_ENV !== 'production') {
+      if (!/^[a-zA-Z][\w-]*$/.test(name)) {      
         warn(
           'Invalid component name: "' + name + '". Component names ' +
           'can only contain alphanumeric characaters and the hyphen.'
         )
       }
-      name = null
     }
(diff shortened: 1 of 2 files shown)


yyx990803:
Does this cause any problem in production?

jingsam:
Noop, I just thought that vue should have the same behaviour in development and production enviroment except extra warnings.

jingsam:
One more thing, I think this [line of code](https://github.com/jingsam/vue/blob/dev/src/core/global-api/extend.js#L32) should be deleted in Vue 2. In vue 1, `name` will be used as a function name, such as:
name='MyComponent' => function MyComponent () {this._init(options)}
However, in Vue 2, no matter what name is, always use  `function VueComponent() {this._init(options)}`

yyx990803:
@jingsam true, I think this PR can be changed to simply removing that line.

jingsam:
Ops, CI failed in this [test](https://github.com/vuejs/vue/blob/dev/test/unit/features/options/name.spec.js#L23-L28):`
```
  it('when incorrect name given it should not contain itself in self components', () => {
    const vm = Vue.extend({
      name: 'Hyper*Vue'
    })

    expect(vm.options.components['Hyper*Vue']).toBeUndefined()
```

In fact, in Vue 2, there are no limitations on the name of components. Even invalid name can be registered. The only limitation is whether the virtual dom  can parse the name tag into  the corresponding component. For example, `Hyper*Vue` can be registered but `<Hyper*Vue></Hyper*Vue>` can not be parsed by virtual dom correctly. `_HyperVue` is an invalid name but it works, as `<_HyperVue></_HyperVue>` can be parsed.

@yyx990803 so, shall I delete that unit test case ?

yyx990803:
Yeah, I think that test case can be removed.

Issue #3832, opened 2016-10-02
let vue-template-compiler work in strict mode


micheletriaca:
Hi,
I really love vue and I would like to use it to customize the user interface of Salesforce. 
The problem is that Salesforce has recently introduced a security feature called LockerService, that enforces strict mode on any 3rd party libraries that are included in the project and in the near future will enforce also CSP. 
I've done some tests and it seems that the only thing that prevent a Salesforce developer from using vue is that the compiler renders the templates using the "with" functionality. 
Using inline templates is not an option because they use 'new Function', violating CSP. 
It seems to me that it should be possible to modify the compiler to be strict mode compliant.. Is this improvement in the roadmap? Thank you for your awesome work

fnlctrl:
Hi, thanks for filling this issue.

`Is this improvement in the roadmap?`
No.. Using `with this` was the design decision in 2.0 to avoid manually parsing the expressions (which bloats the code and takes more time), so it's unlikely to be changed.

However it's relatively easy to remove `with this` (and you'd have to add `this.` to every expression you write in the template afterwards), so maybe it can be an option for the compiler.

For now, you can try using jsx (if you don't have negative opinions about it..), there's a official [babel plugin](https://github.com/vuejs/babel-plugin-transform-vue-jsx) for compiling jsx to vue's render function, which really works great and is much more flexible than templates (and doesn't require `with this`).

yyx990803:
Hmm, pre-compiled templates, although using `with (this)`, is CPS-compliant. It seems a bit over the top to enforce strict mode on user code though.

As of now, removing `with (this)` from generated render functions involves some fundamental changes to the compiler implementation for very marginal gain, so unfortunately I don't think that will happen any time soon.

The alternative solution, as @fnlctrl suggested, is using render function / JSX.

micheletriaca:
Hi @yyx990803 , the problem here is that Salesforce exposes a SPA in which you can develop components.. these components must not directly manipulate the DOM of others components, but should use the "official" way based on events to communicate with other components. Strict mode and other security features are required to keep a component in a sort of sandbox. That said, I've tried with JSX/render functions. It works, but pre-compiled templates are a lot better. You can still be productive even with JSX except for the lack of a valid v-model alternative (you can use a value + on-click binding but is clearly not optimal if you are doing a lot of data entry)... I understand that my situation is probably an edge case, but I'm still very sad to hear that this feature is not in the roadmap

coolaj86:
[...]

PR #4135, opened 2016-11-06
Enable strict mode by removing with

coolaj86:
This is just a proof-of-concept to address #3832.

It isn't yet fully tested, but it will run the markdown demo.

### DIFF
--- a/src/compiler/codegen/events.js
+++ b/src/compiler/codegen/events.js
@@ -24,7 +24,7 @@ const modifierCode = {
 export function genHandlers (events: ASTElementHandlers, native?: boolean): string {
   let res = native ? 'nativeOn:{' : 'on:{'
   for (const name in events) {
-    res += `"${name}":${genHandler(events[name])},`
+    res += `"${name}":v.${genHandler(events[name])},`
   }
   return res.slice(0, -1) + '}'
 }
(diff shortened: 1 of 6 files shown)


yyx990803:
Sorry, but this is more complicated than just prefixing the expression. You need to do proper parsing to prefix all **valid identifiers**, and there's also scope analysis involved inside functions with arguments.

coolaj86:
@yyx990803 would you mind explaining this to me or linking me to a place in the code of what you're talking about?

My experience has been that anything that can be done with `with` and `eval` can be done without them (just like any iterative program can be recursive, right?)

By scope analysis do you mean angular-style dynamic injection?

Angular 1 with ng-strict is currently the only framework that works securely in true JavaScript (i.e. ES5.1) without build tools, but it's so bloated. I really want to use vue.js because it seems like it wouldn't require that much to make strict and secure sans build tools and I'm willing to put in work to make that happen.

yyx990803:
It's true that it **can** be done, but it's not as trivial as you might have thought.

E.g. how do you transform the following:

``` js
with (this) {
  return h('div', {
    class: { a: someVariable },
    children: list.map(item => h('span', item.text))
  }
}
```

1. Notice that the class binding expression is an object literal: `"{ a: someVariable }"` - you'd need to parse the expression to properly prefix `someVariable` instead of the entire expression. And because users can use arbitrarily complex expressions in the template, you essentially need a full JavaScript parser (which is why ng1 is bloated).

2. Notice inside the scope of the map function, `item` is a local variable because of the argument, and it **should not** be prefixed. This requires keeping track of the identifier information in current function scope while doing the prefixing.

So basically, it is possible, but only with a proper JavaScript parse -> AST transform -> codegen process, a la Babel. This is why in https://github.com/vuejs/vue/issues/4115#issuecomment-258525758 I suggest separating the prefixing from the current compiler implementation.

PR #4165, opened 2016-11-09
data/store to normal object

salihsagdilek:
Hello i work on a project with vue.js.
But sometimes i need the raw data.
Actually same problem with this issue; [How to convert the object of Vue to normal object?
](https://github.com/vuejs/Discussion/issues/292)

```javascript
JSON.parse(JSON.stringify(vm.$data))
```
this solution is great but if you need it more than few times it's uncomfortable. so i wrote normalize; 
```javascript
/** vue */
vm.$data.normalize(); // return to normal object

/** vuex */
this.$store.state.normalize(); // return to normal object
```

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -41,6 +41,9 @@ export class Observer {
     this.dep = new Dep()
     this.vmCount = 0
     def(value, '__ob__', this)
+    def(value, 'normalize', function(){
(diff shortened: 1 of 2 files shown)


yyx990803:
Sorry, but I think it's a bad idea to extend data objects like that. This could easily lead to naming conflicts. Just creat util functions of your own.

PR #4224, opened 2016-11-16
remove unused variables

lgabster:


yyx990803:
This is already done in a local branch as part of some additional refactoring, but thanks for the PR.

PR #4226, opened 2016-11-16
fix kebab vs camel-case event names

lgabster:
This PR fix the following problem
```javascript
v-on:click="$emit('myEvent')"
v-on:myEvent="doSomething()"
```

Now `$emit('myEvent')` trigger `v-on:myEvent`, `v-on:myevent`, `v-on:my-event`, and log error message to the console when event isn't hook: and cannot find in registered events.

### DIFF
--- a/src/core/instance/events.js
+++ b/src/core/instance/events.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { bind, toArray } from '../util/index'
+import { bind, toArray, warn } from '../util/index'
 import { updateListeners } from '../vdom/helpers/index'
 
 export function initEvents (vm: Component) {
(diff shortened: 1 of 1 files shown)


yyx990803:
FYI this already works if you are using string templates instead of in-dom templates. We've discussed this in the past but decided that ambiguous event name matching may end up causing subtle bugs.

The short answer is stick to kebab-case if you really have to use in-com templates.

src/core/instance/events.js
```diff
@@ -66,13 +66,17 @@ export function eventsMixin (Vue: Class<Component>) {
 
   Vue.prototype.$emit = function (event: string): Component {
     const vm: Component = this
-    let cbs = vm._events[event]
+    const camelCaseRE = /([a-z0-9._:])([A-Z])/g
+    let cbs = vm._events[event.toLowerCase()] ||
+              vm._events[event.replace(camelCaseRE, '$1-$2').toLowerCase()]
```
sirlancelot: there's a utility function for this already, check out: https://github.com/vuejs/vue/blob/v2.0.7/src/shared/util.js#L103

lgabster:
OK, but now there is no any warning message, when event cannot find in vm._events

yyx990803:
@lgabster that's a good point. I think what we can do is:

- In dev mode, if an event misses, check for a loose match. If there is a loose match, warn the user about the event casing problem.

PR #4232, opened 2016-11-17
destroyed elements pass their styles onto others that are similar

lgdexter:
test to confirm issue #4227


defcc:
Got it. I created a PR  #4235 to fix it

PR #4239, opened 2016-11-18
Fix v-move class when name isn't specified

decademoon:
If `name` isn't specified for a `<transition-group>`, then the class name vue uses for moving items will be `undefined-move` instead of `v-move`.


yyx990803:
Thanks!

PR #4267, opened 2016-11-21
Modifier once for v-on

KingMario:
Modifier once for v-on is added.

### DIFF
--- a/src/platforms/web/runtime/modules/events.js
+++ b/src/platforms/web/runtime/modules/events.js
@@ -3,9 +3,6 @@
 
 import { updateListeners } from 'core/vdom/helpers/index'
 
-const onceFlags = {}
-const randomHex = () => (Math.random() + '').substr(2).toString(16)
-
 function updateDOMListeners (oldVnode, vnode) {
   if (!oldVnode.data.on && !vnode.data.on) {
     return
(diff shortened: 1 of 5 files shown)


test/unit/features/directives/on.spec.js
```diff
@@ -102,6 +102,24 @@ describe('Directive v-on', () => {
...
+  it('should support once', () => {
+    const numberPushed = []
+    vm = new Vue({
+      el,
+      template: `
+        <div @click.once="foo">
+        </div>
+      `,
+      methods: {
+        foo () { numberPushed.push(1) }
```
defcc: you could create a jasmine.createSpy(),  and use `expect(spy).not.toHaveBeenCalled()` to check it.

src/platforms/web/runtime/modules/events.js
```diff
@@ -3,13 +3,27 @@
...
   const on = vnode.data.on || {}
   const oldOn = oldVnode.data.on || {}
-  const add = vnode.elm._v_add || (vnode.elm._v_add = (event, handler, capture) => {
+  const add = vnode.elm._v_add || (vnode.elm._v_add = (event, handler, capture, once) => {
+    if (once) {
+      const randomKey = randomHex() + new Date().valueOf().toString(16) + randomHex()
+      onceFlags[randomKey] = false
+      const oldHandler = handler
+      handler = () => {
+        if (!onceFlags[randomKey]) {
```
KingMario: That's a good question.

Actually, the pr is way not well-tested. I'm working on it.

test/unit/features/directives/on.spec.js
KingMario: Wait for it, sir.

KingMario: Because the second click will be triggered very fast in the unit test, the spy will be called twice even if the handler is removed in the first call. Could you please tell me how to hold the test for a while between the two clicks.

Will there be any side effect if I use setTimeout for the second click?

KingMario:
```
    triggerEvent(vm.$el, 'click')
    expect(spy).toHaveBeenCalled()
    triggerEvent(vm.$el, 'click')
    expect(spy).not.toHaveBeenCalled()
```
Will the second assertion be misjudged by the first call?

```
  it('should support capture and once', () => {
    const callOrder = []
    vm = new Vue({
      el,
      template: `
        <div @click.capture.once="foo">
          <div @click="bar"></div>
        </div>
      `,
      methods: {
        foo () { callOrder.push(1) },
        bar () { callOrder.push(2) }
      }
    })
    triggerEvent(vm.$el.firstChild, 'click')
    expect(callOrder.toString()).toBe('1,2')
    triggerEvent(vm.$el.firstChild, 'click')
    expect(callOrder.toString()).toBe('1,2,2')
  })
```
The assertions in this test both success.

KingMario: Figured out that, in this case, should use spy.calls.count().

KingMario:
@defcc Finished. Please review again. Thanks.

KingMario:
Anybody home? Please review it. Thanks.

KingMario:
As in the commits, there is a **bug fix** for the removal of event handlers.

According to MDN,
> Removal of a capturing listener does not affect a non-capturing version of the same listener, and vice versa.

PR #4271, opened 2016-11-21
Support v-if multiple conditions

defcc:
Support v-if multiple conditions

```html
<div v-if="tester1"></div>
<div v-elseif="tester2"></div>
<div v-elseif="tester3"></div>
<div v-else></div>
```

Collects `if` and `v-elseif` and `v-else` into element.conditions when parsing and generates code like (a)?_h:(c)?_h:_e()


blake-newman:
`v-elseif` or `v-else-if`?

Latter might be better for non-native users

flow/compiler.js
```diff
@@ -41,6 +41,7 @@ declare type ModuleOptions = {
 }
 
 declare type ASTModifiers = { [key: string]: boolean }
+declare type ASTIfConditions = Array<{ exp: ?string; block: ASTElement }>
```
zzz6519003: what's this doing

test/unit/features/directives/if.spec.js
```diff
@@ -88,6 +88,55 @@ describe('Directive v-if', () => {
...
+          <span v-if="foo">hello</span>
+          <span v-elseif="bar">elseif</span>
+          <span v-else>bye</span>
+        </div>
+      `,
+      data: { foo: true, bar: false }
+    }).$mount()
+    expect(vm.$el.innerHTML.trim()).toBe('<span>hello</span>')
+    vm.foo = false
+    waitForUpdate(() => {
```
zzz6519003: what's the waitForUpdate magic

test/unit/features/directives/if.spec.js
```diff
@@ -88,6 +88,55 @@ describe('Directive v-if', () => {
...
+      expect(vm.$el.innerHTML.trim()).toBe('<span>bye</span>')
+      vm.bar = true
+    }).then(() => {
+      expect(vm.$el.innerHTML.trim()).toBe('<span>elseif</span>')
+      vm.bar = false
```
zzz6519003: if it wants a then chaining,
```js
+      expect(vm.$el.innerHTML.trim()).toBe('<span>elseif</span>')
+      vm.bar = false
```
should return a promise,
but I see no promise T.T

zzz6519003: @fnlctrl help wanted XD

test/unit/features/directives/if.spec.js
```diff
@@ -88,6 +88,55 @@ describe('Directive v-if', () => {
...
+    }).then(() => {
+      expect(vm.$el.innerHTML.trim()).toBe('<span>hello</span>')
+      vm.foo = false
+      vm.bar = {}
+    }).then(() => {
+      expect(vm.$el.innerHTML.trim()).toBe('<span>elseif</span>')
+      vm.bar = 0
+    }).then(() => {
+      expect(vm.$el.innerHTML.trim()).toBe('<span>bye</span>')
+      vm.bar = []
```
zzz6519003: why would it be necessary for so many kind of `vm.bar` @fnlctrl

PR #4297, opened 2016-11-23
Fix nested templates in SFC

blake-newman:
If a tag was nested inside a template it would cause the initial block to be considered ended

- Check depth is 0 when checking special tag in SFC
- Moved depth decrement on SFC parser to make use clearer
- Updated tests to ensure regression is covered

Edit: See below comment

### DIFF
--- a/src/compiler/parser/html-parser.js
+++ b/src/compiler/parser/html-parser.js
@@ -48,13 +48,14 @@ let IS_REGEX_CAPTURING_BROKEN = false
 // Special Elements (can contain anything)
 const isScriptOrStyle = makeMap('script,style', true)
 const hasLang = attr => attr.name === 'lang' && attr.value !== 'html'
-const isSpecialTag = (tag, isSFC, stack) => {
+const isSpecialTag = (tag, isSFC, stack, depth) => {
   if (isScriptOrStyle(tag)) {
     return true
   }
-  if (isSFC) {
+  // Check SFC block if at depth 0
+  if (isSFC && depth === 0) {
     // top-level template that has no pre-processor
-    if (tag === 'template' && stack.length === 1 && !stack[0].attrs.some(hasLang)) {
+    if (tag === 'template' && !stack[0].attrs.some(hasLang)) {
       return false
     } else {
       return true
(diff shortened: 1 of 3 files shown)


src/compiler/parser/html-parser.js
```diff
@@ -48,13 +48,14 @@ let IS_REGEX_CAPTURING_BROKEN = false
...
-const isSpecialTag = (tag, isSFC, stack) => {
+const isSpecialTag = (tag, isSFC, stack, depth) => {
   if (isScriptOrStyle(tag)) {
     return true
   }
-  if (isSFC) {
+  // Check SFC block if at depth 0
+  if (isSFC && depth === 0) {
     // top-level template that has no pre-processor
-    if (tag === 'template' && stack.length === 1 && !stack[0].attrs.some(hasLang)) {
```
posva: Isn't this actually required or can't the stack ever be empty?

test/unit/modules/sfc/sfc-parser.spec.js
```diff
@@ -87,7 +90,7 @@ describe('Single File Component parser', () => {
         <my-button ref="button">Hello</my-button>
       </example>
       <example name="with props">
-        <my-button color="red">Hello</my-button>
+        <template lang="en" color="red">Hello</template>
```
posva: I suggest to add this into another test, something like `doesn't parse template inside custom blocks`

test/unit/modules/sfc/sfc-parser.spec.js
```diff
@@ -99,7 +102,7 @@ describe('Single File Component parser', () => {
       }
       </test>
     `)
-    expect(res.customBlocks.length).toBe(3)
```
posva: This is actually the important part of the test 😆 
As said below, I think we should just add another test

blake-newman:
@yyx990803 

@defcc suggested that we may not need this isSpecialTag check. Seems to work, is there any reason that it should still exist? Can reverse commit if it is required.

src/compiler/parser/html-parser.js
```diff
@@ -90,8 +74,8 @@ export function parseHTML (html, options) {
   let last, lastTag
   while (html) {
     last = html
-    // Make sure we're not in a script or style element
-    if (!lastTag || !isSpecialTag(lastTag, options.sfc, stack)) {
+    // Make sure we're not in a specia tag
```
blake-newman: Need to change this line back, will wait for feedback from Evan first

defcc:
the `<` in template seems to have been fixed with this [commit](https://github.com/vuejs/vue/commit/e2a361a052dfdfb3f531e9a8e26b12d48309f6b9) ? So there may be no need to do the sfc check.

yyx990803:
Turns out the fix could be easier: https://github.com/vuejs/vue/commit/9d0bc9768b40fa5fab3a648724f50385f2552ff7

But thanks a lot for looking into this!

blake-newman:
No problem! Super simle

PR #4325, opened 2016-11-27
Add some functions related to render

kaorun343:
Although these methods' prefix is `_`, using them are beneficial when writing `render`, I think.
So, I added them into the definitions.

### DIFF
--- a/types/test/vue-test.ts
+++ b/types/test/vue-test.ts
@@ -47,6 +47,23 @@ class Test extends Vue {
     });
     this.$nextTick().then(() => {});
     this.$createElement("div", {}, "message");
+
+    this._h('div', 'Hello world.')
+    this._s(true)
+    this._n('10')
+    this._e()
+    this._q(true, false)
+    this._i([true], true)
+    this._m(1, false)
+    this._o(this._e(), 0, 'key')
+    this._o([this._e()], 0, 'key')[0]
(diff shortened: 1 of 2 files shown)


ktsn:
Can you show any use cases by using these methods?
I'm not sure why we need to expose these private apis. `$createElement` is not suffice for you?

kaorun343:
I don't need all of these private methods. What I really need are these below. 

* `_m()`  to access to `staticRenderFns`
* `_l()`  We don't have to prepare the functions like `_.map()`
* `_f()` to access to the filters registered globally, some plugins are doing so
* `_k()` We don't have to know what the keyCode is

Also,  I tried to define these apis as `protected`. However, when specify `this`, I cannot access to these protected methods.

ktsn:
Well, I meant that I want you to show some **practical use cases**.
I think we shouldn't expose private apis for use cases that can be achieved easily on the userland. Because it would complicate to maintain declaration files as they can be changed without updating major version. And if we really want to use them for some minor usage, we can declare them in our own project easily.

So, I'm still not sure when/why we need to use these apis rather than `$createElement` and other userland utilities.

kaorun343:
OK, I'll close this 'cause I couldn't show the merits of this PR.
Also I didn't think about the risks of putting them, private apis,  as public.

PR #4349, opened 2016-11-30
fix static style parse error (fix #4346)

### LINKED ISSUE #4346: Template compilation error in style attribute with url
I ran into this issue yesterday. I'm working on a page that includes images with CSS shape-outside for flowing the text around them based on a shape image. To do this, I include the URL to the shape image in an inline style attribute.

During my work this stopped working after I upgraded Vue to the latest. The earliest reproducible version is 2.0.6. It's working on 2.0.5. When the HTML is compiled, the url from the style attribute string gets broken, and instead of the full URL you just end up with "http".

### Vue.js version
2.0.6+

### Reproduction Link
https://jsfiddle.net/mottokrosh/5Lppm3ce/7/

### Steps to reproduce
Go to the above JSBin. Ignore the CORS error about failing to load the shape image. 

### What is Expected?
Inspect the HTML of the `<figure>` in the dev console. The full `style="shape-outside: url(https://grove.sidhewoods.com/assets/verdant-reaches-players-guide/art/NPC_Sheriff_Lucas_Quint-shape.png); shape-margin: 1rem;"` should be seen, as in the specified HTML. 

Instead, it's become `style="shape-outside: url("https"); shape-margin: 1rem;"`.

defcc:
fix #4346 

Remove the unproper hasBackground check and add more test cases.

### DIFF
--- a/src/platforms/web/util/style.js
+++ b/src/platforms/web/util/style.js
@@ -4,10 +4,8 @@ import { cached, extend, toObject } from 'shared/util'
 
 export const parseStyleText = cached(function (cssText) {
   const res = {}
-  const hasBackground = cssText.indexOf('background') >= 0
-  // maybe with background-image: url(http://xxx) or base64 img
-  const listDelimiter = hasBackground ? /;(?![^(]*\))/g : ';'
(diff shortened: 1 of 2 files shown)


yyx990803:
👍

Issue #4347, opened 2016-11-30
Ctrl modifier with enter key is not working


allesan:
### Vue.js version
2.1.3

It looks like Ctrl modifier with Enter key is not working.
This works <textarea @keyup.alt.67="clear">
but <textarea @keyup.ctrl.enter="clear"> fires on Enter key no matter if ctrl is pressed or not.

### What is Expected?

With <textarea @keyup.ctrl.enter="clear"> it's expected to trigger "clear" only when Ctrl key is pressed down and hit Enter.

### What is actually happening?

"clear" is triggered every time Enter is hit, wether Ctrl is pressed or not.

defcc:
Related to this commit https://github.com/vuejs/vue/commit/b45b974a5c866abd8dc777ccac511af30b72b91b#diff-cd513113ac480c30cffa092f5fe266ed

Maybe it should not be limited to mouse events ?  It's a common usage in keyboard events handling.

cc @yyx990803

yyx990803:
The primary reason I did that is because allowing it for key events would break the following:

``` js
Vue.config.keyCodes.ctrl = 17
```

Although the above usage seems rare, this would still technically be a breaking change.

defcc:
We may need document it here https://vuejs.org/v2/guide/events.html#Modifier-Keys. I'll update the doc.

yyx990803:
Alternatively, we can add a deprecation warning when the user attempts to define a built-in modifier as a custom keyCode alias.

defcc:
I think that would be better, considering the common usage of keyboad event. I'll try to make a update.

defcc:
@allesan , currently you might need to deal with the special key in your event handle like `if(!evt.ctrlKey)return` and that's what vue does internally.

KingMario:
I didn't noticed that limit and accidentally modified the events document by adding the keyboard events, note on meta key and samples.

Should have studied the codes and the release commit list before modifying the document. Feel sorry.

allesan:
@defcc thanks, that works.

defcc:
Closed with https://github.com/vuejs/vue/commit/e8d6bd9dc73cacbc665ce132d81a7b629bef764b

And we need update the relative doc, will track it in the vuejs.org issues.

zzjalpaca:
https://user-images.githubusercontent.com/19635504/53807719-40b33080-3f8b-11e9-97cb-116cb3ca6fbc.png
sdad

PR #4356, opened 2016-12-01
Modifier key v-on modifier for keyboard event && Spelling correct &&…

KingMario:
Modifier keys v-on modifier for keyboard event.

Spelling correct.

Invalid v-on modifier key alias warning.

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -1,8 +1,37 @@
 import Vue from './instance/index'
 import { initGlobalAPI } from './global-api/index'
 import { isServerRendering } from 'core/util/env'
+import { warn } from 'core/util/index'
+import { modifierCode } from '../compiler/codegen/events'
 
 initGlobalAPI(Vue)
+
+for (const key in modifierCode) {
+  Vue.config.keyCodes[key] = undefined
+}
+
+const configWatcherVM = new Vue({
+  data: {
+    config: Vue.config
+  },
+  watch: {
+    'config.keyCodes': {
+      handler: (val, oldVal) => {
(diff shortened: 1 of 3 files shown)


KingMario:
https://github.com/vuejs/vue/issues/4347 addressed @defcc

defcc:
@KingMario I think you could check the special keycode alias in the `_k` function and introduce the warning once.

KingMario:
@defcc What if `_k` is never called? I guess a global triggered warning can educate the user whenever a built-in alias is defined.

Though the watcher in a dedicated Vue instance solution is no so neat.

src/core/index.js
```diff
@@ -1,9 +1,38 @@
 import Vue from './instance/index'
 import { initGlobalAPI } from './global-api/index'
 import { isServerRendering } from 'core/util/env'
+import { warn } from 'core/util/index'
+import { modifierCode } from '../compiler/codegen/events'
```
yyx990803: Modules under `/core` should not import from modules under `/compiler`. The only coupling between core and compiler happens in entry files.

src/core/index.js
```diff
@@ -1,9 +1,38 @@
...
 
+for (const key in modifierCode) {
+  Vue.config.keyCodes[key] = undefined
+}
+
+new Vue({
+  data: {
+    config: Vue.config
+  },
+  watch: {
```
yyx990803: this is definitely less than ideal...

yyx990803:
Implemented the warning with Proxy: e8d6bd9

KingMario:
受教了，谢谢大神！

PR #4378, opened 2016-12-03
Reference doc in component data function warning

chrisvfritz:
The reason `data` has to be a function in components seems to be a point of confusion for many, so I've linked to an explanatory example in the warning.


chrisvfritz:
@yyx990803 I don't remember if you're getting email notifications of PRs on the main repo anymore, so tagging you. 🙂

yyx990803:
Yeah I still get notifications for the main repo :)

Issue #4369, opened 2016-12-02
vue-template-compiler does not bind value to select elements


rmp135:
Take the following
```html
<div>
  <select :value='something'></select>
</div>
```
Currently, vue-template-compiler generates the render function with `{attrs:{"value":something}}`. However, vue supports the `v-bind:value` syntax with selects. I would think that instead, it should generate with `{domProps:{"value":something}}` in a similar manner to inputs.

This means no bindings can take place on the select. This works with a `v-model`, however this causes vuex to attempt to update the store without going through a mutator.

Looks like it was caused by the fix to #4233 that filters which tag / attribute combos are bound as props.

Issue #4233, opened 2016-11-17
Attribute doesn't update if binded to a property called `selected`


hay:
If you have an element that is binded to an attribute called `selected` it won't update in the DOM. For example:

    <p v-bind:selected="selected">selected is {{selected}}</p>
    <button v-on:click="selected = !selected">toggle</button>

Note that the `{{selected}}` template works fine, and the variable updates as well. It's just the attribute that doesn't update. 

Here's a codepen that reproduces the problem: http://codepen.io/hay/pen/mOOrWK

Problem occurs in Vue 2.0.7. This example works fine in 1.0.28.

yyx990803:
Selected is bound as a property instead of an attribute. P tags don't have selected attribute, why would you want to bind that?

hay:
This is indeed for styling purposes. But why should you be using a class? Attributes can be styled in CSS, and sometimes they're more useful than classes.

hay:
Just checked and the patch in the 2.0.8 fixed this. Thanks so much!

PR #4382, opened 2016-12-04
bind value as property for <select> when use template compiler

j3l11234:
Hi,

I found the same problem [#4233](https://github.com/vuejs/vue/issues/4369) today, and found the source of the problem. 
I appreciate your timely patch, But maybe I think something was missed, so I added it.

Btw, I have used vue for a week and which used React previously, I think vue is cooler~
Thinks!


defcc:
@j3l11234 , the codes in the package are generated automatically when new version released. They should not be changed by hand.

j3l11234:
@defcc Thanks.

PR #4383, opened 2016-12-04
add ignore .idea directory

sampsonli:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist/vue.common.min.js
 test/e2e/reports
 test/e2e/screenshots
 coverage
+.idea


posva:
yeah, why not. It makes things easier for contributors working with idea IDEs

JounQin:
😂not only `.idea`, but also `.iml`.

posva:
@JounQin I've never used Idea's IDEs so having some user input on this is nice, thanks 👍

fnlctrl:
Adding `.idea` should suffice though, since all `.iml` files are stored under the `.idea` directory

JounQin:
@fnlctrl Not exactly. I'm using IDEA Modules in one project, what means the git root path is not same with `.idea` but is with `*.iml`.

petervmeijgaard:
What's the deal with [creating a global `.gitignore`](http://stackoverflow.com/questions/7335420/global-git-ignore) and including the `.idea`-directory in there? You would never have to worry about accidentally committing the `.idea`-directory.

And what about other IDEs and editors? When you are developing in Visual Studio Code, the `.vscode`-directory is created. Same goes for NetBeans with the `nbproject`-directory. Should these directories be added to the `.gitignore` as well?

I don't think it's the responsibility for Vue to ignore these directories. To exclude all these directories, the `.gitignore`-file would get a bit too bloated.

posva:
@petervmeijgaard thanks for sharing that, didn't know about it!
Makes sense not to merge this PR because we would have to add every ide, text editor specific folder to .gitignore and because there's a better solution, which is creating a global .gitignore 👍

PR #4402, opened 2016-12-07
Improve elm.value update (fix #4392)

### LINKED ISSUE #4392: Input [v-model & type=number] is truncated when (unrelated) DOM update happens.
### Vue.js version
v2.1.3

### Reproduction Link
http://jsfiddle.net/1y2nk1L9/5/

### Steps to reproduce
Enter a number with some zeros after the comma into a type="number" input field, while Vue performs some page rendering in the background with (unrelated) variables of the same component.

### What is Expected?
Entering a number like 1.00001 works without issues.

Every time Vue triggers a DOM change all v-model input fields are "re-rendered" ("0.000" -> "0") which makes it impossible to enter decimal numbers that contain zeros in busy components. While this sounds like a minor issue it is a **huge problem**, since the entered number could be completely wrong. Imagine entering 1.00345, but the input is silently changed to 1345. This has already led to some serious trouble in our (financial) Vue application.

defcc:
fix #4392

We should update the elm.value when in these cases:
1. the elm is not active and the value is not equal to model.value
2. the elm needs to be casted to number or trimed, and the casted value is not equal to model.value


src/platforms/web/util/model.js
```diff
@@ -0,0 +1,19 @@
...
+  const directives = vnode.data.directives || []
+  for (let i = 0, directive; i < directives.length; i++) {
+    directive = directives[i]
+    if (directive.name === 'model') {
+      return directive.modifiers
+    }
+  }
+}
+
+export function isToNumber (modifiers: ?ASTModifiers, type: ?string): boolean {
```
phanan: I guess you can use a better name, as `isToNumber` isn't the most meaningful of them IMHO.

src/platforms/web/util/model.js
```diff
@@ -0,0 +1,19 @@
...
+  return modifiers && modifiers.number || (type === 'number')
```
phanan: The brackets are unnecessary, aren't they?

src/platforms/web/util/model.js
```diff
@@ -0,0 +1,19 @@
...
+
+export function getModelModifier (vnode: VNodeWithData): ASTModifiers | void {
+  const directives = vnode.data.directives || []
+  for (let i = 0, directive; i < directives.length; i++) {
+    directive = directives[i]
+    if (directive.name === 'model') {
+      return directive.modifiers
+    }
+  }
+}
```
kazupon: I think that this function is not problem, however, If modifier did not find, in end of function, 
when you explicitly return, we can easily understand about this function.

defcc: Added an explictly `return undefined` statement

src/platforms/web/util/model.js
defcc: I think so, I'll try to find a better one

defcc:
@yyx990803  CI dosen't run, will you please help to have a check on it ?

src/platforms/web/runtime/modules/dom-props.js
```diff
@@ -44,6 +47,31 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
   }
 }
 
+function isValueChanged (vnode: VNodeWithData, val: string): boolean {
+  const value = vnode.elm.value
+  const modifiers = getModelModifier(vnode)
+  const needNumber = (modifiers && modifiers.number) || vnode.elm.type === 'number'
+  const needTrim = modifiers && modifiers.trim
+  if (needNumber) {
+    return toNumber(value, val) !== toNumber(val)
```
posva: the first toNumber has an extra argument

src/platforms/web/runtime/modules/dom-props.js
```diff
@@ -44,6 +47,31 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
...
+
+function getModelModifier (vnode: VNodeWithData): ASTModifiers | void {
+  const directives = vnode.data.directives || []
+  for (let i = 0, directive; i < directives.length; i++) {
+    directive = directives[i]
+    if (directive.name === 'model') {
+      return directive.modifiers
+    }
+  }
+  return undefined
```
posva: Isn't better to return an empty object so at the top you can do `modifiers.number` instead of `modifiers && modifiers.number`?

src/platforms/web/runtime/modules/dom-props.js
defcc: I'll correct it

src/platforms/web/runtime/modules/dom-props.js
defcc: Yes, thanks

yyx990803:
I found an easier way: simply prevent updating when in focus, and force update on blur (see https://github.com/vuejs/vue/commit/8a7b02a159dea19ff76c2308bceb9f1ef4738998)

Thanks for the PR still, used your test cases :)

yyx990803:
Hmm, looks like the `isValueChanged` check is indeed needed (e2e tests failing after my patch). Wondering if there's a way to simplify it...

yyx990803:
I ended up including your `isValueChanged` check to ensure correct behavior in e2e, sorry for closing it!

PR #4435, opened 2016-12-10
fix v-bind.prop parse (fix #4432)

### LINKED ISSUE #4432: [2.x] v-bind not removing attributes with falsy values when another attribute with `.prop` modifier appears first
Howdy!

It seems like `v-bind` is not removing attributes with falsy values when another attribute with the `.prop` modifier appears first in the element's list of attributes. More details in the comments of the fiddle.

### Vue.js version
2.x (Tested with 2.1.3 and 2.1.4)

### Reproduction Link
http://jsfiddle.net/JosephusPaye/df4Lnuw6/78/

### Steps to reproduce
1. Run the fiddle
2. Inspect the rendered elements using devtools to see the problem described

### What is Expected?
`v-bind` should remove attributes whose values are falsy (`false`, `null`, `undefined`) regardless of the presence of other attributes and the `.prop` modifier.

Attributes with falsy values are not removed when another attribute with the `.prop` modifier appears first. Instead, these attributes are rendered with the string `"false"`, `"null"` or `"undefined"` as their value.

defcc:
fix #4432

### DIFF
--- a/test/unit/features/directives/bind.spec.js
+++ b/test/unit/features/directives/bind.spec.js
@@ -121,6 +121,18 @@ describe('Directive v-bind', () => {
     expect(vm.$el.children[1].innerHTML).toBe('<span>qux</span>')
   })
 
+  it('.prop modifier with normal attribute bind', () => {
+    const vm = new Vue({
+      template: '<input :some.prop="some" :id="id">',
+      data: {
+        some: 'hello',
+        id: 'world'
+      }
+    }).$mount()
+    expect(vm.$el.some).toBe('hello')
+    expect(vm.$el.getAttribute('id')).toBe('world')
+  })
+
   it('.camel modifier', () => {
     const vm = new Vue({
       template: '<svg :view-box.camel="viewBox"></svg>',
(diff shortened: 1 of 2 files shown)


test/unit/features/directives/bind.spec.js
```diff
@@ -121,6 +121,18 @@ describe('Directive v-bind', () => {
     expect(vm.$el.children[1].innerHTML).toBe('<span>qux</span>')
   })
 
+  it('.prop modifier with normal attribute bind', () => {
+    const vm = new Vue({
+      template: '<input :some.prop="some" :id="id">',
+      data: {
+        some: 'hello',
+        id: 'world'
```
posva: Actually, the test case should use a null or false value and check that the id attribute is absent on the element

posva:
I can't believe the fix was that easy 😆 
Good job again @defcc 👍

test/unit/features/directives/bind.spec.js
defcc: Thank, I updated the test case

defcc:
For the test case, the `id` attribute is special, it reflects both with attribute and property.

PR #4450, opened 2016-12-12
expected consistent spacing

mehwww:


defcc:
There should be no space before '}'

PR #4475, opened 2016-12-14
Fix transition + v-show display style changes (fix #4473)

### LINKED ISSUE #4473: Breaking change in 2.1.5 involving v-show and transition
### Vue.js version
2.1.5+

### Reproduction Link
- <s>http://codepen.io/jbruni/pen/JbmZrj</s> (original pen, still valid)
- http://codepen.io/jbruni/pen/vyVzao (new pen, simplified)

### Steps to reproduce
1. Please check the CodePen linked above. It contains the code working in 2.1.4, with simple instructions.
2. Next, click the settings button in the JS pane, update the VueJS version in the CDN link, and save.
3. Try again, and you will immediately see the difference in behaviour.

### What is Expected?
The behaviour shown in the reproduction link should keep working.

After the upgrade to 2.1.5 (or 2.1.6), it does not work anymore.

-----

Based on the release notes, I made a build creating a new branch, from v2.1.6 tag checkout, and then reverting this commit: vuejs/vue@612d32105a4ba1d636e8b617cbcb0a3e59919a9e, and I got things working again. This commit is a fix to vuejs/vue#4418, but the breaking change is coming from it.

**NOTE**: *This note and the new pen in the "Reproduction Link" above are edits to the original submission. Everything else is the same. Below, I will inform details of the original case. They are still relevant, but not essential. The new and simplified case shall suffice.*

As one can see in the reproduction link, as well as at [this gist](https://gist.github.com/jbruni/6a879ae2313c2713c71881617089ff23), we have a directive called `v-click-outside`.  This directive runs the VM method passed as the directive's attribute value, when a click happens outside the DOM element using the directive. We are using it to close dropdown menus when a click is done anywhere outside it. The dropdowns are using `<transition>` to show/hide.

I'm not sure if our implementation relies in the previous imperfect implementation, and we shall fix our code, or if it really should keep working as it has been written - and the fix done in Vue. I have not taken the time to investigate further. By now, I have only identified, reported and traced down the issue to this report.

Thank you.

jbruni:
Please check the comments in the originating issue: vuejs/vue#4473 - thank you.

### DIFF
--- a/src/platforms/web/runtime/modules/transition.js
+++ b/src/platforms/web/runtime/modules/transition.js
@@ -122,8 +122,8 @@ export function enter (vnode: VNodeWithData, toggleDisplay: ?() => void) {
   }
 
   if (vnode.data.show) {
-    toggleDisplay && toggleDisplay()
     enterHook && enterHook(el, cb)
+    toggleDisplay && toggleDisplay()
   }
 
   if (!expectsCSS && !userWantsControl) {
--- a/test/unit/features/transition/transition.spec.js
+++ b/test/unit/features/transition/transition.spec.js
@@ -285,7 +285,6 @@ if (!isIE9) {
             beforeLeaveSpy(el)
           },
           leave: (el) => {
-            expect(el.style.display).toBe('')
             onLeaveSpy(el)
           },
           afterLeave: (el) => {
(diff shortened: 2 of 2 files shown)


yyx990803:
Thanks, see https://github.com/vuejs/vue/issues/4473#issuecomment-267070242

PR #4477, opened 2016-12-14
fixed error of can't read length of null when use broadcast and there no child component

lmf:
当子组件是空的时候，调用$broadcast会报can't read length of null错误

### DIFF
--- a/src/instance/api/events.js
+++ b/src/instance/api/events.js
@@ -137,6 +137,7 @@ export default function (Vue) {
     // then there's no need to broadcast.
     if (!this._eventsCount[event]) return
     var children = this.$children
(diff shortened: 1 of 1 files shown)


src/instance/api/events.js
```diff
@@ -137,6 +137,7 @@ export default function (Vue) {
     // then there's no need to broadcast.
     if (!this._eventsCount[event]) return
     var children = this.$children
+    if (typeof children === null) return
```
mike-grinspan: Does the typeof operator ever actually return null? Shouldn't this be `if(children === null)`?

lmf: You're right, I'll close this and re-create

PR #4490, opened 2016-12-15
Add unbind to v-show (fix #4484)

### LINKED ISSUE #4484: node with v-show inside v-if is reused in the adjacent v-else
### Vue.js version
2.1.6

### Reproduction Link

http://codepen.io/dolymood/pen/gLQmwy

### Steps to reproduce

Code

```html
<div class="el el-if" v-if="bln"><span v-show="false">hidden content</span></div>
<div class="el el-else" v-else><span @click="toggle">visible  content</span></div>
```

1. `bln=true`, the `span` element in `el-if` is hidden

2. `bln=false`, the `span` element in `el-else` is __hidden__

### What is Expected?

When `bln=false`, the `span` element should be __visible__

`patchVnode` will check the nodes is same:

```js
function sameVnode (vnode1, vnode2) {
  return (
    vnode1.key === vnode2.key &&
    vnode1.tag === vnode2.tag &&
    vnode1.isComment === vnode2.isComment &&
    !vnode1.data === !vnode2.data
  )
}
```

Because the old vnode `vnode1.data` is 

```js
{
  directives: [{
     def: {bind: fn, update: fn}
     expression: "false"
     modifiers: {},
     name: "show",
     rawName: "v-show",
     value: false
  }]
}
```

and the new vnode `vnode2.data` is

```js
{
  on: {click: fn}
}
```

So `sameVnode(span1, span2)` result  is `true`, but `updateDirectives` will be call `v-show` directive's `unbind` hook function.

If the `span` elements have different `key` attributes, the result will be correct. But i think it will be better if:

> When `bln=false`, the `span` element should be __visible__

defcc:
fix #4484 

If element exists in dom tree, will set its display property back.

Apply possible transition if value is false.


yyx990803:
Thanks @defcc - there are a few issues with this fix: if we always reset the display in `unbind`, then elements hidden by `v-show` will be revealed if it is inside an element with a leaving transition. So we need to differentiate between a destroy `unbind` vs. a patch `unbind`, and only reset display for patch `unbind`. Transition check is not necessary for patch unbind because a patch unbind cannot happen to common child of the same `<transition>` component.

defcc:
Thanks for your comment.

PR #4517, opened 2016-12-19
fix type definition links

kazupon:
NOTE:
Everytime, we tried not to change links.


yyx990803:
Maybe we can just link to the file, e.g.
```
refer to the `ModuleOptions` type in [flow declarations](path/to/file)
```

PR #4525, opened 2016-12-20
Specify that tags with side-effects will not be parsed.

brycepj:
Where I work, it's not uncommon to come across `<script>` tags in templates that would not be trivial to move into separate files. If Vue will prevent the execution of the contents of these tags, I think the warning should say so. As it is, it reads more like a recommendation than a requirement:

`Templates should only be responsible for mapping the state to the UI. Avoid placing tags with side-effects in your templates, such as <script>`. 

I don't know if the language I've added is best, but it's an example of text that tells the user they should not expect their code to run. Feel free to edit as desired.

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -91,7 +91,7 @@ export function parse (
         process.env.NODE_ENV !== 'production' && warn(
           'Templates should only be responsible for mapping the state to the ' +
           'UI. Avoid placing tags with side-effects in your templates, such as ' +
-          `<${tag}>.`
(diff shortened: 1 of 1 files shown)


nyl-auster:
Is that why i can not embed gists in my blog posts ? A gists is embedded with 
```<script src="https://gist.github.com/nyl-auster/445b557a17197e3c6771af5c0554e439" />```
For now i have to put simple links instead of embedding code in my site : https://yineo.fr/blog/animer-vos-transitions-de-page-facilement-avec-nuxt-js-et-vue-js

Is there a wordaround for this ?

PR #4528, opened 2016-12-20
Update loose equal function, check toString value for primitive type (fix #4514)

### LINKED ISSUE #4514: sometime select's event does't work when option's vlaue in ('',0)
### Vue.js version
2.1.6

### Reproduction Link
http://sandbox.runjs.cn/show/xu7gfzz2

### Steps to reproduce

switch tag `option` between value='' and value='0', then to see ` {{ selected }}`'s change.

### What is Expected?

the value of 'selected' always changes with your choose.

sometime works, sometime does't work.

defcc:
fix #4514

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -208,13 +208,15 @@ export function genStaticKeys (modules: Array<ModuleOptions>): string {
  * if they are plain objects, do they have the same shape?
  */
 export function looseEqual (a: mixed, b: mixed): boolean {
-  /* eslint-disable eqeqeq */
-  return a == b || (
-    isObject(a) && isObject(b)
-      ? JSON.stringify(a) === JSON.stringify(b)
-      : false
-  )
-  /* eslint-enable eqeqeq */
+  const isObjectA = isObject(a)
+  const isObjectB = isObject(b)
+  if (isObjectA && isObjectB) {
+    return JSON.stringify(a) === JSON.stringify(b)
+  } else if (!isObjectA && !isObjectB) {
+    return String(a) === String(b)
+  } else {
+    return false
+  }
 }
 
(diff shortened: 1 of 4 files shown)


src/shared/util.js
```diff
@@ -208,13 +208,15 @@ export function genStaticKeys (modules: Array<ModuleOptions>): string {
...
-    isObject(a) && isObject(b)
-      ? JSON.stringify(a) === JSON.stringify(b)
-      : false
-  )
-  /* eslint-enable eqeqeq */
+  const isObjectA = isObject(a)
+  const isObjectB = isObject(b)
+  if (isObjectA && isObjectB) {
+    return JSON.stringify(a) === JSON.stringify(b)
+  } else if (!isObjectA && !isObjectB) {
```
kazupon: LGTM.
(It may be better to check more strictly ... 😉 )

dsonet: It's loose equal. Which I think just add the 0 check is enough. So we can skip unnecessary methods invoke here and keep it simple.
```
if(a === 0) {
  a = '0'
}
if(b === 0) {
  b = '0'
}
return a == b || (
     isObject(a) && isObject(b)
      ? JSON.stringify(a) === JSON.stringify(b)
      : false
 )
```

defcc: We need strict equality for primitive type, e.g., '0' is equal to false.  I'll update the test case like bellow
```js
new Vue({
  data: {
    val: false,
    test: 0
  },
  template:
    '<div>' +
      '<input type="radio" value="" v-model="test">' +
      '<input type="radio" value="0" v-model="test">' +
      '<input type="radio" :value="val" v-model="test">' +
    '</div>'
}).$mount('#app');
```

defcc: done via #4542

PR #4539, opened 2016-12-21
Don't cast empty string to Boolean if prop types include String and Boolean (fix #4538)

### LINKED ISSUE #4538: Multi-type prop validation [Boolean, String] cast empty string to `true`
### Vue.js version
2.0.2

### Reproduction Link
https://jsfiddle.net/fliptheweb/zjr41svn/4/

### What is Expected?
An empty string `''` must pass to the component like an empty string.

It's type cast type to boolean `true`

fliptheweb:
Additional info in issue  #4538


yyx990803:
Thanks!

PR #4553, opened 2016-12-23
In some cases this will trigger an error when use and the is null

lmf:


yyx990803:
Thanks for the PR, however, this should only occur when `$broadcast` is called on an already destroyed instance, which should not happen in normal use cases. This indicates there are some other causes leading to this error, and this fix is not fixing the root cause.

PR #4558, opened 2016-12-23
[wip] Error mapping option for server render bundler

blake-newman:
POC to improve mapping sourcemaps for better dev experience.

- Has to be contained to the initial error stack, if using the emitted error the stack trace will no longer map up casing an error.

- Todo: if accepted add documentation to use the feature to improve development experience of error traces

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "rollup-plugin-replace": "^1.1.0",
     "rollup-watch": "^2.5.0",
     "selenium-server": "2.53.1",
+    "source-map-stack": "^0.1.7",
     "typescript": "^2.0.9",
     "uglify-js": "^2.6.2",
     "webpack": "^1.13.2",
--- a/src/server/create-bundle-renderer.js
+++ b/src/server/create-bundle-renderer.js
@@ -4,6 +4,7 @@ import { PassThrough } from 'stream'
 export function createBundleRendererCreator (createRenderer) {
   return (code, rendererOptions) => {
     const renderer = createRenderer(rendererOptions)
+    const errorMapper = rendererOptions && rendererOptions.errorMapper
     return {
       renderToString: (context, cb) => {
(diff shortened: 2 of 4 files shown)


blake-newman:
Note: The reason this doesn't catch and fully handle the error is because you can use `node --inspect` and this produces a reasonable stack trace you can deal with. If you manipulate the error the stack will no longer compute to 'real' files. 

The aim is to provide a usable way to provide the error, for other tools to manipulate. This is a POC so final implementation can be different.

blake-newman:
@yyx990803 This is probably cheaper to keep user side. No need to parse around the map. As this is all this truely does. 

Here is a snippet for current reference for those interested in improving their development experience with SSR.

@atinux Nuxt.js could probably utilize this. 

Unsure why there is differences with webpack source maps might be due to issue with https://github.com/webpack/webpack/issues/3165

`npm install --save-dev source-map-stack` / `yarn add -D source-map-stack`

```js
// webpack.server.config.js
module.exports = {
  // Inline sourcemaps
  // webpack 1.0
  devtool: '#inline-source-map',
  // webpack 2.0
  devtool: '#inline-cheap-source-map',
}
```

```js
// server.js 

// import source-map-stack
const sourceMapStack = require('source-map-stack');

// With generated server bundle get the sourcemap
const map = sourceMapStack.get(bundle);

// With vue-server-renderer's error handler we can correctly map the source
app.get('*', (req, res) => {
  const renderStream = renderer.renderToStream(context);

  renderStream.on('error', error => {
    // Output mapped stack trace
    const mappedError = sourceMapStack.stack(map, error);
    console.error(mappedError);
  });
});
```

blake-newman:
Note: Edited above snippet to include `try/catch` block, as if mapping is wrong ie an internal error occurs then the stack trace can still be modified.

PR #4572, opened 2016-12-26
Enforce to loop through children to get the correct normalize type (fix #4564)

### LINKED ISSUE #4564: <template> tag with v-for render undefined
if i use a Array as a prop of custom component,
the template element with v-for directive just render undefined

```
<component :list="list"></component>
<template v-for="item of list">{{item.name}}</template> <!-- output undefined -->
```

it's strange.
only template element tag has this problem.
 

### Vue.js version
2.1.7

### Reproduction Link
simple demo here
https://github.com/shjyh/vue_error_demo/tree/master/2016-12-26

defcc:
fix #4564


defcc:
More test cases needed for normalization optimization.

PR #4613, opened 2016-12-30
Small refactor for template parser

defcc:
1. Remove the unnecessary argument of parseEndTag.
2. Store the lowerCasedTag in the stack and use it directly.
3. Remove isSpecialTag logic for simplification.

As for **isSpecialTag** logic, the original implementation checks the tagName and its attribute. If it's a **template** with **lang attribute**, it will skip to parse the content inside the template. I think maybe we could simplify it, because the template element is guaranteed by the SFC parser.

```js
const hasLang = attr => attr.name === 'lang' && attr.value !== 'html'
```
For this line, is it an edge case?  There is no proper test case about it, I am not very sure.

Please let me know if anything I missed.

### DIFF
--- a/src/compiler/parser/html-parser.js
+++ b/src/compiler/parser/html-parser.js
@@ -52,15 +52,11 @@ const isSpecialTag = (tag, isSFC, stack) => {
   if (isScriptOrStyle(tag)) {
     return true
   }
-  if (isSFC && stack.length === 1) {
-    // top-level template that has no pre-processor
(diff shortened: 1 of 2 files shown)


yyx990803:
Ah cool. For the template tag, the special check was introduced [here](https://github.com/vuejs/vue/commit/7dc38dea27ba7b99d8cb56e1cbb9be0659e217f7), but it is no longer necessary because we have improved the HTML parser to forgivingly handle `<` symbols inside templates now.

Issue #4521, opened 2016-12-20
checkbox can not be selected if it's in a element with @click listener?


qbaty:
### Vue.js version
2.1.6 

### Reproduction Link

http://codepen.io/qbaty/pen/NboZoz?editors=1111

### Steps to reproduce
my html is div > input[type=checkbox]
when div with v-on:click, I found that the input element can not be selected.

### What is Expected?
when click this input checkbox, checkbox should be selected or canceled.
div click event will trigger.

### What is actually happening?
I have not read the source code，But I think div capture this click event and should not stop this event immediate, it kinda like use “event.stopImmediatePropagation”  stop the event progagation?

Marshane:
<input type="checkbox" v-model="checked">
v-model ?!

qbaty:
@Marshare I do not think you understood my problem。

Yunkou:
@qbaty  `@click.self="add" `

LinusBorg:
@qbaty Thanks for rteporting this, looks like a bug indeed!

I took the freedom to correct the Vue version in your OP, as you use 2.1.6, noch 2.0.3

qbaty:
@LinusBorg Okey first I met this bug when I am using vue 2.0.3,and I try to update vue to the lastest version. this bug still exsit, so ... version should be 2.x

FKobus:
Seems fixing this bug created a new bug for :checked="" binding when using in combination with vuex states for maintaining checkbox states.

```javascript
<template>
	<ul>
		<li v-for="option in visibleOptions" :key="option.value">
			<input
				type="checkbox"
				:name="filter.key"
				:value="option.value"
				:checked="inArray(option.value, filter.active)"
				@change="updateFilter">
		</li>
	</ul>
</template>
```
In this code example is :checked set to true or false whenever it's in the active filter array.
It seems when this component is updated (via vuex) the :checked stated doesn't get's updated like in 2.1.6. In my application is the state managed _almost_ totally by Vuex. 

### Vue.js version
2.1.7+

### Reproduction Link
http://vroom.automotivated.nl/

### Howto reproduce
When selecting several checkboxes (like brands) and use your **browser back button** to fallback to previous states. (it's managed by the popstate event)

### What is expected
The checkbox will go on or off depending on the previous action.

### What is actually happening?
[...]

PR #4639, opened 2017-01-03
Use 'click' event for checkbox and radio (fix #4620)

### LINKED ISSUE #4620: Checkbox :checked is not fully reactive when checkbox is in focus
Example: https://jsfiddle.net/vfmxxzeb/3/

What the example does:
1. Resets `values` to an array of values every one second
2. Each line has two checkboxes, both of which are bound to the same value

Steps to reproduce:
1. *Click* on the enabled checkbox on any line

Expected behaviour:
1. After one second, the value should be reset. Both checkboxes should reflect the same value

Actual behaviour (tested on Chrome):
1. The disabled checkbox correctly reflects the value, but the enabled checkbox does not reflect the value. The enabled checkbox will reflect the correct value only after blurring from the checkbox

defcc:
fix #4620 

1. use *click* event instead of *change* event.
2. add test case for #4521

### DIFF
--- a/src/platforms/web/compiler/directives/model.js
+++ b/src/platforms/web/compiler/directives/model.js
@@ -59,7 +59,7 @@ function genCheckboxModel (
       `?_i(${value},${valueBinding})>-1` +
       `:_q(${value},${trueValueBinding})`
   )
-  addHandler(el, 'change',
+  addHandler(el, 'click',
     `var $$a=${value},` +
         '$$el=$event.target,' +
         `$$c=$$el.checked?(${trueValueBinding}):(${falseValueBinding});` +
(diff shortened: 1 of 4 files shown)


sirlancelot:
Will this event still trigger if someone isn't clicking it? For instance, through keyboard tabbing & spacebar?

defcc:
Keyboard's spacebar will change the checkbox's checked status and trigger click event on Mac Chrome. If the browser supports the spacebar, it should also trigger the click event. 

https://www.w3.org/TR/WCAG20-TECHS/SCR35

yyx990803:
Hmm, the CI failed after merging (but it was passing for the PR itself...) Need to investigate a bit more.

defcc:
I'll take a look.

defcc:
Something wrong with the checkbox test case. The checkbox and radio need the same test case, so I made a mistake when doing  it.  I'll check why it works in the PR.

defcc:
I'll also update the contribution guide later.

Issue #4655, opened 2017-01-05
.once modifier for v-on except for the last one doesn't work if more than one .once modifier used for v-on in multiple elements


KingMario:
### Vue.js version
2.1.8

### Reproduction Link
https://jsfiddle.net/bacft9s7/

### Steps to reproduce
Clicks on the <A> element more than once.

### What is Expected?
The event handler should be removed after the first click.

### What is actually happening?
The event handler except for the last one with .once modifier is not removed.

Fixed in pr #4657

PR #4657, opened 2017-01-05
Multiple elements event handler with .once modifier no removal bug fixed

KingMario:
Fix issue #4655


KingMario:
@defcc OK. Thanks

yyx990803:
Thanks, it is intentional to avoid creating new `add` and `remove` functions on every update call for performance reasons.

Fixed in ee6ad6a which preserves the current optimization.

PR #4674, opened 2017-01-07
Use ES2015 destructuring assignment to swap variables

phanan:
This commit uses ES2015 destructuring assignment to swap two variables instead of using a `tmp` var. Flow has an open issue (facebook/flow#183) with this feature, though, so `// $FlowFixMe` is used to temporarily suppress the error.

### DIFF
--- a/src/core/observer/watcher.js
+++ b/src/core/observer/watcher.js
@@ -125,13 +125,11 @@ export default class Watcher {
         dep.removeSub(this)
       }
     }
-    let tmp = this.depIds
-    this.depIds = this.newDepIds
-    this.newDepIds = tmp
-    this.newDepIds.clear()
-    tmp = this.deps
-    this.deps = this.newDeps
-    this.newDeps = tmp
+    // $FlowFixMe
+    [this.depIds, this.newDepIds] = [this.newDepIds, this.depIds]
+    this.newDepIds.clear();
+    // $FlowFixMe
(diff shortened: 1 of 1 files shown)


src/core/observer/watcher.js
```diff
@@ -125,13 +125,11 @@ export default class Watcher {
...
     }
-    let tmp = this.depIds
-    this.depIds = this.newDepIds
-    this.newDepIds = tmp
-    this.newDepIds.clear()
-    tmp = this.deps
-    this.deps = this.newDeps
-    this.newDeps = tmp
+    // $FlowFixMe
+    [this.depIds, this.newDepIds] = [this.newDepIds, this.depIds]
```
HerringtonDarkholme: Maybe leading semicolon is better for `[]`? Since the change is packed in one line so it is harder for one to forget adding semicolon.

phanan: If they forget, the lint/test process will certainly remind them, right?

yyx990803:
Checked the output in Buble and it uses an intermediate Array for the purpose: https://buble.surge.sh/#%5Bthis.depIds%2C%20this.newDepIds%5D%20%3D%20%5Bthis.newDepIds%2C%20this.depIds%5D - which leads to more allocation than necessary - and this part is pretty perf-sensitive.

P.S. personally I'm not really a fan of this syntax's readability :/

phanan:
Oh damn! Well, the more you know...

PR #4684, opened 2017-01-10
fine-tuning of toNumber function

mingmingwon:
1. parseFloat() support only one param, see MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat
2. maybe it's better to use isNaN() to judge the conversion result?

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -16,8 +16,8 @@ export function _toString (val: any): string {
  * If the conversion fails, return original string.
  */
 export function toNumber (val: string): number | string {
-  const n = parseFloat(val, 10)
-  return (n || n === 0) ? n : val
+  const n = parseFloat(val)
+  return isNaN(n) ? val : n
 }
 
 /**


src/shared/util.js
```diff
@@ -16,8 +16,8 @@ export function _toString (val: any): string {
  * If the conversion fails, return original string.
  */
 export function toNumber (val: string): number | string {
-  const n = parseFloat(val, 10)
-  return (n || n === 0) ? n : val
+  const n = parseFloat(val)
+  return isNaN(n) ? val : n
```
znck: With regex

``` js
if(/^(\-|\+)?([0-9]+(\.[0-9]+)?|Infinity)$/
      .test(val))
      return Number(val);
return val;
```

mingmingwon: regex is also good…

posva: I don't think using a regex is a good idea. parseFloat seems to be the fastest way to convert a string to a number: https://jsperf.com/number-vs-plus-vs-parsefloat

yyx990803:
Thanks!

src/shared/util.js
heroBack: what

PR #4714, opened 2017-01-15
Use shields.io compliant badge for SauceLabs

FezVrasta:
Just a small change to make the readme nicer.

image

(I'm pushing to master since this change affects only the readme file)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@
   <a href="https://codecov.io/github/vuejs/vue?branch=dev"><img src="https://img.shields.io/codecov/c/github/vuejs/vue/dev.svg" alt="Coverage Status"></a>
   <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/dt/vue.svg" alt="Downloads"></a>
   <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/v/vue.svg" alt="Version"></a>
-  <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/l/vue.svg" alt="License"></a>
+  <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/l/vue.svg" alt="License"></a> 
+  <br />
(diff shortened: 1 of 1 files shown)


FezVrasta:
The badge I used supports multiple browsers and versions.

image

yyx990803:
Ah cool. It's showing Android only because I recently ran a single-browser test.

README.md
```diff
@@ -6,8 +6,7 @@
   <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/dt/vue.svg" alt="Downloads"></a>
   <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/v/vue.svg" alt="Version"></a>
   <a href="https://www.npmjs.com/package/vue"><img src="https://img.shields.io/npm/l/vue.svg" alt="License"></a>
-  <br>
-  <a href="https://saucelabs.com/u/vuejs"><img src="https://saucelabs.com/browser-matrix/vuejs.svg" alt="Sauce Test Status"></a>
+  <a href="https://saucelabs.com/u/vuejs"><img src="https://badges.herokuapp.com/sauce/vuejs" alt="Sauce Test Status"></a>
```
yyx990803: Let's preserve the `<br>` here so that browser matrix is on a separate line.

FezVrasta:
done

posva:
This is very subjective, but wouldn't this be too many badges?
I really think the matrix version is easier to read. The line break wouldn't suffice imo to make the badge equivalent readable enough

FezVrasta:
I personally don't like the matrix because it's very big and takes a lot of vertical space. The badges are nice and people are used to them.

posva:
I see. So many badges on a single line are harder to read, even if we're used to reading badges. I personally prioritise readability over vertical space on a readme

FezVrasta:
Still personal preference, but to me it's easier to read: "Browser: version, version, version" instead of  the matrix style. Especially because in the matrix style you have also the OS version that confuses a lot, for example myself get confused often by the windows version and the IE version 🙄

yyx990803:
On a second thought, I still find the matrix to provide a better at-a-glance view of browser compatibility than badges. Thanks for the PR though!

PR #4724, opened 2017-01-16
Fix hot reload in IE in strict mode

### LINKED ISSUE #4465: "Invalid calling object" error in IE
### Vue.js version
1.0.28

### Reproduction Link

I've identified what is needed to fix the bug and I think, by inspection, it is the right thing to do.  Producing a fiddle to reproduce this is somewhat complex and seems out of proportion to the fix, however if I absolutely have to then I will.

### Steps to reproduce
Where there is an app with dynamic elements with transitions defined with callback hooks on the transitions, open it in IE (tested in 11 and 9,10 through emulation.  No problem in other browsers)

### What is Expected?
The transition completes with the hooks called at the right times.

The transition does not complete and an error is thrown in Vue: "Invalid calling object" on this code (which starts on line 37 src/transition.js, the error is thrown on line 40):

```
const raf = inBrowser && window.requestAnimationFrame
const waitForTransitionStart = raf
  /* istanbul ignore next */
  ? function (fn) { raf(function () { raf(fn) }) } //<-- error thrown here!
  : function (fn) { setTimeout(fn, 50) }
```

The error seems to be because the `requestAnimationFrame` method of the `window` object is being called without its parent object.  

### What seems to fix it?
By changing line 37 to
```
const raf = inBrowser && window.requestAnimationFrame.bind(window)
```
the error goes away.  Using bind seems, to me, obviously correct.  Clealy in other browsers because the default `this` is the `window` object anyway it should not matter, however IE seems to take a dislike to this approach and throw the error.  Binding window to the method would seem harmless otherwise.

I'm an old-hand at coding but new to github (and relatively new to JS and vue) - The fix works but I don't really know if there are other reasons why this is not a good idea so I'm cautious about just closing it.

posva:
Fix #4465

### DIFF
--- a/src/transition/transition.js
+++ b/src/transition/transition.js
@@ -32,9 +32,12 @@ const animDurationProp = animationProp + 'Duration'
  * transitions but not raf (e.g. Android 4.2 browser) - since
  * these environments are usually slow, we are giving it a
  * relatively large timeout.
+ * 
+ * Binding to window is necessary to make hot reload work in
+ * IE in strict mode
  */
 
-const raf = inBrowser && window.requestAnimationFrame
+const raf = inBrowser && window.requestAnimationFrame.bind(window)
 const waitForTransitionStart = raf
   /* istanbul ignore next */
   ? function (fn) { raf(function () { raf(fn) }) }


yyx990803:
2.x can use this fix as well in `transition-util.js`

posva:
@yyx990803 Ok, I'll create the PR too

Issue #4187, opened 2016-11-13
Recusive Local Components - Uncaught RangeError: Maximum call stack size exceeded


cseufert:
### Vue.js version
1.0.28

### Reproduction Link

https://jsfiddle.net/25fdgLgr/3/

### Steps to reproduce
Just run new Vue() or let app demo app load

### What is Expected?
App loads as expected. (btw this does work in vue 2.0, but we are not quite ready to migrate the rest of our app.

### What is actually happening?
JS Error
vue.js:10148 Uncaught RangeError: Maximum call stack size exceeded
    at new Function (<anonymous>)
    at createClass (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:10148:12)
    at Function.Vue.extend (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:10113:15)
    at guardComponents (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:1927:31)
    at mergeOptions (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:2004:3)
    at Function.Vue.extend (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:10117:19)
    at guardComponents (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:1927:31)
    at mergeOptions (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:2004:3)
    at Function.Vue.extend (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:10117:19)
    at guardComponents (https://cdnjs.cloudflare.com/ajax/libs/vue/1.0.28/vue.js:1927:31)

KingMario:
I'll try to resolve it.

neobats:
Just noticed that this is marked 1.x, should it be closed? If not, has/should the PR be reviewed? Or am I missing something, which is likely.

KingMario:
perhaps v1 is discontinued?

PR #4739, opened 2017-01-18
Bug fix for issue 4187 (recursive components)

KingMario:
Hopefully fixed the issue. Don't know how to write the unit test for this.

### DIFF
--- a/src/global-api.js
+++ b/src/global-api.js
@@ -87,7 +87,7 @@ export default function (Vue) {
    * @param {Object} extendOptions
    */
 
-  Vue.extend = function (extendOptions) {
+  Vue.extend = function (extendOptions, components, key) {
     extendOptions = extendOptions || {}
     var Super = this
     var isFirstExtend = Super.cid === 0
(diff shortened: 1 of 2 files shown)


posva:
Hey, thanks for the PR!
Could you add a test to it, please?

KingMario:
@posva I'll do it later.

posva:
@KingMario No worries 🙂  Thanks!

KingMario:
Will deal with the unit test error tomorrow. Hold on.

KingMario:
Want to discuss with you guys.

If there's a common component used in option like this:
```
    components: {
      outer: {
        template: '<div v-if="$parent.show">' +
          '<slot></slot>' +
          '</div>' +
          // this is to test that compnents that are not in the if block
          // should not fire attach/detach when v-if toggles
          '<transcluded></transcluded>',
        components: {
          transcluded: transcluded // common component
        }
      },
      transcluded: transcluded // common component
    }
```

I suggest that the reused component `transcluded`, should be extended by `Vue.extend()`. The syntax sugar is not allowed.
And if recursive components are used, they should be declared just as same as in issue #4187, i.e., allComps should be declared and used in the components option of each component.

The saying cases are not very common in usage, so the restricts are acceptable to most users.

cseufert:
Is there any reason this has not been merged?

KingMario:
@cseufert From the official site, the version plan for Vue 1 is in a half-year interval. So I guess this will not be merged till the day Vue 1.2 is released.

You may just modify your file as in this pull request in node_modules if it's an urgency.

PR #4745, opened 2017-01-18
[weex] Fix the memory crosstalk between multiple instance

Hanks10100:
In Weex, all instance share a same runtime, the closure of `Dep`, `Scheduler` and so on, are also shared by multiple instance.

If the previous instance was crashed by accident, some shared variables might still remain the old value when creating the new instance.

## Modification

In particular, the following variables have the protential risk of memory crosstalk:

+ [`core/observer/dep.js`](https://github.com/vuejs/vue/blob/994aee4f869d68a00c3b9d87daa9c8d84623ac42/src/core/observer/dep.js#L48-L49)
  + `Dep.target`
  + `targetStack`
+ [`core/observer/scheduler.js`](https://github.com/vuejs/vue/blob/994aee4f869d68a00c3b9d87daa9c8d84623ac42/src/core/observer/scheduler.js#L12-L16)
  + `queue`
  + `has`
  + `circular`
  + `waiting`
  + `flushing`

The modification is simple: Just reset those variables to its initial value before `createInstance` and `destroyInstance`. (*Actually, reset the value in `createInstance` is enough, but do likewise in `destroyInstance` is more reliable*)

## Example

Here is an example:

```html
<template>
  <div class="list">
    <div class="group">
      <div class="panel" v-for="obj in lists">
        <text class="text">{{obj.name}}</text>
      </div>
    </div>
  </div>
</template>

<script>
  export default {
    data () {
      return {
        lists: [
          { name: '1' },
          { name: '2' }
        ],
      }
    },
    mounted () {
      Vue.set(this.lists, 0, { name: 'A'})

      setTimeout(() => {
        Vue.delete(this.lists, 1)
        // this.lists.length--
      }, 1000)
    }
  }
</script>

<style scoped>
  .panel {
    width: 250px;
    height: 250px;
    flex-direction: column;
    justify-content: center;
    border-width: 2px;
    border-style: solid;
    border-color: rgb(162, 217, 192);
    background-color: rgba(162, 217, 192, 0.2);
  }
  .group {
    flex-direction: row;
    justify-content:space-between;
    width: 600px;
    margin-left: 75px;
    margin-top: 100px;
  }
  .text {
    font-size: 128px;
    text-align: center;
    color: #41B883;
  }
</style>
```

> Compile the code into js bundle by `weex-loader`, and run the code in Android or iOS device with WeexSDK would recurrence this bug.

The page will show `A` and `2` initially. After one second, the `Vue.delete` statement will occur an error, the call stack would be canceled, but right now, the `flushing` variable in scheduler is `true`.

Refresh the page (on native), will get `1` and `2`, which is not correct. Because the initial value of `flushing` is `true` in scheduler, the watcher of `lists` will never run.

### DIFF
--- a/src/core/observer/dep.js
+++ b/src/core/observer/dep.js
@@ -46,7 +46,7 @@ export default class Dep {
 // this is globally unique because there could be only one
 // watcher being evaluated at any time.
 Dep.target = null
-const targetStack = []
+let targetStack = []
 
 export function pushTarget (_target: Watcher) {
   if (Dep.target) targetStack.push(Dep.target)
@@ -55,4 +55,9 @@ export function pushTarget (_target: Watcher) {
 
 export function popTarget () {
   Dep.target = targetStack.pop()
+}
+
+export function resetTarget () {
+  Dep.target = null
+  targetStack = []
 }
--- a/src/core/observer/dep.js
+++ b/src/core/observer/dep.js
@@ -46,7 +46,7 @@ export default class Dep {
 // this is globally unique because there could be only one
 // watcher being evaluated at any time.
 Dep.target = null
-let targetStack = []
+const targetStack = []
 
 export function pushTarget (_target: Watcher) {
   if (Dep.target) targetStack.push(Dep.target)
@@ -59,5 +59,5 @@ export function popTarget () {
 
 export function resetTarget () {
   Dep.target = null
-  targetStack = []
+  targetStack.length = 0
 }
--- a/src/core/observer/scheduler.js
+++ b/src/core/observer/scheduler.js
@@ -19,7 +19,7 @@ let index = 0
 /**
  * Reset the scheduler's state.
  */
-function resetSchedulerState () {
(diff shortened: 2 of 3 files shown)


yyx990803:
Is it possible for `createInstance` or `destroyInstance` to be called when the scheduler is pending? That may create some weird edge cases.

I think the cleaner way to handle this is probably expose Vue via a factory function and basically create a new instance of Vue for each Weex instance. (which would also solve the `Vue.config` usage problem)

Hanks10100:
The `createInstance` and `destroyInstance` is called by native, which depends on the user's behavior. Even in the same *activity* would contain multi Weex instance.

I also think expose Vue through factory function is a more elegant way to fix this. But doing that may change too much (still worthy), and should also take care of those variables in closure. This is a much cheaper way to fix this without side effects.

yyx990803:
closed via 53d059222f568b14334f1aa9d161da0c79fef58e

PR #4747, opened 2017-01-19
Be able to delete array item in `Vue.delete`

Hanks10100:
Since the `Vue.set` support to insert an item to array, the `Vue.delete` should also support to delete an item from array.

Take the following code as an example:

```js
const vm = new Vue({
  template: '<div><span v-for="obj in lists">{{obj.name}}</span></div>',
  data: {
    lists: [
      { name: 'A' },
      { name: 'B' }
    ]
  }
}).$mount()

Vue.delete(vm.lists, 1)
// vm.lists.length--
```

The `Vue.delete(vm.lists, 1)` just delete the property `1` from the array, but the `vm.lists.length` is still be 2. Evaluating `obj.name` at the second time will caught a TypeError (undefined is not an object).

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -218,7 +218,11 @@ export function set (obj: Array<any> | Object, key: any, val: any) {
 /**
  * Delete a property and trigger change if necessary.
  */
-export function del (obj: Object, key: string) {
+export function del (obj: Array<any> | Object, key: any) {
+  if (Array.isArray(obj)) {
+    obj.splice(key, 1)
+    return
+  }
   const ob = obj.__ob__
   if (obj._isVue || (ob && ob.vmCount)) {
     process.env.NODE_ENV !== 'production' && warn(
--- a/test/unit/features/global-api/set-delete.spec.js
+++ b/test/unit/features/global-api/set-delete.spec.js
@@ -72,6 +72,7 @@ describe('Global API: set/delete', () => {
         expect(vm.$el.innerHTML).toBe('')
       }).then(done)
     })
+
     it('be able to delete an item in array', done => {
       const vm = new Vue({
         template: '<div><p v-for="obj in lists">{{obj.name}}</p></div>',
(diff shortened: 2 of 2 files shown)


src/core/observer/index.js
```diff
@@ -218,7 +218,11 @@ export function set (obj: Array<any> | Object, key: any, val: any) {
 /**
  * Delete a property and trigger change if necessary.
  */
-export function del (obj: Object, key: string) {
+export function del (obj: Array<any> | Object, key: any) {
```
posva: key is either a string or a number, isn't it?

Hanks10100: Yes, I just keep the type check be same with `Vue.set`

posva: I see, in that case, we should probably update both

test/unit/features/global-api/set-delete.spec.js
```diff
@@ -72,5 +72,29 @@ describe('Global API: set/delete', () => {
         expect(vm.$el.innerHTML).toBe('')
       }).then(done)
     })
+    it('be able to delete an item in array', done => {
```
yyx990803: blank line here please :)

src/core/observer/index.js
Hanks10100: The conflict is the first argument of `splice` must be number, can't be string.

posva: I must be missing something 😅  what is wrong with `key: number | string`

test/unit/features/global-api/set-delete.spec.js
Hanks10100: OK, I will also remove the blank line after `.$mount()`

src/core/observer/index.js
Hanks10100: I have tried, if I set the `key: number | string`, the type check of `obj.splice(key, 1)` would be failed. Because the first argument must be a number.

PR #4753, opened 2017-01-19
fix(sfc): component contains '<' only

dhcmrlchtdj:
```vue
<template>
    <span><</span>
</template>
```

Vue will run into infinite loop with such component.
It's better to give some warning?


dhcmrlchtdj:
@yyx990803 thanks for your review.

PR #4797, opened 2017-01-27
Avoid double-decoding on attribute value

jddxf:
Chained `replace`s here may cause problems if the order isn't carefully arranged.Currently we get `"` after decoding `&amp;quot;` while `&quot;` is expected.

For a quick fix, I could move the `replace` which replaces `&amp;`s to the last.However, I prefer one single `replace` so that we don't bother with chaining order.

### DIFF
--- a/src/compiler/parser/html-parser.js
+++ b/src/compiler/parser/html-parser.js
@@ -49,21 +49,17 @@ let IS_REGEX_CAPTURING_BROKEN = false
 const isScriptOrStyle = makeMap('script,style', true)
 const reCache = {}
 
-const ltRE = /&lt;/g
-const gtRE = /&gt;/g
-const nlRE = /&#10;/g
-const ampRE = /&amp;/g
-const quoteRE = /&quot;/g
+const decodingMap = {
+  '&lt;': '<',
+  '&gt;': '>',
+  '&quot;': '"',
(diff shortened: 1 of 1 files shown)


src/compiler/parser/html-parser.js
```diff
@@ -49,21 +49,17 @@ let IS_REGEX_CAPTURING_BROKEN = false
...
 function decodeAttr (value, shouldDecodeNewlines) {
-  if (shouldDecodeNewlines) {
-    value = value.replace(nlRE, '\n')
-  }
-  return value
-    .replace(ltRE, '<')
-    .replace(gtRE, '>')
-    .replace(ampRE, '&')
-    .replace(quoteRE, '"')
+  const re = shouldDecodeNewlines ? /&(lt|gt|quot|amp|#10);/g : /&(lt|gt|quot|amp);/g
```
yyx990803: Can we hoist the two RegExps out so that they don't get instantiated on every call?

yyx990803:
Good catch!

src/compiler/parser/html-parser.js
jddxf: That makes sense!

jddxf:
It took me one hour to come up with the variable names:sob:

PR #4807, opened 2017-01-29
[WIP] Support for ref callback

znck:
Changes proposed in this pull request:
- Add ref callback

/ping @yyx990803


yyx990803:
Thanks a lot for starting the work on this. I felt it was really awkward that the user needs to manually remove a ref in the callback, so I adjusted the strategy to simply reset the $refs object on re-renders.

Also turns out the test was failing because on CI the tests are run in PhantomJS which doesn't support arrow functions...

Issue #4571, opened 2016-12-26
$props


Akryum:
A `$props` special attribute would be very handy to reuse generic components inside more specific ones. It would hold the props passed down to the component.

For example, I'd like to create a `InputModal` which would be a `Modal` with some predefined content (for example,one `<input/>`).

```html
<template>
  <modal v-bind="$props">
    <input v-model="inputValue" :placeholder="placeholder" />
  </modal>
</template>

<script>
import Modal from './Modal.vue'

export default {
  props: {
    ...Modal.props,
    value: {},
    placeholder: String,
  },

  // ...
}
</script>
```

Here is a [jsfiddle](https://jsfiddle.net/Akryum/z66hve7n/) of the use case.

PR #4848, opened 2017-02-04
add $props

yantene:
I've implemented a `$props` special attribute on components to access its props because I thought the feature proposed in #4571 was great.
For this implementation, I rewritten `proxy` function in src/core/instance/state.js to general purpose one.
Thanks for the wonderful suggestion, @akryum!

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -39,6 +39,7 @@ const isReservedProp = { key: 1, ref: 1, slot: 1 }
 
 function initProps (vm: Component, props: Object) {
   const propsData = vm.$options.propsData || {}
+  vm.$props = {}
   // cache prop keys so that future props updates can iterate using Array
   // instead of dyanmic object key enumeration.
   const keys = vm.$options._propKeys = []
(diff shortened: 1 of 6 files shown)


gebilaoxiong:
Seems to be a destructive modification

src/core/instance/state.js
```diff
@@ -67,8 +68,9 @@ function initProps (vm: Component, props: Object) {
         }
       })
     } else {
-      defineReactive(vm, key, validateProp(key, props, propsData, vm))
+      defineReactive(vm.$props, key, validateProp(key, props, propsData, vm))
     }
+    proxy(vm, '$props', key)
```
gebilaoxiong: There is a problem here.
I think that is... if you want to do this.
You need to create a proxy on vm, not on `$props`
Or may bring destructive changes

yantene:
@blake-newman Thank you for reviewing!
I added the type of props in typescript definitions.
But I'm not very confident about this commit, please check it.

types/vue.d.ts
```diff
@@ -42,6 +42,7 @@ export declare class Vue {
   readonly $slots: { [key: string]: VNode[] };
   readonly $scopedSlots: { [key: string]: ScopedSlot };
   readonly $isServer: boolean;
+  $props: Object;
```
ktsn: In TypeScript, it would be better to use `any` rather than `Object` because the latter does not allow property access as it indicates super class of all object type (not *any* object type).
e.g. We cannot write `this.$props.foo` even if we define `foo` props.

In addition, can we add `readonly` modifier since `$props` is readonly property on userland? 🙂

yyx990803:
Thank you for the PR!

Issue #4856, opened 2017-02-05
$on with an array of events


markbrown4:
Would be neat if `$on` supported an array of events:

```js
bus.$on(['CLICK_BODY', 'KEYUP_ESCAPE'], () => {
  this.active = false
})
```

Without it we need duplication

```js
bus.$on('CLICK_BODY', () => {
  this.active = false
})
bus.$on('KEYUP_ESCAPE', () => {
  this.active = false
})

// or

bus.$on('CLICK_BODY', deactivate)
bus.$on('KEYUP_ESCAPE', deactivate)
```

yyx990803:
Closed via #4860

PR #4860, opened 2017-02-06
support $on multi event - #4856

Kingwl:
#4856

### DIFF
--- a/src/core/instance/events.js
+++ b/src/core/instance/events.js
@@ -38,13 +38,19 @@ export function updateComponentListeners (
 
 export function eventsMixin (Vue: Class<Component>) {
   const hookRE = /^hook:/
-  Vue.prototype.$on = function (event: string, fn: Function): Component {
+  Vue.prototype.$on = function (event: string | Array<string>, fn: Function): Component {
     const vm: Component = this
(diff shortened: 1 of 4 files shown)


Kingwl:
@kazupon oh sorry...

src/core/instance/events.js
```diff
@@ -38,13 +38,18 @@ export function updateComponentListeners (
...
   const hookRE = /^hook:/
-  Vue.prototype.$on = function (event: string, fn: Function): Component {
+  Vue.prototype.$on = function (event: string | Array<string>, fn: Function): Component {
     const vm: Component = this
-    ;(vm._events[event] || (vm._events[event] = [])).push(fn)
-    // optimize hook:event cost by using a boolean flag marked at registration
-    // instead of a hash lookup
-    if (hookRE.test(event)) {
-      vm._hasHookEvent = true
+    const events: Array<string> = Array.isArray(event) ? event : [event]
```
yyx990803: I'd prefer using for loop only when it's an Array instead of always turning it into an Array.

In the loop you can just do `this.$on(events[i], fn)`

Kingwl: done

yyx990803:
Thanks!

PR #4863, opened 2017-02-06
fix `addClass` utlil bug

hashplus:
`<transition-group move-class=“ value with leading or trailing whitespace ” />`

reproduced:
http://output.jsbin.com/nijuhow/1


defcc:
Thanks @hashplus, will you please add an appropriate test case?

hashplus:
hi @defcc , test case added

Issue #3050, opened 2016-06-10
[Feature] Mouse event listener modifiers


edjroot:
First of all, I'd like to thank everyone who has contributed to this absolutely wonderful library.

I was reading the docs on [key modifiers](https://vuejs.org/guide/events.html#Key-Modifiers) (for example, `@keyup.enter`) and was wondering if it would be worth having the same kind of thing for mouse buttons. This would make handling much simpler since the same events that are available for the left button by default would be for the other ones as well without the need for repetitive checking in the JS code.

For instance:

```
<button @click.left="hello" @click.right="bye">Say hello or bye</button>
<div @mousedown.middle="drag">I'm draggable!</div>
```

The only possible "downside", I guess, is that `@click.right` would become a synonym for `@contextmenu`, but it doesn't sound that bad.

Is this viable?

Thanks in advance!

zoomclub:
Yes, there is some good potential with the @event interface. Like the suggestions made and also had a question/suggestion along the same lines for an @stream interface here: https://github.com/vuejs/vue-rx/issues/5

Kingwl:
ping @posva

PR #4866, opened 2017-02-06
support mouse modifier - #3050

Kingwl:
#3050


posva:
I don't think using the `buttons` api is a good idea because it always returns undefined in Safari.

Kingwl:
@posva yes u r right

PR #4874, opened 2017-02-07
fix select bound with error index - fix #3821

### LINKED ISSUE #3821: <select> bound with array performs incorrectly after splice()
Consider the following code:

```
<div id="demo">
  <select v-model="selected" number>
    <option v-for="opt in options" :value="$index" number>{{opt}}</option>
  </select>
</div>
```

In `<select>`, each option value is bound with `$index`. If we perform `splice()` on `options`, like, `options.splice(0,1)`, `$index` does not correctly sync with options.

For example, let `options=['a','b']` and the rendered HTML looks like (but not really):

```
<div id="demo">
  <select selectedIndex="0">
    <option value="0">a</option>
    <option value="1">b</option>
  </select>
</div>
```

After `options.splice(0,1)`, then `options=['b']`, but the rendered HTML becomes:

```
<div id="demo">
  <select selectedIndex="0">
    <option value="1">b</option>
  </select>
</div>
```

The value of 'b' option does not become 1, which should be 0 because of `$index`. This may not be a bug, but this is kind of odd.

Below is the live demo of this issue:
https://jsfiddle.net/peteranny/trwp98g9/4/

Kingwl:
#3821

### DIFF
--- a/src/directives/public/for.js
+++ b/src/directives/public/for.js
@@ -16,7 +16,8 @@ import {
   def,
   cancellable,
   isArray,
-  isPlainObject
+  isPlainObject,
+  nextTick
 } from '../../util/index'
 
 let uid = 0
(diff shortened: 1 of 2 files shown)


linghengqian:
- Now that https://github.com/vuejs/vue/issues/3821 is closed, this PR should also be closed.

Issue #3870, opened 2016-10-07
A child components `destroyed` hook is called before its `detached` hook


sirlancelot:
### Vue.js version
1.0.28

### Reproduction Link
https://jsfiddle.net/w0r85tus/

### Steps to reproduce
- Open the JSFiddle, then open devtools
- Click "show modal" in the result view
- Click "Ok" in the modal
- Observe the console.log statement order

### What is Expected?
- Components should be detached before they are destroyed.
- The destroyed hook should always be the very last hook fired.

### What is actually happening?
- The destroyed hook is getting called first, then the detached hook later.

Note that it makes no difference if the component has a `transition` applied. You can remove the `transition` attribute from the modal but the hook order stays the same.

PR #4888, opened 2017-02-09
fix the wrong orders of call 'detached' and 'destroyed' hook - #3870

Kingwl:
#3870


Kingwl:
btw
my eslint does not seem to work

```
/vue/src/compiler/compile.js
  8:1  error  '../util/index' import is duplicated  no-duplicate-imports

/vue/src/global-api.js
  19:1  error  './util/index' import is duplicated  no-duplicate-imports

/vue/test/unit/specs/parsers/expression_spec.js
  181:17  error  Unnecessary escape character: \'  no-useless-escape
  181:20  error  Unnecessary escape character: \'  no-useless-escape
```

but ' never mod those file

Kingwl:
@kazupon thanks XD

MrMitch:
Hi,

Was this ever merged ?

PR #4890, opened 2017-02-09
Standardize the JSFiddle starter link

phanan:
We have two different JSFiddle starter links here, none of which is the version on Gitter.im (which appears to be more aligned with Vue's coding standard i.e. no root `<body>` tag, spacings and such). Modifying this issue template to use the Gitter.im version instead.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -26,7 +26,7 @@ Hi! I’m really excited that you are interested in contributing to Vue.js. Befo
 
 - Issues with no clear repro steps will not be triaged. If an issue labeled "need repro" receives no further input from the issue author for more than 5 days, it will be closed.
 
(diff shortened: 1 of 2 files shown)


posva:
Personally, I like to put the vue.js in a script tag so we can easily change versions to check a regression and verify the user is using the latest version. Something like this: http://jsfiddle.net/posva/wsr78d2w/

phanan:
I'm not against it. The whole idea is, we need to stick to only one template, and therefore, one standard.

posva:
@phanan yes, sorry, being too direct 😆 
Yes, I think a single link would be better too and that should agree on a standard.
I think that having the script at the top is the most flexible and explicit approach:

```html
latest
<script src="https://unpkg.com/vue"></script>
latest v1
<script src="https://unpkg.com/vue@1"></script>
fixed version to test regressions
<script src="https://unpkg.com/vue@2.1.7"></script>
```

kazupon:
>I think that having the script at the top is the most flexible and explicit approach:

I think so too

phanan:
Guys, feel free to push more changes into this as you deem fit, as I (GitHub actually) opened a branch on vue.js repo itself.

PR #4907, opened 2017-02-11
Let userland remove the production tips info

smolinari:
Hi,

From the discussions here:

https://forum.vuejs.org/t/hide-notify-you-are-running-vue-in-development-mode/5661

and here:

https://github.com/vuejs/vue/commit/aa6f7b4cdb889fba9575710c5a100234ccfb7975

and because it also annoys me, I decided to try and set up a flag, which will allow a dev to "consciously" remove the production tips info. 

It is also related to (requires) this PR for the webpack template.

https://github.com/vuejs-templates/webpack/pull/510

The flag is currently set to "false", so the client dev will have to go into the file and set it to true, to remove the production tips info. 

If you let me know how to best document the availability of the flag (if the two PRs get accepted), I'd be also glad to add into about the flag to the docs too. 

Also, this currently only works for the webpack setup. I haven't looked into the other build possibilities. 

Scott

### DIFF
--- a/src/entries/web-runtime.js
+++ b/src/entries/web-runtime.js
@@ -38,7 +38,10 @@ Vue.prototype.$mount = function (
 }
 
 if (process.env.NODE_ENV !== 'production' &&
-    inBrowser && typeof console !== 'undefined') {
+    inBrowser &&
+    typeof console !== 'undefined' &&
+    process.env.ADD_DEV_WARNING
+  ) {
   console[console.info ? 'info' : 'log'](
     `You are running Vue in development mode.\n` +
     `Make sure to turn on production mode when deploying for production.\n` +
--- a/src/entries/web-runtime.js
+++ b/src/entries/web-runtime.js
@@ -40,7 +40,7 @@ Vue.prototype.$mount = function (
 if (process.env.NODE_ENV !== 'production' &&
     inBrowser &&
     typeof console !== 'undefined' &&
-    !process.env.REMOVE_DEV_WARNING
+    !process.env.REMOVE_PROD_TIP
   ) {
   console[console.info ? 'info' : 'log'](
     `You are running Vue in development mode.\n` +
--- a/src/entries/web-runtime.js
+++ b/src/entries/web-runtime.js
@@ -40,7 +40,7 @@ Vue.prototype.$mount = function (
 if (process.env.NODE_ENV !== 'production' &&
     inBrowser &&
     typeof console !== 'undefined' &&
-    process.env.ADD_DEV_WARNING
+    !process.env.REMOVE_DEV_WARNING
   ) {
   console[console.info ? 'info' : 'log'](
     `You are running Vue in development mode.\n` +


yyx990803:
Thanks for the PR, but using `process.env` is actually not ideal, because if the user neglected to provide this env var e.g. via Webpack's `DefinePlugin`, Webpack will attempt to emulate `process` by including a process shim module. The standalone build config also needs to change to make sure this reference to `process` is built-away.

I think it's better to just add a runtime config option. We will likely address this in the next release.

smolinari:
Hi Evan. ( @yyx990803 )

Thanks for addressing the suggestion. 

I has a feeling I was skating on thin ice with this solution, but I didn't want to disrupt what seems to be a "no configuration" mantra of Vue too much and this was the only way I could figure it out. 

On that point, it seems to me, this mantra is what makes the warning necessary in the first place. There are certainly clear decisions that userland should make, one being setting up Vue for development or production and I think it is ok to allow that as a config setting. 

So, I request this not simply be a decision to remove the warning with an extra flag, but rather to have a clear setting for userland to setup Vue to run in "production" or "development" mode. This would negate the necessity of the warning altogether, because the situation "production" or "development" is then a userland responsibility and not a possible mistake of using Vue incorrectly. 

Scott   

P.S. don't forget to close the other PR in the Webpack Template. 😄

jaketoolson:
I loathe the fact Vue is the only package of 50+ I use that actually adds something like this to console.  It's frustrating.

posva:
@jaketoolson This has already been released: `Vue.config.productionTip = false`

smolinari:
Thank you Vue team for taking my suggestion to heart and making it happen! 👍 

Scott

shepelevstas:
Somehow `Vue.config.productionTip = false` is ignored and the message appears anyway.

shepelevstas:
Made it work. Had to put it right after vue.js:
```
...
<script src=".../vue.js">
<script>Vue.config.productionTip=false</script>
<script src=".../vuex.js">
<script src=".../vue-router.js">
...
```

drupal999:
:8080/#/:6 Error parsing a meta element's content: ';' is not a valid key-value pair separator. Please use ',' instead.

MartinManev:
I am setting productionTip and devtools to false, but I'm still occasionally getting the messages printed in the console. Sometimes during testing too. It's really annoying.

LLily7:
It's annoying me too

garyo:
`Vue.config.productionTip=false` is working for me, but could there be a similar tip-removal config flag for the "Download the Vue Devtools extension" message? I don't want to turn off devtools altogether in my unit tests, because sometimes it's useful to have that for debugging. I just don't always want my tests to show this message. It clutters the output.

gubikmic:
To anyone coming here via search: This no **longer works in Vue 3**.
You have to include `https://unpkg.com/vue@next/dist/vue.global.prod.js` to get rid of the warning.

gubikmic:
I realized when using the prod version the Vue DevTools no longer work.

Why was this removed in Vue 3?

I find it really distracting to get this message on every single reload.

smolinari:
AFAIK, and looking at a Vue 3 app, this warning is no longer there. 🤔 

Scott

Issue #4900, opened 2017-02-10
Transition-group transition on empty array


wbijker:
### Vue.js version
2.1.10

### Reproduction Link
https://jsfiddle.net/z11fe07p/686/#

### Steps to reproduce
There seems to be an issue with the transition-group component when transitions is applied on each element iterating on a v-for with an initial empty array.

Whenever you add (push) an item to the array before the previous transition is completed all previous transition will stop.  But if you wait for the first transition to finish - but only on the first instance - and then try to add items all works as expected. Similar, if you fill the initial array with 1 or more values, all works as expected.

### What is Expected?
All items should continue to animate.

### What is actually happening?
Animation stops for all elements except for the last one added.
It seems that Vue detects a change in position and therefore replace all transition classes with the move class.

Kingwl:
so sorry about my three pr
i make some mistake

posva:
hey, no need to apologise 😄 
Thanks for taking the time to make a PR

Kingwl:
@posva 
i cannot recurrent that when i write the unit test .
i try to move transition from .wave-enter-active to .item then it work fine
so this does not look like a bug
https://jsfiddle.net/z11fe07p/706/

posva:
@Kingwl Sorry for the delay. I believe it's a bug because if you wait for the first transition to finish there's no problem with the rest of the animations. It only happens if you add many items on the empty array without waiting for the first transition to finish. I'll check and see if I can help out 🙂

ralphchristianeclipse:
https://jsfiddle.net/RalphChristianEclipse/9L8gvuej/16/

#4913

PR #4911, opened 2017-02-11
Fix transition

posva:
Following work from @kingwl
#4900

### DIFF
--- a/test/unit/features/transition/transition-group.spec.js
+++ b/test/unit/features/transition/transition-group.spec.js
@@ -293,5 +293,44 @@ if (!isIE9) {
       }).$mount()
       expect('<transition-group> children must be keyed: <div>').toHaveBeenWarned()
     })
+
+    // Issue #4900
+    it('stagger animations on initial empty children', done => {
+      const vm = new Vue({
+        template: `
+          <div>
+            <transition-group name="test-empty-array">
+              <div v-for="item in items" :key="item">{{ item }}</div>
+            </transition-group>
+          </div>
+        `,
+        data: {
+          items: []
+        }
+      }).$mount(el)
+      vm.items.push(0)
+      waitForUpdate(() => {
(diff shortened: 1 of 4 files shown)


test/unit/features/transition/transition-group.spec.js
```diff
@@ -293,5 +293,44 @@ if (!isIE9) {
...
+            `<div class="test-empty-array-enter test-empty-array-enter-active">0</div>` +
+          `</span>`
+        )
+      }).thenWaitFor(duration / 3).then(() => {
+        vm.items.push(1)
+      }).thenWaitFor(nextFrame).then(() => {
+        expect(vm.$el.innerHTML).toBe(
+          `<span>` +
+            `<div class="test-empty-array-enter-active test-empty-array-enter-to">0</div>` +
+            `<div class="test-empty-array-enter-active test-empty-array-enter-to">1</div>` +
```
posva: I think the class should actually be `test-empty-array-enter test-empty-array-enter-active` 😕

Kingwl:
sorry for the delay
the day before  yesterday is a Chinese festival: the Lantern Festival so I went to accompany my family
i think something is wrong and i'm try to find that

Kingwl:
@posva 
hello i think i fix that
and i take a pr to your repo
please review that

test/unit/features/transition/transition-group.spec.js
Kingwl: NextFrame looks nextTwoFrame

posva:
@Kingwl ooh 😆 I wasn't expecting that!

> sorry for the delay
the day before yesterday is a Chinese festival: the Lantern Festival.
so I went to accompany my family

No need to apologise!

yyx990803:
The root cause is in the move detection method, I think it's more appropriate to fix it there. But many thanks to both of you @posva and @Kingwl for looking into this!

Issue #4885, opened 2017-02-08
Feature request: transition-group provide hook when all children moves are finished 


AlexandreBonaventure:
Hi,
This will allow more control on transitions. My use-case is a transition in two steps :
`giving an array model like [1, 2, 3]`

1
2
3

**1.  that reorder elements (`[2, 3, 1]`) with move transition**

2
3
1

**2.  that remove elements (`[2]`)**
2
<- 3
<- 1
 
If I had a hook I will be firing the second step after 1 has finished moving. Instead, as a workaround, I'm firing it after a timeout.

https://github.com/vuejs/vue/blob/ae93d2bef8e20402ae3501aa5cdcf14b0ee19c12/src/platforms/web/runtime/components/transition-group.js
Looking at the sources, it seems to be pretty trivial to implement, since moving is all about setting and remove class. We could fire an event after all transitionend events are finished. Could provide a PR if accepted.

Thanks

Kingwl:
#3974

yyx990803:
see https://github.com/vuejs/vue/issues/3974#issuecomment-316372415

PR #4916, opened 2017-02-13
add ‘beforeMove’ and ’afterMove’ hook in transition-group - #4885 

Kingwl:
#4885


Kingwl:
It looks like this is not a good implement

PR #4923, opened 2017-02-13
Fix filters parsing incorrectly because of missing spaces (fix #4881)

### LINKED ISSUE #4881: Chaining multiple filters without spaces doesn't work
### Vue.js version
1.0.28

### Reproduction Link
https://jsfiddle.net/8L7f9u53/2/

### Steps to reproduce
1. Define few filters
2. Apply them each other

### What is Expected?
Pipelining (chaining) filters

Only first filter executes

---
```html
<div id="mount-point">
  <p v-text="test|foo"></p>      <!-- string.foo -->
  <p v-text="test|bar"></p>      <!-- string.bar -->
  <p v-text="test|foo|bar"></p>  <!-- string.foo but expected string.foo.bar -->
</div>
```
```js
Vue.filter('foo', function(s) {
  return s + '.foo';
})
Vue.filter('bar', function(s) {
  return s + '.bar';
})

new Vue({
  el: '#mount-point',
  data: {
    test: 'string'
  }
})
```

baizulin:
Fixing parseFilter returning only the first filter of a chain in case of expression missing spaces.
See #4881

### DIFF
--- a/src/parsers/directive.js
+++ b/src/parsers/directive.js
@@ -115,9 +115,7 @@ function parseExpression () {
       if (chr === pipeChr) {
         next()
       } else {
-        if (state === startState || state === filterArgState) {
-          state = filterState
-        }
+        state = filterState
         break
       }
     } else if (chr === spaceChr && (state === filterNameState || state === filterArgState)) {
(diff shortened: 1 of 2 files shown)


baizulin:
This issue is absent in 1.1 and 2.0 branches because of a different approach to parsing. It would still be great to add tests for this case. I've added the test for 1.1 in my fork, will make a PR if everything will be good with this one.
It's my first contibution to Vue.js so please explain patiently if I did something wrong :)

baizulin:
guys?

neobats:
Saw this come through my inbox via CodeTriage. So, in the spirit of trying to be helpful, should this be merged? Is the fix already implemented in existing release? Hoping this helps to resolve a PR that's been stale for a while.

posva:
ah yeah, this PR is for v1 which is not receiving any updates anymore. We can close this

Issue #3961, opened 2016-10-17
[feature request] Add 'max-stack' in component 'keep-alive'


leaves4j:
like this:

``` html
<keep-alive :max-stack='10'>
    ....
</keep-alive>
```

vue will only cache 10 switched-out components

posva:
When does this come in handy? @fnlctrl

fnlctrl:
Things like an infinite list/virtual scroll, where there could be many many
detached but kept-alive components.

yyx990803:
This seems to have little traction and the virtual list use case is imo not what `<keep-alive>` was designed for. Also cached components can already be to some extent controlled via the `include` and `exclude` props.

PR #4930, opened 2017-02-14
support max-alive in keep-alive - #3961 

Kingwl:
#3961

### DIFF
--- a/src/core/components/keep-alive.js
+++ b/src/core/components/keep-alive.js
@@ -47,11 +47,32 @@ export default {
 
   props: {
     include: patternTypes,
-    exclude: patternTypes
+    exclude: patternTypes,
+    maxAlive: Number
+  },
+
+  methods: {
+    toCache (key: string | number, vnode: VNode) {
+      if (this.cache[key]) {
+        vnode.componentInstance = this.cache[key].componentInstance
+      } else {
+        if (this.maxAlive > 0) {
+          while (this.queue.length >= this.maxAlive) {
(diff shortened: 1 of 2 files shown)


znck:
This PR is based to `dev` branch. Then why is it tagged `1.x`?

src/core/components/keep-alive.js
```diff
@@ -47,11 +47,32 @@ export default {
...
+  methods: {
+    toCache (key: string | number, vnode: VNode) {
+      if (this.cache[key]) {
+        vnode.componentInstance = this.cache[key].componentInstance
+      } else {
+        if (this.maxAlive > 0) {
+          while (this.queue.length >= this.maxAlive) {
+            const willDestroyKey = this.queue.shift()
+            pruneCacheEntry(this.cache[willDestroyKey])
+            this.cache[willDestroyKey].data.keepAlive = false
```
Mafieu: Isn't that a useless line since you put everything to `null` next line ?

Kingwl: no it isn't
next line set a object property (a reference) only

Mafieu: Oh, that reference is used somewhere else then.
Thanks for the answer.

Kingwl: :XD

posva:
what happened @Kingwl 🙁

znck:
I want this too 😞

Kingwl:
😂😂refact.....

PR #4933, opened 2017-02-14
Added the possibility to have function declared as required propOption

vTripes:
Added the possibility to declare a function as value for required PropOption which should return a truthy or falsy value as the following:
```JavaScript
props: {
  myProp: {
    type: String,
    required: function () {
      /* Treatment here */
      return result // truthy or falsy
    }
  }
}
```

### DIFF
--- a/flow/options.js
+++ b/flow/options.js
@@ -70,6 +70,6 @@ declare type ComponentOptions = {
 declare type PropOptions = {
   type: Function | Array<Function> | null;
   default: any;
-  required: ?boolean;
+  required: ?any;
   validator: ?Function;
 }
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -85,9 +85,9 @@ function assertProp (
   vm: ?Component,
   absent: boolean
 ) {
-  const required = prop.required instanceof Function ?
-        prop.required() :
-        prop.required
+  const required = prop.required instanceof Function 
+    ? prop.required()
+    : prop.required
   if (required && absent) {
     warn(
       'Missing required prop: "' + name + '"',
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -7,7 +7,7 @@ import { warn } from './debug'
 type PropOptions = {
   type: Function | Array<Function> | null,
   default: any,
-  required: ?boolean,
+  required: ?any,
   validator: ?Function
 };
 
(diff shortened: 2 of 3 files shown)


posva:
Why do you need the required property to change? I can understand it changing depending on other props (eg one of two different props being required) but in that case, you wouldn't be able to rely on it because of the order. What is your usecase?

vTripes:
The case I needed it is a component which has two different "states" (one to create and another to modify).
Thus, there are some properties that are needed in a state, but not in the other.
So here would be the case:
```JavaScript
props: {
  state: {
    type: String,
    required: true,
    validator: function (val) {
      return !!~['create','edit'].indexOf(val)
    }
  },
  content: {
    type: String,
    required: function () {
      return this.state === 'edit'
    }
  }
}
```
If you are in edit mode, you have to pass some content as you can't have something to edit where there's nothing, and you can have pre-loaded content in create mode.

Of course, the order of the props is important in that case, as not all of it will be loaded, but it could give some more power and flexibility to props.

posva:
> Of course, the order of the props is important in that case, as not all of it will be loaded, but it could give some more power and flexibility to props (still there are some workarounds).

What I was saying is that using an object doesn't let you rely on the order. Furthermore, you may want something like

```js
props: {
  one: {
    required () {
      'one' in this || 'two' in this
    },
  two: {
    required () {
      'one' in this || 'two' in this
    }
}
```

But that'd require the `required` function to be called after all props have been initialized.
IMO this is more of a validator function but it's true that validator doesn't have access to `this` (it's a pure function).

yyx990803:
imo relying on other props in the initialization phase can be problematic because the Object enumeration order is inconsistent in different JavaScript engines - e.g. your code may work in Chrome but break in Safari.

I think it's safer to just check and emit your custom warning in the `created` hook of your component when all props have been initialized.

vTripes:
That's right that browsers are really capricious with objects.
Well, seems like I'll stay with workarounds 👍

posva:
I recommend you to add checks on the `created` hook and display warnings that are stripped off in when bundling in production (this needs some Webpack/rollup configuration):

```js
// inside of created
if (process.env.NODE_ENV !== 'production') {
// check here
}
```

If you used the webpack template for your project, you don't need to configure anything 🙂

PR #4949, opened 2017-02-16
Ignore tag modifiers like <div.active@idname> when closing tags

shreeve:
Helps the parser to know that when using tags like `<div.active@menu>`, a simple `</div>` is enough to close the tag.


yyx990803:
Again... thanks for opening pull requests, but please beware that new features added to the framework need to be discussed for its use cases, tradeoffs, potential implications etc. The proposed change is essentially adding new non-standard syntax into the template language, which I don't believe is valid HTML even on the syntax level.

PR #4950, opened 2017-02-17
Warn when defining a method with same name as a prop

chriscasola:
When you specify a method on a component with the same name as a prop you get an error about mutating a prop directly, but it can be hard for people new to vue to figure out what this means.

This PR adds a more explicit warning message indicating exactly what went wrong.


rohmanhm:
That's was great 👍🏼

chriscasola:
@posva thanks for the feedback. Updated.

yyx990803:
Thanks!

Issue #4957, opened 2017-02-17
__vue__ property leaked in DOM element, even in vue.min.js


andreiglingeanu:
### Vue.js version
dev branch

### Steps to reproduce
1. Open [this](https://github.com/vuejs/vue/blob/dev/dist/vue.min.js)
2. Search for `__vue__`

### What is Expected?

To not see any `__vue__` leaks onto the actual DOM element.

### What is actually happening?
Vue instance is stored inside `__vue__` onto the DOM element itself.

<hr>

Is this by purpose? I think the only usage for `__vue__` property is to make the instances discoverable by `vue-devtools`. Should that be present in minified build? I can come up with a pull request with a fix for that.

yyx990803:
Minified build can still use the devtools when `Vue.config.devtools` is explicitly set to `true`. This is intended.

andreiglingeanu:
This means that if I got some `__VUE_DEVTOOLS_GLOBAL_HOOK__` in my `window` object - devtools will walk through my _all_ DOM elements.

That's the same problem I got in #4958 - multiple Vue dependencies load on the same page.

andreiglingeanu:
@yyx990803 Can we at least go about removing those `__vue__` properties based on an ENV variable, other than `NODE_ENV='production'`?

I'll be happy to build Vue by myself.

I really need it to be as closed from the outside world as possible.

yyx990803:
Why?

andreiglingeanu:
I don't want the components compiled under production env to be detectable by devtools. This happens if I have components that are compiled under development env.

yyx990803:
It's only **detectable**, but not **inspectable**. The devtool will walk your page nodes anyway regardless of whether `__vue__` is exposed or not.

andreiglingeanu:
They __do__ appear in devtools and they clash with other components, if they have the same `_uid`.

yyx990803:
Okay, then I think the root of the problem is in the clash of multiple copies of Vue on the same page, which I strongly recommend you to avoid.

andreiglingeanu:
Yep, that is right and I totally can't avoid that, even if I would want to. Basically, I'm creating third-party components that can be just plugged in (multiple number of them, quite often on the same page) somewhere and they should avoid clashing one with another as much as possible.

It is really a pain the ass to develop them if devtools are broken in such a way.

yyx990803:
Can't these components all just use a shared global Vue?

andreiglingeanu:
[...]

PR #4959, opened 2017-02-17
add process.env.VUE_DISMISS_DEVTOOLS guard around __vue__ property

andreiglingeanu:
By following up from #4957, I'm adding that guard against `__vue__` properties.

This won't change existing builds in any way, as long as you don't have a `VUE_DISMISS_DEVTOOLS` environment variable.

@yyx990803 Maybe there should be a better name for that thing?


yyx990803:
See https://github.com/vuejs/vue/issues/4957#issuecomment-280691805

PR #4973, opened 2017-02-20
a bit more random _uids

andreiglingeanu:
Continuing the work done in https://github.com/vuejs/vue-devtools/pull/260, I'm proposing a change for a bit more dynamic `_uid`s, by adding a small random offset to them.

This will allow components from multiple Vue dependencies to be inspectable in `vue-devtools` without any hassle. Also, one thing to note is the fact that [transitions](https://github.com/vuejs/vue/blob/f7062b9b75179a0a515205e47cfe6227f33b95f0/src/platforms/web/runtime/components/transition.js#L134) may break in some rare cases.

### DIFF
--- a/src/core/instance/init.js
+++ b/src/core/instance/init.js
@@ -13,7 +13,7 @@ export function initMixin (Vue: Class<Component>) {
   Vue.prototype._init = function (options?: Object) {
     const vm: Component = this
     // a uid
-    vm._uid = uid++
+    vm._uid = (Math.floor(Math.random() * 100) + 1) + uid++
     // a flag to avoid this being observed
     vm._isVue = true
     // merge options


vectart:
@andreiglingeanu simple test proofs that your solution makes them even less unique:
```javascript
let uids = []

for (let uid = 0 ; uid < 1000; uid++) {
  uids.push((Math.floor(Math.random() * 100) + 1) + uid)
}

for (let i = 0; i < uids.length; i++) {
  let uid = uids[i]
  let matches = uids.filter(_uid => _uid === uid).length
  if (matches > 1) {
    console.warn(`${uid} generated ${matches} times`)
  }
}
```

image

andreiglingeanu:
@vectart This will be guaranteed to not break anything in the current projects because we are doing the `uid` increment besides relying on the randomness of `Math.floor(Math.random() * 100) + 1`.

Actually, increasing the maximum number for the `Math.random` expression will make those repetitions occur not that often. And, it won't be the end of the world if they'll repeat.

Do you think bringing a general purpose uuid algorithm will be more suitable?

cc. @yyx990803

andreiglingeanu:
With enough _randomness_ I'm not getting repetitions at all.

yyx990803:
Unfortunately I don't think this would work. Although it makes uids from multiple copies of Vue less likely to clash, it also makes the uids **less* unique when there's only a single copy of Vue. "A big random number" is simply not safe.

In addition, the internal scheduler relies on an important characteristics of the uids: the earlier an instance is created, the smaller its uid is. The scheduler uses this to determine job priorities.

I think a better fix to your original problem (multiple copies of Vue with vue-devtools) is to detect multiple copies of Vue and assign each copy an id, then prefix all instance ids with the Vue id they are associated with.

andreiglingeanu:
Got that. Thank you. I'll give a thought to your idea.

yyx990803:
@andreiglingeanu btw what I described is a change in devtools not in Vue core

andreiglingeanu:
Yeah, I got thank you once again.

PR #4975, opened 2017-02-20
Adding the ability to suppress notifications in Vue.set and Vue.delete

sffc:
Currently, whenever you call `Vue.set` or `Vue.delete`, all observers on the object being modified are executed.  This is a "safe" practice, because it ensures that any objects who had previously attempted to access the key being added or deleted get their dependencies re-computed.

However, consider the case where you have an object with many different keys and watchers, which might be expensive to call.  In this case, `Vue.set` will do a lot of work executing all of the watchers, even if none of them respond to the new field in the object, causing poor performance.

Here is a fiddle to demonstrate the issue: https://jsfiddle.net/exe3nxzg/2/ .  Adding new keys to the object causes the watcher on the original field to be triggered, even though that watcher doesn't care about the values of the new keys.

This PR offers a possible fix to the problem, which is to add an options object as the final argument of `Vue.set` and `Vue.delete`.  If the user wishes to manually suppress notifications on the object and prevent all watchers from being computed, they can pass `{ notify: false }` as the final argument.

Another possible fix would be to add a new function called `Vue.get`.  If `Vue.set` replaces the syntax `obj[key]=value` and `Vue.delete` replaces the syntax `delete obj[key]`, then `Vue.get` would replace the syntax `obj[key] || default`.  More precisely, a definition could look something like:

```js
Vue.get = function(obj, key, defaultValue) {
	if (!obj.hasOwnProperty(key)) {
		// create the key on the object with the default value.
		// call defineReactive(), etc.
		// however, do NOT call notify on the object.
	}
	return obj[key];
};
```

This way, whenever a key access is attempted, a reactive is left behind, such that setting the property in the future, even via `Vue.set`, will not cause a full notify.

Thoughts?


src/core/observer/index.js
```diff
@@ -211,14 +211,17 @@ export function set (obj: Array<any> | Object, key: any, val: any) {
     return
   }
   defineReactive(ob.value, key, val)
-  ob.dep.notify()
+  if (typeof options !== "object" || !hasOwn(options, "notify") || options.notify) {
```
LinusBorg: Tests failed because you didn't follow the lint rules. We only use single quotes.

sffc:
I fixed the lint failures.

test/unit/modules/observer/observer.spec.js
```diff
@@ -268,6 +268,17 @@ describe('Observer', () => {
     // should ignore deleting non-existing key
     delProp(obj1, 'a')
     expect(dep1.notify.calls.count()).toBe(3)
+    // should ignore notifications if { notify: false } is set
+    setProp(obj1, 'd', 1, {}) // should notify
+    setProp(obj1, 'e', 1, { foo: 'bar' }) // should notify
+    setProp(obj1, 'f', 1, { notify: true }) // should notify
+    setProp(obj1, 'g', 1, { notify: false }) // should NOT notify
```
posva: you should test the amount of calls before this line as well

test/unit/modules/observer/observer.spec.js
```diff
@@ -268,6 +268,17 @@ describe('Observer', () => {
...
+    expect(dep1.notify.calls.count()).toBe(6)
+    delProp(obj1, 'd', {}) // should notify
+    delProp(obj1, 'e', { foo: 'bar' }) // should notify
+    delProp(obj1, 'f', { notify: true }) // should notify
+    delProp(obj1, 'g', { notify: false }) // should NOT notify
```
posva: same here

yyx990803:
For why it needs to trigger notification for anything that depends on the object itself, consider the following expression:

``` js
a.b + c
```

When `a` is still `{}`, `a.b` will not be able to detect a dependency on the key `.b`, because the getter doesn't exist yet. The only way to ensure this expression gets re-evaluated when `.b` is created, is by registering the entire object `a` as the expression's dependency. This is simply how Vue's reactivity system works given the current constraints.

That said, I don't really understand the use case for this PR. The only reason you'd use `Vue.set` is to explicitly trigger updates. If you don't trigger a notification on the object itself at the time of registering a new key, then expressions like the one mentioned above will never know a new key was added and will never update.

sffc:
Here's my use case.  I have an object that caches results from I/O operations, indexed by their filename.  For example, `data.cache["/path/to/filename.png"]`.  A watcher can subscribe to one or more of the keys in the cache, and they should update whenever their particular field changes.  At an instant in time, the cache contains around 20 keys.  However, each field in the cache can have multiple watchers, so there may be several hundred watchers.  Triggering a notify on every single watcher when a new key is added to the cache is very expensive.

My pattern for subscribing to a new key currently looks like:

```
if (!cache.hasOwnProperty(key)) {
	Vue.set(obj, key, defaultValue, { notify: false });
}
// do stuff with cache[key]
```

In other words, I'm careful to make sure that the field exists prior to accessing it via the `[]` operator, in order to avoid the problem Evan mentioned.

Giving this a second thought, I think the best way forward might be to add the *Vue.get()* function instead of the `{ notify: false }` option parameter.  The implementation of *Vue.get()* would be something along the lines of:

```js
/**
 * Gets a property from an object, creating the property with the specified default value
 * if it does not exist.
 */
export function get (obj: Array<any> | Object, key: any, defaultVal: any) {
  if (hasOwn(obj, key)) {
    return obj[key]
  }
  const ob = obj.__ob__
  if (!ob) {
    return obj[key]
  }
  defineReactive(ob.value, key, defaultVal)
  return obj[key]
}
```

This way, when you are tempted to write `a.b + c`, you can instead write `Vue.get(a, "b", 0) + c` and ensure that you are subscribed to future changes to that field.  Thoughts?

simplesmiler:
> Triggering a notify on every single watcher when a new key is added to the cache is very expensive.

Watcher expressions (the first argument to `watch`) are not designed to be heavyweight. Complex logic should go to the watcher handler instead.

Your watcher handler gets executed exactly because you artificially return an object from the watch expression. If you weren't doing that, there's won't be a problem in the first place.

Here is your modified demo with five thousand watchers (an order of magnitude more than several hundred), giving decent performance: https://jsfiddle.net/simplesmiler/exe3nxzg/3/. With this much watchers (assuming you have a few watchers per component) the performance of watcher expressions would not be a bottleneck.

Can you provide a reproduction where executing several hundred watch expressions (without artificially returning distinct values) actually causes performance issues?

sffc:
My watchers do image manipulations, which are synchronous activities that run on the order of milliseconds.  A hundred watchers and you're waiting close to a second for everyone to re-compute.  [Here is an example](https://github.com/sffc/vue-watcher-demo) for illustration.  Run the example using `npm start`, which will start up a local server.  On the server, click the "Load Next Image" button a few times.  Even after just 5 or 6 clicks, there is a noticeable lag.

Also, I noticed that [in core/instance/state.js](https://github.com/vuejs/vue/blob/dev/src/core/instance/state.js#L142), functions you set as a "computed" simply get delegated as watcher expressions.  It says [in the guide](https://vuejs.org/v2/guide/computed.html#Computed-Caching-vs-Methods) (emphasis mine):

> Why do we need caching? Imagine we have *an expensive computed property A, which requires looping through a huge Array and doing a lot of computations*. Then we may have other computed properties that in turn depend on A. Without caching, we would be executing A’s getter many more times than necessary! In cases where you do not want caching, use a method instead.

In other words, the guide seems to indicate that it's OK to have an expensive function as your computed, and in turn as your watcher expression.

*Even if* the watcher expressions are lightweight, I still don't like the idea of re-computing them when we know for sure that their dependencies haven't changed.  If I have a watcher listening on `cache.a`, creating `cache.b` logically shouldn't cause the other watcher to re-compute as it does now.

---

I've been hacking around in my copy of Vue, and I committed two new features to my branch.  I reverted the `{notify:false}` suggestion.

1. A modification to *Vue.delete()* that triggers notifications only on subscriptions to the property that was deleted, instead of globally to the object.  This is purely an internal optimization; the external behavior is the same, as all watchers that need to be notified get notified.
2. An implementation of *Vue.get()*, the function I had mentioned previously.  I actually like the *Vue.get()* API better than the `{notify:false}` API because if you use it correctly, your app will get faster, and if you don't use it, you simply get the old behavior.

Feature 1 doesn't seem to have any real downsides, and there are no changes to the API.

The gotcha with Feature 2 is that if you mix call sites (with some using *Vue.get()* and with others using the `[]` operator), the call sites using `[]` risk never getting their subscriptions added.  This only happens if you mix `[]` with *Vue.get()* on the same property, and there is a test case demonstrating it.  However, the guide already recommends that you use the sister functions *Vue.set()* and *Vue.delete()* globally, so as long as this restriction is well-documented, I think it should be okay.

test/unit/modules/observer/observer.spec.js
sffc: Acknowledged. I changed the PR to not use the `{notify:false}` API anymore, but I still have it saved in a branch if we want to restore it.

yyx990803:
So - I appreciate the feedback and effort put into this PR, but I'm not convinced of the tradeoff of adding a new API for such a specific use case. In particular, the original perf problem can be circumvented by checking if the specific key has the same value (since you already know which key you depend on).

In the future Proxy-based reactivity system, this problem will go away without the need for an explicit API. Adding an API in the current version to solve a problem that doesn't exist in a future version means we will have to remove it when the new version lands, so ideally we want to avoid that.

PR #4994, opened 2017-02-23
Add support for array as inline style value (client/ssr)

fnlctrl:
When exploring [css-in-js solutions](https://github.com/MicheleBertoli/css-in-js) for vue, I came across the fact that some auto prefixed style objects (for example, by [postcss-js](https://github.com/postcss/postcss-js)) cannot be directly used as `style` option on elements, when they have an array as value, e.g.
```
{ display: ['-webkit-box', '-webkit-flex', '-ms-flexbox', 'flex'] }
```
...which happens when a rule value needs to be prefixed, other examples:
```
{ backgroundImage: ['radial-gradient(...)', '-webkit-radial-gradient(...)']
{ cursor: ['-webkit-grab','grab'] }
```
This PR adds support for directly consuming such objects as `style` option.        
         
     
      

This is really an edge case (inline style option + style object with value as an array), so I totally understand if this is not going into the core.
In fact I only need this for rendering email templates that require inline styles on html elements with vue SSR, though it seems nice to add support for such usage maybe to give people more freedom when choosing css-in-js solutions.


yyx990803:
Can we add test cases for these?

fnlctrl:
I'll add it asap.

fnlctrl:
Oops... messed up the log when rebasing onto dev, will open another PR with tests....

PR #5024, opened 2017-02-27
compiler codegen improvement

xujiongbo:
Inside genChildren function, not need to invoke getNormalizationType() when checkSkip is false.


src/compiler/codegen/index.js
```diff
@@ -310,11 +310,9 @@ function genChildren (el: ASTElement, checkSkip?: boolean): string | void {
         el.tag !== 'slot') {
       return genElement(el)
     }
-    const normalizationType = getNormalizationType(children)
+    const normalizationType = checkSkip ? getNormalizationType(children) : false
```
yyx990803: Can we use `0` here so that this value is always a number?

xujiongbo: Yeah, that's more reasonable because 0 also means no normalization needed

yyx990803:
Thanks!

PR #5039, opened 2017-02-28
Add the runtime-only warning for el in-DOM HTML template

KingMario:
Make the warning more explicit for template complier.
Plus, el option which will be used to extract the template should also emit the warning.

### DIFF
--- a/src/core/instance/lifecycle.js
+++ b/src/core/instance/lifecycle.js
@@ -141,8 +141,8 @@ export function mountComponent (
     vm.$options.render = createEmptyVNode
     if (process.env.NODE_ENV !== 'production') {
       /* istanbul ignore if */
-      if (vm.$options.template && vm.$options.template.charAt(0) !== '#' ||
-        vm.$options.el) {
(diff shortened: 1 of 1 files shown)


posva:
`el` is also used in runtime only:

```js
new Vue({
  el: '#app',
  render: h => h(App)
})
```

KingMario:
@posva, The warning is emitted when there's no render function (handwritten or compiled), so it's no problem whether `el` is used or not in runtime only.

posva:
@KingMario You're right, I didn't look the code above and thought this warning was a different one (`Failed to mount component: template or render function not defined`)

posva:
Can you add a test then, please?

KingMario:
The tests passed in my local build.
Have no idea.

KingMario:
Added npm run build in ci.sh.
But don't know whether it's acceptable.

posva:
@KingMario I just realised there're no tests for that, sorry. I removed the extra commits and added a check for `el`

PR #5047, opened 2017-03-01
NonParsingTags should not be case-insensitive

jingsam:
In vue 2.0, the template is parsed by a Virtual DOM. The paring process is case-sensitive. For example, the VDOM can distinguish between `<Button>` and `<button>`. However, the function is `isNonParsingTag` is case-insensitive which shouldn't be.


yyx990803:
Thanks, good catch.

icarusion:
Bravo

PR #5049, opened 2017-03-01
[weex] wrap IFFE and use strict mode to run js bundle

Hanks10100:
This is a corresponding change with https://github.com/alibaba/weex/pull/1483 and https://github.com/alibaba/weex/pull/1626 .

Both legacy `.we` and  Rax are using strict mode to run the js bundle.

### DIFF
--- a/src/platforms/weex/framework.js
+++ b/src/platforms/weex/framework.js
@@ -102,6 +102,10 @@ export function createInstance (
     weex: weexInstanceVar,
     __weex_require_module__: weexInstanceVar.requireModule // deprecated
   }, timerAPIs)
+
+  // wrap IFFE and use strict mode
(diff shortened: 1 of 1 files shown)


Hanks10100:
Some test case failed because of the `with(this){}` . How could I use https://github.com/vuejs/vue-template-es2015-compiler to compile the template without `with` in test case ?

yyx990803:
I think `vue-template-es2015-compiler` needs to become a dependency of `weex-template-compiler` and used by default in its `compile` method.

This is the relevant part in `vue-loader`: https://github.com/vuejs/vue-loader/blob/master/lib/template-compiler.js#L82-L93

Hanks10100:
The `weex/framework.js` has changed a lot. This PR has no meaning to modify.

PR #5052, opened 2017-03-01
[weex] Support unary and left open tags

Hanks10100:
Fix https://github.com/weexteam/weex-vue-loader/issues/8 .

And add more reserved tags.

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -5,6 +5,7 @@ declare type CompilerOptions = {
   staticKeys?: string; // a list of AST properties to be considered static; for optimization
   directives?: { [key: string]: Function }; // platform specific directives
   isUnaryTag?: (tag: string) => ?boolean; // check if a tag is unary for the platform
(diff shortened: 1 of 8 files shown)


yyx990803:
I think currently `html-parser.js` is hard coded to use `canBeLeftOpenTag` from `platforms/web`: https://github.com/vuejs/vue/blob/dev/src/compiler/parser/html-parser.js#L13

We will need to change the compiler to pass on `canBeLeftOpenTag` as an option to `parser/index`, and then pass it on to `html-parser`. The web compiler entry also needs to be updated.

Hanks10100:
Actually only need to implement the `isUnaryTag` can fix that bug. I also add a test case for using unary tag in Weex.

Anyway, I also moved `canBeLeftOpenTag` to the compiler option, it can be passed form entry. The `baseOptions` of web compiler and the flow type of `CompilerOptions` has also been updated.

PR #5059, opened 2017-03-02
SFC parseComponent pads complete content with spaces

### LINKED ISSUE #5058: Feature request: parseComponent's pad: true option should use spaces for padding
When an external tool is using vue-template-compiler to grab the script block, `{ pad: true }` fills in preceding non-script lines with `//` so that line+column-oriented tools can get the line numbers right. 

However, this doesn't work as well for character-oriented tools like the Typescript language service, which treat positions as a single index into a single string. It would be better if `{ pad: true }` replaced the contents of all non-script blocks with spaces.

For example:

```js
import { parseComponent } from 'vue-template-compiler'
const desc = parseComponent(`<template>
blah
</template>
<script>
export default {
  data: { example: 12 }
  // etc etc
}
</script>`, { pad: true })
console.log(desc.script.content)
```

Current behaviour prints:

```
//
//
//

export default {
  data: ...
```

Desired behaviour prints:

```
......................................................
.......................................
....
...........
........
export default {
  data: ...
```

except instead of `.`, it should print ` `.

## Motivation

For background, I'm working on a Vue plugin for the Typescript language service so that editors like VS Code, Visual Studio, Sublime, emacs, etc, can get Javascript and Typescript completions inside .vue files. The next version of Typescript will support full completions on Vue and Vue options objects with no type annotations, and the plugin will make that work inside `.vue` files as well as `.js` and `.ts` files.

I tried using the `start/end` offsets returned from `parseComponent`, which kind of works, but Typescript has previously only used offsets for non-critical parsing like JSDoc. I'd like to use the space-replacing technique since I know it has worked in the past, for example in Visual Studio's handling of Javascript embedded in HTML.

## Implementation

Replace the body of `padContent` in `src/sfc/parser.js` with `return content.slice(0, block.start).replace(/./g, ' ')`. This correctly avoids replacing both `\n` and `\r\n`, at least on node 7.4 on Linux.

I'll send a PR as soon as I get Vue building and have added a test.

sandersn:
SFC `parseComponent` pads all content with spaces when `{ pad: true }` is provided. That is, all content is converted to spaces. Previously, each line was truncated to `//`. The new padding method works better with character-oriented tools that calculate positions by distance from the beginning of the file instead of by line
number.

Fixes #5058.

### DIFF
--- a/packages/vue-template-compiler/README.md
+++ b/packages/vue-template-compiler/README.md
@@ -85,4 +85,9 @@ Parse a SFC (single-file component, or `*.vue` file) into a descriptor (refer to
 
 #### Options
 
-- `pad`: with `{ pad: true }`, the extracted content for each block will be padded with newlines to ensure that the line numbers align with the original file. This is useful when you are piping the extracted content into other pre-processors, as you will get correct line numbers if there are any syntax errors.
+#### `pad`
+
+`pad` is useful when you are piping the extracted content into other pre-processors, as you will get correct line numbers or character indices if there are any syntax errors.
+
+- with `{ pad: "line" }`, the extracted content for each block will be prefixed with one newline for each line in the leading content from the original file to ensure that the line numbers align with the original file. 
+- with `{ pad: "space" }`, the extracted content for each block will be prefixed with one space for each character in the leading content from the original file to ensure that the character count remains the same as the original file.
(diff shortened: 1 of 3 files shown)


yyx990803:
This is an interesting change - one of the implications (and the reason why I used comments before) is when used with a linter that checks whitespaces or blank lines, this would raise a lot of warnings... (e.g. `unnecessary whitespace / newlines`)

sandersn:
Good idea. I didn't think about the linter case. I'll change the type of pad to `true | false | "space" | "line"`, where both "line" and true give the old padding behaviour.

PR #5060, opened 2017-03-02
fix mouse modifier override the keycode ( Fixed #5046 )

### LINKED ISSUE #5046: v-on:keydown.right and left seems to be broken in Vue 2.2.1
Hi,

Upgrading to 2.2.1 broke the following line <li v-on:keydown.right="focusMenu(1)"> Right and left modifiers on keydown seems to be broken, but v-on:keydown.39 is still working.

Thanks in advance,
grofzero

Kingwl:
#5046


posva:
ping @yyx990803 This PR fixes a regression

yyx990803:
Thanks, figured out a better way to keep the same usage by checking `'button' in $event` in 3294eba

Issue #5067, opened 2017-03-02
State of provide/inject


znck:
Provide/Inject allows ancestor component to serve as a dependency injector. Injected dependencies are not reactive. As per docs, if an observed property is provided it remains reactive.

I've created a fiddle (https://jsfiddle.net/1hcpbrjz/6/) to demonstrate few scenarios.

This feature is very new, so docs are too shallow.

Question: How to provide reactive data props and computed prop?

> Update: 3rd March

( #5075 ) Initialising injections before state would allow using injections in props. https://jsfiddle.net/fsurfc5w/

znck:
@Akryum Can you update the fiddle with an example? 

I think we should add some examples in docs too.

gefangshuai:
https://github.com/vuejs/vue/pull/5075

gefangshuai:
https://github.com/vuejs/vue/pull/5229

PR #5075, opened 2017-03-03
Init injections before state

rayrutjes:
This makes it possible to provide an injected value as the default for a prop

```js
export default {
  inject: ['_searchStore'],
  props: {
    searchStore: {
      type: Object,
      default () {
        return this._searchStore
      }
    }
  }
}
```

That way the user can either explicitly pass the property, or otherwise we resolve it with the injected one.

From my understanding, `initInjections` doesn't really conflict with `initState`, is there a specific reason for having initState before initInjections?

### DIFF
--- a/src/core/instance/init.js
+++ b/src/core/instance/init.js
@@ -49,8 +49,8 @@ export function initMixin (Vue: Class<Component>) {
     initEvents(vm)
     initRender(vm)
     callHook(vm, 'beforeCreate')
-    initState(vm)
     initInjections(vm)
+    initState(vm)
     callHook(vm, 'created')
 
     /* istanbul ignore if */


kazupon:
Thank you for your PR!

`provide/inject` feature improvement is discussioning in #5067 becase of some issues.
You can join to it. 😺

yyx990803:
Thanks for the suggestions! Unfortunately simply moving the whole block would break the case where `provide()` rely on data/props. I implemented a proper change with tests in 16765db.

rayrutjes:
Arg, sorry about the noise, I naively looked at the code and read the tests expecting them to pass o_O

Thank you @kazupon for the link, I will definitely join the conversation, as this seems like a very promising feature that fills some gaps.

Thank you @yyx990803 for being so reactive and constructive, really appreciated!

rayrutjes:
@yyx990803 do you have an idea about the ETA of the next release? I'm really looking forward to use the new implementation you came up with ;)

yyx990803:
@rayrutjes that may take a few days

rayrutjes:
That sounds great, using a local build of Vue in the meantime.
I can tell that the implementation you did is perfect for my use case.

PR #5085, opened 2017-03-03
Remove unnecessary else statement after return

sqal:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactoring
- [ ] Build related changes
- [ ] Other... Please describe:

- [ ] Yes
- [x] No
`

**Other information**:

Maybe we should add [no-else-return](http://eslint.org/docs/rules/no-else-return) rule to the eslint config?


posva:
I think it's a good idea to turn it on. I suppose you used `eslint --fix`, didn't you?

yyx990803:
I actually prefer explicit branches. I only use early return for "short circuiting" situations.

PR #5102, opened 2017-03-06
chore(parser.spec): fix an IE conditional test with a typo

S-YOU:
**Please check if the PR fulfills these requirements**
- [x] Do not submit PRs against the master branch.
- [x] Checkout a topic branch from the relevant branch, e.g. dev, and merge back against that branch.
- [x] The tests are running: https://github.com/vuejs/vue/blob/dev/.github/CONTRIBUTING.md#development-setup

If adding new feature:
- [ ] Added accompanying test case.
- [ ] In the commit message, provide convincing reason to add this feature.
      Ideally you should open a suggestion issue first and have it greenlighted before working on it.

If fixing a bug:
- [ ] If you are resolving a special issue, add (fix #xxxx[,#xxx]) (#xxxx is the issue id) in your PR title
- [ ] Provide detailed description of the bug in the PR. Live demo preferred.
- [ ] Add appropriate test coverage if applicable.

**What kind of change does this PR introduce?** (check at least one with "x")
```
[ ] Bugfix
[ ] Feature
[ ] Code style update
[x] Refactoring
[ ] Build related changes
[ ] Other... Please describe:
```

**Does this PR introduce a breaking change?** (check one with "x")
```
[ ] Yes
[x] No
```

If this PR contains a breaking change, please describe the impact and migration path for existing applications:

**Other information**:
This is unconfirmed, since I don't have IE at the moment.

### DIFF
--- a/test/unit/modules/compiler/parser.spec.js
+++ b/test/unit/modules/compiler/parser.spec.js
@@ -494,18 +494,16 @@ describe('parser', () => {
     expect(ast.children[0].text).toBe('1 < 2 < 3')
   })
 
-  if (isIE) {
-    it('IE conditional comments', () => {
-      const options = extend({}, baseOptions)
-      const ast = parse(`
-        <div>
-          <!--[if lte IE 8]>
-            <p>Test 1</p>
-          <![endif]-->
-        </div>
-      `, options)
-      expect(ast.tag).toBe('div')
-      expect(ast.children).toBeUndefined()
-    })
-  }
+  it('IE conditional comments', () => {
+    const options = extend({}, baseOptions)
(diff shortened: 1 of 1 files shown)


posva:
I think we want the test to work when it's not IE too so we know the parser can handle it, but I may be wrong

S-YOU:
@posva, The test fail somehow on my machine. Since it is a parser, I am afraid that it would fail on IE too.

S-YOU:
@posva, @kazupon, I have changed to use children.length to be 0, instead of only running on IE.
But still actual behavaior on IE is unconfirmed, since I don't have it.

Issue #5029, opened 2017-02-27
Version of Bootstrap included in benchmarks includes links to files that do not exist


bzmwillemsen:
### Vue.js version
2.2.1

### Reproduction Link

The version of bootstrap in the Vue project located here: https://github.com/vuejs/vue/blob/v2.2.1/benchmarks/dbmon/lib/bootstrap.min.css
contains links to files that do not exist. Specifically: 
```
url(../fonts/glyphicons-halflings-regular.eot);src:url(../fonts/glyphicons-halflings-regular.eot?#iefix) format('embedded-opentype'),url(../fonts/glyphicons-halflings-regular.woff) format('woff'),url(../fonts/glyphicons-halflings-regular.ttf) format('truetype'),url(../fonts/glyphicons-halflings-regular.svg#glyphicons_halflingsregular)
```

### What is Expected?
That the files exist, or that the version of bootstrap that is compiled doesn't depend on font files.

### What is actually happening?
I'm trying to run a build tool with Django, called `collectstatic`, part of collectstatic is cache busting for links in `css` files, when it tries to follow these links it exits with an error because the file does not exist.

yyx990803:
The benchmark doesn't use those glyphs, and the benchmarks are only intended to be run standalone, not intended to be used elsewhere. If you need bootstrap, use proper setup per their latest docs.

bzmwillemsen:
> The benchmark doesn't use those glyphs, and the benchmarks are only intended to be run standalone, not intended to be used elsewhere.

Should the benchmarks be included in the bower package then? Personally I do not see a reason to include it in a bower package, just as `source`, and `.github` should not be included. Actually I just realized that there is no `bower.json` definition in the app, is that on purpose?

posva:
yes: https://github.com/vuejs/vue/issues/4040

PR #5123, opened 2017-03-07
Minimize size of bootstrap used in dbmon benchmark

bzmwillemsen:
**What kind of change does this PR introduce?** (check at least one with "x")
```
[ ] Bugfix
[ ] Feature
[x] Code style update
[ ] Refactoring
[ ] Build related changes
[ ] Other... Please describe:
```

**Does this PR introduce a breaking change?** (check one with "x")
```
[ ] Yes
[x] No
```

**Other information**:
In #5029 I outlined how the bootstrap library that is used in the dbmon benchmark references font files that are not included in the repo.
Rather than including them in the repo I thought a better approach would be to customize the bootstrap library to only include exactly what the benchmark actually needed.

The customization options are stored, by bootstrap, in the minified file. For quick reference you can find it [here.](http://getbootstrap.com/customize/?id=7bd1d75ceeb95d343d9e16d305e782cd)

Size of the library goes from 102KB to 49KB.

This resolves my original problem as the "Glyphicons" module of bootstrap is no longer included and the references to fonts have been removed from the library.

#### Update based on comments:
This PR now removes the bootstrap library that the js-repaint-perfs library was using in favour of using the `styles.css` that the library now uses.


yyx990803:
FYI the bootstrap file used was directly copied from https://github.com/mathieuancelin/js-repaint-perfs , but they have since removed the dependency on bootstrap. I think we can just use [this file](https://github.com/mathieuancelin/js-repaint-perfs/blob/gh-pages/styles.css) instead.

bzmwillemsen:
@yyx990803 thanks for pointing that out. I have made that change, and updated OP.

PR #5135, opened 2017-03-08
fix v-on unit test

xujiongbo:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

It seems that ```input.checked``` is unable to change through the previous way, so we will not get ```true``` even if when removing the ```prevent``` modifier.


yyx990803:
Can you send the PR from a new branch (based off latest dev) in your fork to exclude previous commits?

xujiongbo:
@yyx990803 sure

PR #5138, opened 2017-03-08
Add node using nextSibling as reference

finntl:

- [ x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ x] No

- [x ] It's submitted to the `dev` branch for v2.x (or to a previous version branch), _not_ the `master` branch
- [ x] All tests are passing: https://github.com/vuejs/vue/blob/dev/.github/CONTRIBUTING.md#development-setup

When an item is appended to the end of a list being rendered using v-for, it is appended as a new child to the parent element. If an element has been appended after the list in the meantime, it will appear in the middle of this list instead of after.

For an illustration see https://jsfiddle.net/39epgLj0/30/

This PR fixes this by using nextSibling of the previous node as a reference where available.

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -405,7 +405,15 @@ export function createPatchFunction (backend) {
       }
     }
     if (oldStartIdx > oldEndIdx) {
-      refElm = isUndef(newCh[newEndIdx + 1]) ? null : newCh[newEndIdx + 1].elm
+      if (isUndef(newCh[newEndIdx + 1])) {
+        if (isUndef(newCh[newEndIdx - 1]) || newCh[newEndIdx - 1].elm === undefined) {
+          refElm = null
+        } else {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR, but in general, Vue's patching algorithm assumes that the DOM structure will not be arbitrarily mutated by 3rd party code. It's simply impossible to cater to every possible case - for example, your patch only works when there is a single element appended, what if there are multiple ones appended? The code would have to bloat significantly to cater to all these cases and imo simply not practical nor worth it.

finntl:
Thanks for the reply - the patch does work regardless of the number of elements externally appended as it uses the elm from the vdom as the reference, but makes sense that the patching algorithm shouldn't be expected to deal with external mutations - noted!

Issue #5146, opened 2017-03-09
Incompatible event modifiers on keypress between 2.2.2 and 2.2.1


putc:
From version 2.2.2 to 2.2.1 it appears the logic for handling keypress events with a key-modifier and the prevent-modifier has changed.

In 2.2.1 an 'keypress.enter.prevent' would only prevent default action for the enter key, but in 2.2.2 it prevents every key.

I expected the behaviour to stay consistent with both releases, unless this change is intended.

Compare
2.2.1: https://jsfiddle.net/p4ay5v9c/2/
2.2.2: https://jsfiddle.net/z17jjxh6/1/

From my understanding the changelog for 2.2.2 does not list this change, so I reckon it is not intentional.

posva:
The change is not intentional, thanks for pointing it out. I'll take a look

yyx990803:
Fixed via #5147

sagalbot:
Will this be patched in 2.2.3?

PR #5147, opened 2017-03-09
fix wrong order of generate modifier code

Kingwl:
#5146

@posva 
review need


posva:
Hey, thanks. Can you add a test for stop too?

Kingwl:
@posva sure

Issue #5145, opened 2017-03-09
Warn the user when trying to use props in camelCase


CodinCat:
### Vue.js version
2.2.1

### Reproduction Link

http://codepen.io/CodinCat/pen/Qpdrjq?editors=1010

### What is Expected?

A warning.

### What is actually happening?

The like above doesn't work correctly (at least, not most users expected) because it is passing `aProp` in camelCase. Change the template code to `:a-prop` can fix this problem. 

But as you can see, there is no error message nor any warning. It just displays nothing. Sometimes this may make users spend hours to debug. Besides, [There are](http://stackoverflow.com/questions/41790596/vuejs-custom-properties-on-root-element) [a lot of](http://stackoverflow.com/questions/42087910/how-to-pass-object-vuejs) [questions](http://stackoverflow.com/questions/42689807/vue-passing-data-become-undefined) on Stackoverflow because the camelCase and kebab-case problem.

I have to say I make this mistake sometimes, because I switch between React and Vue. If Vue can show a warning for this case would be very helpful.

posva:
Unfortunately, we can't do anything about this because the Browser parses the HTML before and gives vue a lowercase version of the prop which doesn't exist.

CodinCat:
I believe this can be achieved. Consider the following snippet. When looping the propOptions:

```js
if (containsUppercase(propKey) && attrs.hasOwnProperty(propKey.toLowerCase()) {
  warn(...)
}
```

then the warning will show when user trying to pass a prop which is not in kebab-case, but declared in propOptions in camelCase. I'm not sure this code covers all the cases, it's just a small PoC.

posva:
In that case, feel free to submit a PR with a test for it 🙂 
It's clearly stated [on the guide](https://vuejs.org/v2/guide/components.html#camelCase-vs-kebab-case), that's why I think it's not necessary to add a warning. Running that check against all props in dev mode looks heavy imo

PR #5161, opened 2017-03-10
Warn when not passing props in kebab-case

CodinCat:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Hi,

Please see #5145 

This patch adds a warning when user is not passing props in kebab-case mistakenly, for example:

```html
<foo someProp="bar"></foo>
should be
<foo some-prop="bar"></foo>
```

I know this is stated on the guide, but since it's a very common mistake (based on the questions on stackoverflow (see the issue above) [and the](https://forum.vuejs.org/t/unable-to-pass-data-from-parent-to-child-component/1763/1) [forum](https://forum.vuejs.org/t/need-help-with-passing-a-boolean-prop-to-a-comonent/3996/1 )), I think it is worth to add a warning and I believe this will save a lot of people's debugging time and provide a more friendly development experience.

And for the performance, this is not a complex algorithm. It only adds a fews operations (both `toLowerCase` and `hasOwnProperty` are fast enough) in O(n) so I would say the impact on performance is negligible (almost no impact when running unit test on my machine). 

thanks!

### DIFF
--- a/src/core/vdom/create-component.js
+++ b/src/core/vdom/create-component.js
@@ -287,6 +287,17 @@ function extractProps (data: VNodeData, Ctor: Class<Component>): ?Object {
   if (attrs || props || domProps) {
     for (const key in propOptions) {
       const altKey = hyphenate(key)
+      const keyInLowerCase = key.toLowerCase()
+      if (
+        process.env.NODE_ENV !== 'production' &&
+        key !== keyInLowerCase &&
+        attrs && attrs.hasOwnProperty(keyInLowerCase)
+      ) {
+        warn(
+          `HTML attributes are case-insensitive. camelCased prop names need ` +
+          `to use their kebab-case equivalents. ${key} should be ${altKey}.`
+        )
+      }
       checkProp(res, props, key, altKey, true) ||
       checkProp(res, attrs, key, altKey) ||
       checkProp(res, domProps, key, altKey)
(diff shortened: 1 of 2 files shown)


CodinCat:
Um.. not sure why the CI failed. All the tests passed on my machine

src/core/vdom/create-component.js
```diff
@@ -287,6 +287,17 @@ function extractProps (data: VNodeData, Ctor: Class<Component>): ?Object {
   if (attrs || props || domProps) {
     for (const key in propOptions) {
       const altKey = hyphenate(key)
+      const keyInLowerCase = key.toLowerCase()
```
yyx990803: Can we move this inside the `NODE_ENV` conditional block as well? It's ok to use two nested ifs.

CodinCat: Sure, updated!

yamsellem:
What release should get this on board?

CodinCat:
@yamsellem Already released in [v2.2.3](https://github.com/vuejs/vue/releases/tag/v2.2.3)

yamsellem:
@CodinCat sorry 'bout that. Webpack was lower casing our html attributes (`caseSensitive` defaulting to false in [html-minifier](https://github.com/kangax/html-minifier)) — avoiding the warning to pop.

CodinCat:
If you are using single file components then it's not a problem. You camelCase props will be converted to kebab-case automatically.

PR #5216, opened 2017-03-18
Allow named properties on reactive arrays.

pkaminski:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Since arrays can be treated as normal objects in JavaScript and contain a mix of numeric and named properties, this patch makes `Vue.set` and `Vue.del` deal correctly with named properties on reactive arrays.

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -189,7 +189,7 @@ export function defineReactive (
  * already exist.
  */
 export function set (target: Array<any> | Object, key: any, val: any): any {
-  if (Array.isArray(target)) {
+  if (Array.isArray(target) && typeof key === 'number') {
     target.length = Math.max(target.length, key)
     target.splice(key, 1, val)
     return val
@@ -219,7 +219,7 @@ export function set (target: Array<any> | Object, key: any, val: any): any {
  * Delete a property and trigger change if necessary.
  */
 export function del (target: Array<any> | Object, key: any) {
-  if (Array.isArray(target)) {
+  if (Array.isArray(target) && typeof key === 'number') {
     target.splice(key, 1)
     return
   }
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -198,7 +198,7 @@ export function set (target: Array<any> | Object, key: any, val: any): any {
     target[key] = val
     return val
   }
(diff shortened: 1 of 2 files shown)


src/core/observer/index.js
```diff
@@ -219,11 +219,11 @@ export function set (target: Array<any> | Object, key: any, val: any): any {
  * Delete a property and trigger change if necessary.
  */
 export function del (target: Array<any> | Object, key: any) {
-  if (Array.isArray(target)) {
+  if (Array.isArray(target) && typeof key === 'number') {
     target.splice(key, 1)
     return
   }
-  const ob = target.__ob__
+  const ob = (target : any).__ob__
```
znck: I can't understand this change. Could you explain?

pkaminski: I can't say I understand it either, but without this cast Flow complains that "Array doesn't have an
__ob__ property" -- why it decided that `target` is an array, I don't know.  It appears to be an instance of [this Flow issue](https://github.com/facebook/flow/issues/1330), though my knowledge of Flow is minimal so I can't tell for sure.

yyx990803:
I'm not sure if this is a good idea - custom properties on Arrays are ignored during JSON.stringify and cannot be persisted. I personally would never rely on them.

pkaminski:
I'm confused -- what do reactive properties have to do with persistence?  I certainly don't do this often myself, but it's occasionally useful to pass metadata through an API/SPI where you control both ends but can only transmit an array through the library.  IMHO the principle that you can safely replace any member assignment with a call to `Vue.set` outweighs trying to "fix" iffy JavaScript semantics.

yyx990803:
@pkaminski yeah, I guess it doesn't matter that much. I'm just thinking about the implications this has to do with normal persistence scenarios and in particular Vuex snapshots.

pkaminski:
Honestly, I don't even care if the resulting named array property is reactive (though I think it should be for consistency), I'd just like `Vue.set` not to crash when invoked with an array target and a string key.

pkaminski:
Thanks!

PR #5267, opened 2017-03-23
Warn and handle missing get in computed (fix #5265)

### LINKED ISSUE #5265: Computed with set but no get causes exception
### Version
2.2.4

### Reproduction link
[http://codepen.io/anon/pen/gmKpNj?editors=1111](http://codepen.io/anon/pen/gmKpNj?editors=1111)

### Steps to reproduce
Create a component with a computed which has a setter but does not have a getter. In my example, you get an exception in the console and the page does not render

### What is expected?
Preferably a better error message or warning from Vue rather than an uncaught exception. 

Or for backwards compatibility with Vue 2.1.8, have Vue allow a component to have a computed with no getter.

Receiving an console error and vue does not render anything.

Uncaught TypeError: Cannot read property 'toString' of undefined
    at new Watcher (vue.js:2417)
    at initComputed (vue.js:2743)
    at initState (vue.js:2652)
    at VueComponent.Vue._init (vue.js:3696)
    at new VueComponent (vue.js:3862)
    at createComponentInstanceForVnode (vue.js:3080)
    at init (vue.js:2888)
    at createComponent (vue.js:4579)
    at createElm (vue.js:4522)
    at createChildren (vue.js:4647)

---
This did not cause any exceptions in vue 2.1.8 and the component rendered and worked correctly.

This is code written by my coworker, and after upgrading vue today, this page was broken. I don't understand why it was written the way it was, and I do not think that the get is ever used, while the set is. It seems nonsensical to me to have code like this, but it exists. The problem is it took quite a while to hunt down why this was happening. 

I can imagine a potential scenario where a developer misspells "get" and receives this error without any warning why, and is similarly confused on how to fix it.

kmc059000:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

I ran all test locally. I'm having some failures related to SSR and selenium which appear to be completely unrelated. I've confirmed the new test works by reverting my change and adding it back in.

I am using `noop` for the getter if one does not exist to avoid exceptions when calling functions on the getter.

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -147,7 +147,16 @@ function initComputed (vm: Component, computed: Object) {
 
   for (const key in computed) {
     const userDef = computed[key]
-    const getter = typeof userDef === 'function' ? userDef : userDef.get
+    let getter = typeof userDef === 'function' ? userDef : userDef.get
+    if (process.env.NODE_ENV !== 'production') {
+      if (getter === undefined) {
+        warn(
+          `No getter function has been defined for computed property "${key}".`,
+          vm
+        )
+        getter = noop
+      }
+    }
     // create internal watcher for the computed property.
     watchers[key] = new Watcher(vm, getter, noop, computedWatcherOptions)
 
--- a/test/unit/features/options/computed.spec.js
+++ b/test/unit/features/options/computed.spec.js
@@ -48,6 +48,33 @@ describe('Options computed', () => {
     }).then(done)
   })
 
+  it('warn with setter and no getter', () => {
+    const vm = new Vue({
+      template: `
+        <div>
+          <test></test>
+        </div>
+      `,
+      components: {
+        test: {
+          data () {
+            return {
+              a: 1
+            }
+          },
+          computed: {
+            b: {
+              set (v) { this.a = v }
+            }
+          },
+          template: `<div>{{a}}</div>`
+        }
+      }
+    }).$mount()
+    expect(vm.$el.innerHTML).toBe('<div>1</div>')
+    expect('No getter function has been defined for computed property "b".').toHaveBeenWarned()
+  })
+
   it('watching computed', done => {
     const spy = jasmine.createSpy('watch computed')
     const vm = new Vue({


kmc059000:
#5265

src/core/instance/state.js
```diff
@@ -147,7 +147,15 @@ function initComputed (vm: Component, computed: Object) {
 
   for (const key in computed) {
     const userDef = computed[key]
-    const getter = typeof userDef === 'function' ? userDef : userDef.get
+    let getter = typeof userDef === 'function' ? userDef : userDef.get
+    if (getter === undefined) {
+      warn(
+        `computed "${key}" is requires a "get" to be defined if a "set" is defined.` +
```
rayrutjes: What about saying: "No getter function has been defined for computed property `${key}`."

src/core/instance/state.js
```diff
@@ -147,7 +147,15 @@ function initComputed (vm: Component, computed: Object) {
...
+        `Did you define or misspell the get?`,
```
rayrutjes: I don't think we need this.

src/core/instance/state.js
```diff
@@ -147,7 +147,15 @@ function initComputed (vm: Component, computed: Object) {
 
   for (const key in computed) {
     const userDef = computed[key]
-    const getter = typeof userDef === 'function' ? userDef : userDef.get
+    let getter = typeof userDef === 'function' ? userDef : userDef.get
+    if (getter === undefined) {
```
posva: Can you wrap the if in a non-prod if please? (`if (process.env.NODE_ENV !== 'production')`)

src/core/instance/state.js
kmc059000: Will do.

src/core/instance/state.js
kmc059000: I'll remove.

src/core/instance/state.js
kmc059000: Yep.

kmc059000: Should only the warning be wrapped? I think getter should be assigned to noop regardless of whether it is in production mode or not, so that the behavior is consistent.

posva: I don't think it's necessary to have a fallback on production because we'd get the error at runtime, so wrapping the whole if looks better imo. It'll also prevent an unnecessary check for all computed properties

src/core/instance/state.js
```diff
@@ -147,7 +147,16 @@ function initComputed (vm: Component, computed: Object) {
 
   for (const key in computed) {
     const userDef = computed[key]
-    const getter = typeof userDef === 'function' ? userDef : userDef.get
+    let getter = typeof userDef === 'function' ? userDef : userDef.get
+    if (process.env.NODE_ENV !== 'production') {
```
xujiongbo:
```js
if(process.env.NODE_ENV !== 'production' && !getter) {
  // warning
}
```
might be better

kmc059000:
[Here](https://github.com/vuejs/vue/commit/406352baba459288473915218b0ee81cf8066898#diff-cb87aa7890fa694dd923ea9f162256d2L137) is the change that changed the behavior. Notice on this line that `noop` is used if `get` is falsy. Does anybody think that this original behavior is preferable to the warning?

```
const getter = typeof userDef === 'function' ? userDef : (userDef.get || noop)
```

This would be consistent with how [`defineComputed()`](https://github.com/vuejs/vue/blob/0bb529a275072b9a99f645338a8c1f13439caa96/src/core/instance/state.js#L172) is defined here when the `get` is undefined. 

IMO this is better because it would not cause errors in production mode, avoid a new warning, keep the 2.1.* behavior, and make the behavior between production and non-production modes consistent.

src/core/instance/state.js
znck: I don't think so. Former is better as define plugin would replace `process.env.NODE_ENV` with `'production'` and it can be stripped off.

posva: both are stripped of actually

PR #5303, opened 2017-03-28
Update grid.js

TilakMaddy:
Double arrow functions for easier reading

### DIFF
--- a/examples/grid/grid.js
+++ b/examples/grid/grid.js
@@ -9,7 +9,7 @@ Vue.component('demo-grid', {
   },
   data: function () {
     var sortOrders = {}
-    this.columns.forEach(function (key) {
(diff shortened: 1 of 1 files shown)


posva:
Hey, thanks for your interest. It looks like you still need to learn Vue and arrow functions a bit more 😛 . The arrows functions are binding the context, therefore the `this` changes. This is important for computed properties. Also, with arrows function you can omit the return in one-liners:

```js
(param) => param.toUpperCase()
// equivalent to
(param) => {
  return param.toUpperCase()
}
```

About the examples, they're kept simple on purpose, to make it easy to understand them for any js developer, even if they still don't know about arrow functions. Cheers

TilakMaddy:
Oops... Sorry

Issue #4876, opened 2017-02-07
Warning when overriding native directives?


texelate:
In Vue 1.0 with `.config.debug` set to `true` I just spent ages trying to find out why `v-html` was causing errors and finally realised it was because I had a custom directive v-html for a WYSIWYG editor.

Would it be a good idea to issue a warning when this happens?

Thanks. 

Amazing work, by the way. I _love_ Vue!

gebilaoxiong:
@yyx990803 
Can I submit a PR?

When the constructor `directives ()` && `extend` is called and the time of initialization

Check whether the directive is a built-in directive

yyx990803:
@gebilaoxiong the check should probably be done in `core/util/options` when merging options.

gebilaoxiong:
@yyx990803 Yes, but this time options may be built-in options

yyx990803:
@gebilaoxiong you can just import the built-in options and check during merge (also, only need to check child value)

gebilaoxiong:
@yyx990803 understand, thank you very much .

yyx990803:
See https://github.com/vuejs/vue/pull/5328#issuecomment-291042541

PR #5328, opened 2017-03-30
warn when overwrite built-in directives (#4876)

gebilaoxiong:
https://github.com/vuejs/vue/issues/4876

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -14,6 +14,7 @@ export type Config = {
   keyCodes: { [key: string]: number | Array<number> };
   // platform
   isReservedTag: (x?: string) => boolean;
+  isBuitInDirective: (x?: string) => boolean;
   parsePlatformTagName: (x: string) => string;
   isUnknownElement: (x?: string) => boolean;
   getTagNamespace: (x?: string) => string | void;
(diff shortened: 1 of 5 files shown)


dsonet:
IMO it's not necessary to detect this. Kinda like to detect if someone overwrite the Vue instance.
And if really really needs, should only in development mode.

gebilaoxiong:
@dsonet thanks for you reply.

https://vuejs.org/v2/guide/custom-directive.html#Intro

> In addition to the default set of directives shipped in core (v-model and v-show), Vue also allows you to register your own custom directives

Like `checkComponent()`, I was checking in development mode

yyx990803:
Hmm thanks for the PR - bt this doesn't seem to detect cases where user attempts to register other built-ins e.g. `v-if` (which would just fail silently).

After thinking about it, it seems the only way would be to have a constant list of built-ins for each platform. Feels too much of an edge case for this extra maintenance burden.

Still, thank you for the effort put into this PR!

gebilaoxiong:
Yes, it is also bothering me, in addition also need to distinguish between compile-time and runtime

PR #5622, opened 2017-05-06
only set a key to VNode have similar shape, fix #5618

### LINKED ISSUE #5618: Elements unnecessarily re-created when an above v-for changes.
### Version
2.3.2

### Reproduction link
[https://jsfiddle.net/gazugafan/a8kverza/2/](https://jsfiddle.net/gazugafan/a8kverza/2/)

### Steps to reproduce
Change the inputs marked "Change Me", then add a row. Any inputs following the v-for rows get re-created (and lose their value), while inputs above remain as they are. This only seems to happen when the inputs and v-for element are contained in a parent component's slot.

### What is expected?
All inputs remain unchanged--regardless of what happens to the sibling v-for element.

Any input elements following the v-for are re-created--resetting their value.

pengchongfu:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

in issue #5618 , then VNode in the first layer of the array don't get a key when father component call `normalizeChildren`,
but get a key when child component call `normalizeChildren`.

this leads to a changeable key, and component was re-created.

### DIFF
--- a/src/core/vdom/helpers/normalize-children.js
+++ b/src/core/vdom/helpers/normalize-children.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import VNode, { createTextVNode } from 'core/vdom/vnode'
-import { isDef, isUndef, isPrimitive } from 'shared/util'
+import { isDef, isUndef, isPrimitive, isObject } from 'shared/util'
 
 // The template compiler attempts to minimize the need for normalization by
 // statically analyzing the template at compile time.
@@ -45,7 +45,11 @@ function normalizeArrayChildren (children: any, nestedIndex?: string): Array<VNo
     last = res[res.length - 1]
     //  nested
     if (Array.isArray(c)) {
-      res.push.apply(res, normalizeArrayChildren(c, `${nestedIndex || ''}_${i}`))
+      if (hasNestedIndex(c)) {
+        res.push.apply(res, normalizeArrayChildren(c, `${nestedIndex || ''}_${i}`))
+      } else {
+        res.push.apply(res, normalizeArrayChildren(c))
+      }
     } else if (isPrimitive(c)) {
       if (isDef(last) && isDef(last.text)) {
         last.text += String(c)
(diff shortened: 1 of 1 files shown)


pengchongfu:
test still failed, I am try to find the reason.

gebilaoxiong:
@pengchongfu 

You can compile the code, and then copy the test case to the console debugging

Hope to help you : )

pengchongfu:
@gebilaoxiong , thanks. 

- My change seems to introduce a bug : (. I hava test locally and found this [test](https://github.com/vuejs/vue/blob/dev/test/unit/modules/vdom/patch/edge-cases.spec.js#L61) failed.

- I make a [demo](https://jsfiddle.net/6zvyzzsh/), I know it is a little hack. `compVm._watcher` was inactive, I am try to find the reason.

pengchongfu:
@yyx990803 , my change would introduce new bug :(, because if we force every VNode has a key by order, this will have side effect for patch, see this [demo](https://jsfiddle.net/6zvyzzsh/2/).

I have update my change, maybe we should filter the VNode before set a key.
only set a key to VNode have similar shape, likely generated by v-for.

src/core/vdom/helpers/normalize-children.js
```diff
@@ -45,7 +45,11 @@ function normalizeArrayChildren (children: any, nestedIndex?: string): Array<VNo
     last = res[res.length - 1]
     //  nested
     if (Array.isArray(c)) {
-      res.push.apply(res, normalizeArrayChildren(c, `${nestedIndex || ''}_${i}`))
```
gebilaoxiong: I think here only need to determine whether `c` is created by `slot`, avoid to set `key ` twice

gebilaoxiong: let me try

gebilaoxiong:
There are some problems with nested child elements as `v-for`

``` html
  <div id="app"></div>
```

``` javascript
const vm = new Vue({
  el: '#app',
  data: { n: 1 },
  render(h) {
    const list = []

    for (let i = 0; i < this.n; i++) {
      list.push(h('input', {
        attrs: {
          value: 'a',
          type: 'text'
        }
      }))
    }

    list.push(h('input', {
      attrs: {
        value: 'b',
        type: 'text'
      }
    }))
    
    return h('div', [list])
  }
})
```

If the nested list is followed by the same type of element

When the data changes

it will be the last element as `v-for` created

In the above example he will generate a new input, and the value is b

pengchongfu:
@gebilaoxiong , you are right, the best way is coding carefully...

gebilaoxiong:
@pengchongfu You can leave a contact, my QQ is 173457117

PR #5627, opened 2017-05-07
Fix: nested child elements can not be updated correctly, fix #5618

### LINKED ISSUE #5618: Elements unnecessarily re-created when an above v-for changes.

gebilaoxiong:
https://github.com/vuejs/vue/issues/5618

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


gebilaoxiong:
I think e2e shows the error is not what I am causing

gebilaoxiong:
Well, it's an interesting thing to name the title 🤣

gebilaoxiong:
Wait a moment, the test code seems to have some problems

src/core/vdom/helpers/normalize-children.js
```diff
@@ -45,7 +45,9 @@ function normalizeArrayChildren (children: any, nestedIndex?: string): Array<VNo
     last = res[res.length - 1]
     //  nested
     if (Array.isArray(c)) {
-      res.push.apply(res, normalizeArrayChildren(c, `${nestedIndex || ''}_${i}`))
+      // avoid elements to be set key, which created by slot
+      const subNestedIndex = isTrue(c._rendered) ? undefined : `${nestedIndex || ''}_${i}`
```
pengchongfu: `_renndered` is set when `process.env.NODE_ENV !== 'production'`.

and not only slot, hand-written render functions could have this problem too.

gebilaoxiong: @pengchongfu 

I did not find any problems in the ` render function`

html:
```html
 <div id="app"></div>
```
javascript: 
``` javascript
var vm = new Vue({
  el: '#app',
  data: {
    count: 1
  },
  render: function (h) {
    var list = []

    for (var i = 0; i < this.count; i++) {
      list.push(h('span', [this.count]))
    }

    return h('div', [
      list,
      h('input', {attrs: { 
        value: 'a',
        type: 'text'
      }})
    ])
  }
})
```
You can input something in the input box,
then modify the value of `count` in the console

and input`s value did not changed

gebilaoxiong: @pengchongfu 

Can you give me a render example of `hand-written render functions`? 🌹

pengchongfu: @gebilaoxiong , [demo](https://jsfiddle.net/Lzxnmn9p/), hack... maybe we should ignore it 😂

and `_rendered` is not set in production, see this [line](https://github.com/vuejs/vue/blob/dev/src/core/instance/render-helpers/render-slot.js#L24)

gebilaoxiong: @pengchongfu 

💋  Thanks , I'll take a look at this question

gebilaoxiong:
@pengchongfu 

Now when using render function does not exist this problem

I'm going to add a test case : )

pengchongfu:
@gebilaoxiong , but all array in hand-written render function will not get a key, I don't know whether it is a good solution.

gebilaoxiong:
@pengchongfu 

I think this happens only when there are nested arrays in children

just like:

```javascript
children: [
  [vnode, vnode...]
]
```

In the `render` of the template, it will be treated as `v-for`

but `slot` and` render function` will also generate this structure

so we just skip to set the key

gebilaoxiong:
@pengchongfu wait, I have a better way to solve this problem！

gebilaoxiong:
@pengchongfu 

I marked the `v-for` generated elements

Only v-for generated elements will be set to `key`

avoiding `render function` and` slot` to generate the same structure when an error occurs

gebilaoxiong:
@yyx990803 

Well, it looks like I'm the last one waiting for today

Do not close my PR, tell me this is not the best way 😄

src/core/instance/render-helpers/render-list.js
```diff
@@ -9,23 +9,29 @@ export function renderList (
...
     keys = Object.keys(val)
     ret = new Array(keys.length)
     for (i = 0, l = keys.length; i < l; i++) {
       key = keys[i]
-      ret[i] = render(val[key], key, i)
+      vnode = render(val[key], key, i)
+      vnode.isListItem = true
+      ret[i] = vnode
     }
   }
```
yyx990803: Instead of adding a property to every vnode, we can simply mark the `ret` array here with something like:

``` js
if (isDef(ret)) {
  (ret: any)._isVList = true // work around flow
}
```

Then check `children._isVList` in `normalizeArrayChildren`.

gebilaoxiong: Yes, I also feel that there are some problems, very grateful

gebilaoxiong:
@yyx990803 

Fix bug is really a very interesting thing

Maybe my level is not enough, just think you need to chew and then get the truth

yyx990803:
I pushed a commit to ensure `nestedIndex` is always passed down for nested arrays. This is necessary in cases where there are two `v-for` lists nested under non `v-for` arrays, e.g.

``` js
[[[v-for list], ..., [[v-for list]]]
```

gebilaoxiong:
understand, thank you very much

Issue #5605, opened 2017-05-04
Warn when component option should be an object, but is a function


phoenixeliot:
### What problem does this feature solve?
I often get mixed up on which properties of a component should be functions and which shouldn't, and mistakenly define some as functions when they should have been objects.

In the opposite scenario, when a property that needs to be a function is instead an object, we already get a very helpful warning message:
`[Vue warn]: The "data" option should be a function that returns a per-instance value in component definitions. `

But if you define the `watch`, `methods`, `created`, or `computed` properties (and probably some others) as functions instead, Vue either silently fails to use the property at all, or gives a confusing error message. I've spent an embarrassing amount of my Vue dev time because of this simple mistake.

Here's what those ones do if the value is a function instead of an object:

`watch` will simply be ignored if it's a function instead of an object.

`components` will prevent any children from rendering and log this message: `Unknown custom element: <child-component> - did you register the component correctly? For recursive components, make sure to provide the "name" option. (found in root instance)`

`methods` will cause a runtime error with this message: `[Vue warn]: Property or method "someMethod" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option. (found in root instance)`

`computed` — similar error `methods`

Most of these errors are at least somewhat misleading as to what the real problem is.

The main benefit of a warning for this kind of mistake would be to spend less time futzing around with fixing the boilerplate after getting it slightly wrong.

### What does the proposed API look like?
If a component is initialized with any properties whose values are functions, but should be objects, emit a warning similar to the warning for `data`. eg:

`[Vue warn]: The "computed" option should be an object, not a function, in component definitions.`

posva:
There's nothing to be embarrassed by 🙂 
What interests me more is why did you think using an object instead of a function or vice versa should have worked in any of those properties. Dev warnings are cool but they do have a cost to maintain and at execution too. I thought the docs covered that quite well

posva:
ping @phoenixeliot

phoenixeliot:
It's less a lack of understanding and more going through the boilerplate a little too fast, and then later on when actually using the options, things fail in ways that aren't clear.

When making a new component, I'll just throw in the common options to start with, then fill them out later, just to take the mental overhead out of remembering exactly what they're called. (We haven't been using Vue for terribly long, so I don't have them down by heart yet).

Also, side note: We use CoffeeScript. But I think in ES6 notation it could look pretty similar.

So I'll throw together a new component and it'll look something like this (when done right):
```
[...]

PR #5642, opened 2017-05-09
Warn when component option should be an object, but is not (#5605)

lokson:
https://github.com/vuejs/vue/issues/5605

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -60,16 +60,14 @@ const isReservedProp = {
   slot: 1
 }
 
-function parseObjectOption (vm: Component, name: string) {
+function checkOptionType (vm: Component, name: string) {
   const option = vm.$options[name]
   if (!isPlainObject(option)) {
-    process.env.NODE_ENV !== 'production' && warn(
+    warn(
       `${name} should be an object.`,
       vm
     )
-    return {}
   }
-  return option
 }
 
 function initProps (vm: Component, propsOptions: Object) {
@@ -160,7 +158,7 @@ function getData (data: Function, vm: Component): any {
 const computedWatcherOptions = { lazy: true }
 
 function initComputed (vm: Component, computed: Object) {
-  computed = parseObjectOption(vm, 'computed')
+  process.env.NODE_ENV !== 'production' && checkOptionType(vm, 'computed')
   const watchers = vm._computedWatchers = Object.create(null)
   for (const key in computed) {
     const userDef = computed[key]
@@ -225,7 +223,7 @@ function createComputedGetter (key) {
 }
 
 function initMethods (vm: Component, methods: Object) {
-  methods = parseObjectOption(vm, 'methods')
+  process.env.NODE_ENV !== 'production' && checkOptionType(vm, 'methods')
   const props = vm.$options.props
   for (const key in methods) {
     vm[key] = methods[key] == null ? noop : bind(methods[key], vm)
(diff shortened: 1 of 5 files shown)


src/core/instance/state.js
```diff
@@ -148,8 +148,14 @@ function getData (data: Function, vm: Component): any {
 const computedWatcherOptions = { lazy: true }
 
 function initComputed (vm: Component, computed: Object) {
+  if (!isPlainObject(computed)) {
```
posva: You should refactor this block into a function and add the check there too so it is stripped when bundling:

`process.env.NODE_ENV !== 'production' && checkOptionType(vm, 'computed')`

test/unit/features/options/methods.spec.js
```diff
@@ -24,4 +24,11 @@ describe('Options methods', () => {
     })
     expect(`method "hello" has an undefined value in the component definition`).toHaveBeenWarned()
   })
+
+  it('should warn non object', () => {
```
posva: These tests can also be refactored into a function. It should also test it doesn't print the warning in the right scenario

lokson: Should refactored function be a separate file in `test/helpers/`?

posva: yes, after all you need to include it in different test files 🙂

src/core/instance/state.js
```diff
@@ -60,6 +60,17 @@ const isReservedProp = {
   slot: 1
 }
 
+function parseObjectOption (vm: Component, option: Object, name: string) {
```
posva: no need for option arg because you have name

src/core/instance/state.js
```diff
@@ -148,13 +159,7 @@ function getData (data: Function, vm: Component): any {
...
 
 function initComputed (vm: Component, computed: Object) {
-  if (!isPlainObject(computed)) {
-    computed = {}
-    process.env.NODE_ENV !== 'production' && warn(
-      'computed should be an object.',
-      vm
-    )
-  }
+  computed = parseObjectOption(vm, computed, 'computed')
```
posva: This must be called only in non-production env (NODE_ENV !== 'production')

src/core/instance/state.js
```diff
@@ -60,6 +60,17 @@ const isReservedProp = {
   slot: 1
 }
 
+function parseObjectOption (vm: Component, option: Object, name: string) {
+  if (!isPlainObject(option)) {
+    process.env.NODE_ENV !== 'production' && warn(
```
posva: since this function is only called in dev mode, this check is unecessary

src/core/instance/state.js
posva: better call it checkOptionType

lokson: the result is used in assignment, to call it a "check" may be misleading

src/core/instance/state.js
lokson: it may be good to provide a fallback (empty object) if a wrong type is used in prod mode

posva: no, because it will never reach production code

posva: The check must be removed in production

src/core/instance/state.js
posva: The assignment prevents runtime errors but the function is meant to do a check

lokson:
renamed to `checkOptionType` and guarded prod mode

src/core/instance/state.js
```diff
@@ -60,6 +60,16 @@ const isReservedProp = {
   slot: 1
 }
 
+function checkOptionType (vm: Component, name: string) {
+  const option = vm.$options[name]
+  if (!isPlainObject(option)) {
+    warn(
+      `${name} should be an object.`,
```
yyx990803: The message could be a bit more explicit: `component option "${name}" should be an object`

PR #5653, opened 2017-05-11
Added check in merge strat for watches if child is already an array (fix #5652)

### LINKED ISSUE #5652: Support using result of Vue.extend in mixins
### Version
2.3.3

### Reproduction link
[http://jsfiddle.net/zna06rq4/1/](http://jsfiddle.net/zna06rq4/1/)

### Steps to reproduce
Run fiddle, error will be shown in console:

```
Error in callback for watcher "hello": "TypeError: this.cb.call is not a function"
```

### What is expected?
"child" should be logged to the console via the watch handler in Child.

The watch is not being fired

---
In strats.watch (https://github.com/vuejs/vue/blob/086ae4469b0200e94b1ad0f70771e57c63ce18af/src/core/util/options.js#L175) the function unconditionally converts child to an array.
If this has already been converted to an array (via a previous merge operation) it will convert it to a nested array `[[child]]`.

The fix for this seems to be updating line 175 to have:
`Array.isArray(child) ? child : [child]`

strantr:
When performing multiple extends/mixins watches are converted to arrays of handlers.
If the parent does not watch an element, but the children do, it will not run the parent.concat code and instead always wrap the child.
If the child is already an array it does not need to be wrapped again.
Fixing #5652

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -172,7 +172,7 @@ strats.watch = function (parentVal: ?Object, childVal: ?Object): ?Object {
     }
     ret[key] = parent
       ? parent.concat(child)
-      : [child]
+      : Array.isArray(child) ? child : [child]
   }
   return ret
 }
--- a/test/unit/features/options/extends.spec.js
+++ b/test/unit/features/options/extends.spec.js
@@ -46,4 +46,41 @@ describe('Options extends', () => {
     expect(vm.b).toBe(2)
     expect(vm.c).toBe(3)
   })
+
+  it('should work with global mixins and Object.prototype.watch polyfill', done => {
+    if (!Object.prototype.watch) {
+      // eslint-disable-next-line no-extend-native
+      Object.prototype.watch = { remove: true }
+    }
+
+    Vue.use({
+      install: function () {
+        Vue.mixin({})
+      }
+    })
+    const spy = jasmine.createSpy('watch')
+    const A = Vue.extend({
+      data: function () {
+        return { a: 1 }
+      },
+      watch: {
+        a: spy
+      },
+      created: function () {
+        this.a = 2
+      }
+    })
+    new Vue({
+      extends: A
+    })
+    waitForUpdate(() => {
+      delete Object.prototype.watch
+      expect(spy).toHaveBeenCalledWith(2, 1)
+    }).then(done)
+  })
+  afterEach(function () {
+    if (Object.prototype.watch && Object.prototype.watch.remove) {
+      delete Object.prototype.watch
+    }
+  })
 })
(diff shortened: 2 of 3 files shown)


posva:
Could you add a test for this, please?

posva:
Sorry, I checked the PR before the issue, but I'm not sure there's a problem here, mixins are pain javascript objects, not Vue sub classes

dsonet:
@posva  It's very reasonable to add this check, to avoid wrap the child value accidentally.

posva:
@dsonet Yes, but we have to make sure the _accident_ is possible 😄

strantr:
@posva I've added the unit test based on the existence of Object.prototype.watch

test/unit/features/options/extends.spec.js
```diff
@@ -46,4 +46,41 @@ describe('Options extends', () => {
...
+    })
+    new Vue({
+      extends: A
+    })
+    waitForUpdate(() => {
+      delete Object.prototype.watch
+      expect(spy).toHaveBeenCalledWith(2, 1)
+    }).then(done)
+  })
+  afterEach(function () {
```
posva: I think we could group this test into a describe and use `afterAll` so the check is run only once. You can also setup the `watch` property on the `beforeAll`

strantr: Made the suggested change

strantr:
@posva Any idea why the CI test would fail on "Transition Basic: explicit transition type" - works fine locally and has no relation to my change

src/core/util/options.js
```diff
@@ -172,7 +172,7 @@ strats.watch = function (parentVal: ?Object, childVal: ?Object): ?Object {
     }
     ret[key] = parent
       ? parent.concat(child)
-      : [child]
+      : Array.isArray(child) ? child : [child]
```
yyx990803: So I finally took a proper look at this issue - since the error is caused by native `watch` on `Object.prototype`  (which is a function), I think the proper fix would be adding the following at the top of the function:

``` js
if (typeof parentVal === 'function') {
  parentVal = null
}
if (typeof childVal === 'function') {
  childVal = null
}
```

Essentially, ignore if the value is `Object.prototype.watch`. The test case will need to be updated accordingly.

strantr: @yyx990803 I've added a new test for merging multiple layers of extends (I wasn't sure if it was more appropriate in the watch tests or the extends tests) - this causes the issue where child is wrapped multiple times. I am happy to make your change however this second test case will fail with your recommendation but work with the initial fix.

dsonet: IMO the initial fix was correct.The core reason caused bug was because it deal with the child value inconsistently. With the fix it can deal with it correctly.

Issue #5628, opened 2017-05-07
Improve error reporting/debugging of provide/inject, or allow require flag for inject


johnhargrove:
### What problem does this feature solve?
Currently when an `inject` fails to resolve, nothing happens, and the injected value just remains untouched. This seems like an opportunity to warn the developer that an `inject` could not be resolved.

Perhaps this is intended behavior. Maybe it makes sense that sometimes an injection would not be resolved. If that is the case, could there be a way to indicate a given inject is required? A stricter injection method?

My current workaround is to throw an error if the object is undefined as of `create` (since injects are resolved by then), but it's a manual step that must be made in many places.

### What does the proposed API look like?
If a new API is needed to support this, I am not sure how to design it.

As for warning, I was hoping that when an `inject` could not be resolved, there would be a warning in the browser console.

lokson:
@johnhargrove could you give an example of failing `inject`? I'm not sure when can this happen.

johnhargrove:
@lokson This is any situation in which a child component has an `inject` which can't be satisfied by a `provide` in itself or any of its component ancestors, including the root Vue object itself.

I ran into this while developing a validation service. Basically, I have a `ValidationService` class which is a `provide` on a top level component which represents a collection of form inputs. The inputs themselves are very complicated, so they are their own Vue components:

```
 FormComponent (provides ValidationService)
 |-- ComplicatedInput (expects ValidationService inject)
 |-- ComplicatedInput (expects ValidationService inject)
```
If, when building this, I forget to `provide` the ValidationService on the parent `FormComponent`, the `ComplicatedInput` components will silently never have their reference to `ValidationService` injected. This results in errors related to accessing members of an undefined object, etc. 

My theory here and the reason for this feature request is that getting some kind of indication that an `inject` went unsatisfied seems useful for development.

Anamasi30:
anamasi30@gmail.com

PR #5681, opened 2017-05-16
Warn when a inject has been not provided

maciej-ka:
https://github.com/vuejs/vue/issues/5628

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


src/core/instance/inject.js
```diff
@@ -57,6 +57,8 @@ export function resolveInject (inject: any, vm: Component): ?Object {
         }
         source = source.$parent
       }
+      process.env.NODE_ENV !== 'production' && typeof result[key] === 'undefined' &&
```
lonelyclick: maybe `result[key]` just is `undefined`

`hasOwn(result, key)`

src/core/instance/inject.js
```diff
@@ -57,6 +57,8 @@ export function resolveInject (inject: any, vm: Component): ?Object {
...
+        warn(`injection "${key}" not found`, vm)
```
lonelyclick: capitalize `Injection "${key}" not found`

src/core/instance/inject.js
lokson: Thanks! I added a test case and a solution using`Object.keys`.

In the end solution is not based on `result.hasOwnProperty(key)` or a `result.hasOwnProperty && !result.hasOwnProperty(key)` because they failed some tests, not sure why. Also Flow was reporting strange error, which required a change in var definition to an odd: `const result = (Object.create(null): Object)`.

lokson:
@lonelyclick there seems to be no consistency to start short warnings uppercase, some examples: 
'infinite update loop'
'duplicate attribute'
'v-else use on element'
... and much more

This seems like a very low priority but we could change them to uppercase too. Would you like this, @posva?

src/core/instance/inject.js
lonelyclick: `shared/util/hasOwn`

lokson:
@lonelyclick thanks!

src/core/instance/inject.js
```diff
@@ -3,6 +3,7 @@
 import { hasSymbol } from 'core/util/env'
 import { warn } from '../util/index'
 import { defineReactive } from '../observer/index'
+import { hasOwn } from '../../shared/util'
```
yyx990803: can just use `shared/util` instead because it's aliased

src/core/instance/inject.js
```diff
@@ -57,6 +58,8 @@ export function resolveInject (inject: any, vm: Component): ?Object {
         }
         source = source.$parent
       }
+      process.env.NODE_ENV !== 'production' && !hasOwn(result, key) &&
```
yyx990803: let's use `if` here for more than one conditions

Issue #5624, opened 2017-05-07
Expose an option to change contentPlaceholder value in createBundleRenderer function


vanesyan:
### What problem does this feature solve?
Make it more flexible it terms of SSR, allows developers decide on placeholder name, rather strictly dictate which one to use (i.e. `<!--vue-ssr-outlet-->`).

### What does the proposed API look like?
Assume following is `options` object for `createBundleRender` function
```
{
  ...
  contentPlaceholder: '{{{ app }}}'
}
```

posva:
why would it make it more flexible?

vanesyan:
For example, if I build framework something like next.js, I'd like to use layout with custom vue application entry name differ from 'vue-ssr-outlet' to make it a bit abstract out of implementation details.

dsonet:
Would add extra cost to learn but gain barely benefit.

PR #5690, opened 2017-05-17
feat: add ssr template contentPlaceholder

lonelyclick:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
#5624

### DIFF
--- a/src/server/create-renderer.js
+++ b/src/server/create-renderer.js
@@ -28,6 +28,7 @@ export type RenderOptions = {
   shouldPreload?: Function;
   clientManifest?: ClientManifest;
   runInNewContext?: boolean | 'once';
+  contentPlaceholder?: string;
 };
 
 export function createRenderer ({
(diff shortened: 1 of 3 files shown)


dsonet:
I don't think should introduce this since it come with barely benefit.

yyx990803:
Thanks for the PR, but I don't really think this is particularly useful. I'd rather have templates in all Vue SSR projects be consistent.

JounQin:
Why not? What if we only need SSR for some specific urls, what means we need a temple for non-ssr requests. For now we must create two temples, or replace `<div id="app"></div>` to SSR default contentPlaceholder.

And also, there is an option of contentPlaceholder before 2.3.0, so it actually breaks some use cases which is not mentioned in SSR guide.

yyx990803:
@JounQin not sure if I understand your use case - you mean you want to set the placeholder to something else so that the route does not SSR?

The previous option was in a separate package so technically it was not under Vue's version constraints.

JounQin:
Yes, we shared same template for SSR and non-SSR requests before (our placeholder is `<div id="app"></app>` which is a valid root for non-SSR requests).

To developers, we're just migrating from one solution to another, and they share most of developing experience except `contentPlaceholder` option, so maybe it will be better if it's mentioned in document.

yyx990803:
Ah, that's actually clever. Although I kinda feel it's not the use case this option should be designed for. You can easily use the same template file and do this when creating the renderer:

``` js
const renderer = createBundleRenderer({
   // ...
   template: template.replace('<div id="app"></div>', '<!--vue-ssr-outlet-->')
})
```

PR #5704, opened 2017-05-19
feat: Add `async` to body scripts (update: Use `defer` instead)

Atinux:

- [x] Feature

- [x] No

`async` tag let the scripts to not block the render of the page, and since it's for SSR, the scripts don't have to be loaded before the first render.


blake-newman:
Looks good although I can't find any specification on the behaviour of multiple async script tags. Does it keep the linear order or will it execute as soon as script is parsed. 

If the latter then there could be race conditions. This may be worth an investigation.

Atinux:
Based on this answer: http://stackoverflow.com/questions/8996852/load-and-execute-order-of-scripts

It may be better to use `defer` instead of `async` to keep the order, what do you think?

blake-newman:
Good find :) 

Yeah defer is the best, as it will keep predicability.

Atinux:
Commits updated 👍

Issue #5732, opened 2017-05-22
<transition> with <keep-alive>: non-component child is not showing.


DeShadow:
### Version
2.3.3

### Reproduction link
[https://jsfiddle.net/9f1tyowd/2/](https://jsfiddle.net/9f1tyowd/2/)

### Steps to reproduce
I want to create transition between component and div with my data (non-component). But also I want to cache my component. That's why my code looks like:
```
<transition name="fade" mode="out-in">
    <keep-alive>
        <div v-if="myCondition">Example text</div>
        <my-child-component v-else></my-child-component>
    </keep-alive>
</transition>
```

But this code is not working and there is no warnings or errors from Vue.js. `<div>` not showing, but component showing. The problem only in `<keep-alive>`, without it's working.

Without <keep-alive> (all is right): https://jsfiddle.net/9f1tyowd/1/
With <keep-alive> (`<div>` never showing): https://jsfiddle.net/9f1tyowd/2/

That's why the question: what is wrong?

If only components are allowed in `<keep-alive>`, why there is no warnings or errors? How can I switch between component and non-component in transition with caching of component?

If non-components are allowed, why they are simply ignored and not showing?

### What is expected?
Caching component and showing of `<div>`

### What is actually happening?
`<div>` is not showing.

posva:
Yeah, it only works with components.
I think we could include a warning for non-components

DeShadow:
@posva Okay. Is there any way to make transition between non-component and cached component? I don't want to create new component for non-component.

posva:
Actually, everything can be a component, there is not a _non-component_ element 😅

posva:
Closing this as seen in #5757 : You must use components inside `keep-alive`

PR #5757, opened 2017-05-25
Warn when keep-alive's child is not component

jkzing:
[#5732](https://github.com/vuejs/vue/issues/5732)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/components/keep-alive.js
+++ b/src/core/components/keep-alive.js
@@ -3,6 +3,8 @@
 import { isRegExp } from 'shared/util'
 import { getFirstComponentChild } from 'core/vdom/helpers/index'
 
+import { tip } from '../util/index'
+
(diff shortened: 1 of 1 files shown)


yyx990803:
This is actually more tricky than it seems. The else branch could also be reached when `<keep-alive>` has no children (`v-if` with false conditions, `<component :is="null"/>`, etc.). So the check will fire in valid use cases as well.

I think the [docs](https://vuejs.org/v2/api/#keep-alive) is already explicit enough about the "component only case" and I think it's fine to not have this warning.

Thanks for the PR though.

jkzing:
😅

PR #5808, opened 2017-06-02
fix #5804: support object value for key attribute

### LINKED ISSUE #5804: use an object as v-for's key will clear input's value
### Version
2.3.3

### Reproduction link
[https://jsfiddle.net/lmk123/fkLhx1n6/](https://jsfiddle.net/lmk123/fkLhx1n6/)

### Steps to reproduce
 1. Type each person's id in the input fields.
 2. Click `Shuffle` button 10 times

### What is expected?
Every person's id should stay in the input fields.

Some person's id has been cleared.

---
Only if the array starts with more than 3 items will reproduce this bug.

javoski:
#5804 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -23,6 +23,7 @@ import {
   isDef,
   isUndef,
   isTrue,
+  isObject,
   makeMap,
   isPrimitive
 } from '../util/index'
@@ -30,6 +31,8 @@ import {
 export const emptyNode = new VNode('', {}, [])
 
 const hooks = ['create', 'activate', 'update', 'remove', 'destroy']
+
+const VNODE_KEY_PROP = '__key_prefix_' + Date.now()
 
 function sameVnode (a, b) {
   return (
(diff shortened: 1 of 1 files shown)


src/core/vdom/patch.js
```diff
@@ -63,9 +66,14 @@ function createKeyToOldIdx (children, beginIdx, endIdx) {
   const map = {}
```
javoski: Looks like a risk of memory leak here😂

yyx990803:
First, thanks for the PR and see https://github.com/vuejs/vue/issues/5804#issuecomment-305950427

Regarding the implementation - not sure if it's a good idea to pollute the observed objects like this. Ideally we should `Object.defineProperty` to make it non-enumerable. But this operation is quite expensive.

One way to work around this might be using native `Map` when it is available, and fallback to the "secret non-enumerable key" strategy used in this PR. However, I'd like to do some more performance benchmarking before actually adding the changes.

javoski:
Adding a warning should be a better idea before native map object can be take into consideration.

PR #5818, opened 2017-06-04
Feat:computed option support function type

lzxb:
use
```javascript
export default {
  computed () {
    // The custom plug-in wants to be able to inject the calculation properties of the entire module into the component
    return this.$myPlugin.modules.getters
  }
}
```


posva:
Can you explain why (specific use case) would you need a function for the computed property?
The `computed` property of a component is kind of like the definition of the component. Making it a function allows it to be dynamic at runtime

yyx990803:
I think I've seen this request before but I don't think we should support this.

Allowing computed properties to be dynamic makes the component's behavior more unpredictable and harder to properly type-check / be statically analyzed. This is not something we'd like to encourage.

On the other hand - this can already be achieved in userland inside the `beforeCreate` lifecycle hook. If you really need that, just do it in `beforeCreate`.

lzxb:
Sure enough, it is possible

Issue #5603, opened 2017-05-04
optionMergeStrategies.inject


eddow:
### Version
2.3.2

### Reproduction link
[https://jsfiddle.net/yywfw3cL/5/](https://jsfiddle.net/yywfw3cL/5/)

### Steps to reproduce
Make a component extend another one while both parent and child define an injection : the parent' injection is lost (overridden)

### What is expected?
Injections to be merged

### What is actually happening?
Injections override parent' ones

---
The injections initialisation looks for the injections own keys - so we cannot merge with a prototype like it is done in function mergeAssets or with "Other object hashes" : props/methods/computed
We instead have to copy the object and extend it in its own properties.
It might be cleaner if it was allowed to use the prototype.

posva:
Thanks for reporting it.

This should be an easy first contribution because it needs little to no knowledge of Vue. The file that needs to be modified is at https://github.com/vuejs/vue/blob/dev/src/core/util/options.js. Inject are arrays or object, it should be able to merge both. `extend` from `shared/utils` can (and prolly should) be used.
In case someone is looking for a first contribution 🙂

eddow:
jaj, ok, got it ;)

eddow:
Salut Posva,
J'essaye de te contacter pour m'aider à installer l'env de dev pour vue (j'ai du mal) - j'essaye FB, sinon pourrais tu me contacter à l'addresse eeddow at protonmail stp?
J'effacerai ce msg quand tu m'auras répondu.

LinusBorg:
> I have a question though : is the easiest way fork->pull request?

Yes.

ryzokuken:
I would like to resolve this as my first contribution to the project.

jkzing:
@ryzokuken Already an opened PR there. :) ([#5827](https://github.com/vuejs/vue/pull/5827))

ryzokuken:
Okay. I'll look someplace else to get acquainted with the source. Thanks.

PR #5827, opened 2017-06-05
Merge inject when extending a component

jkzing:
[#5603](https://github.com/vuejs/vue/issues/5603)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


jkzing:
Actually I have been seeing this issue for several days. But I'm not sure that if the behavior `inject is not extending` is expected or not, because from code level it seems this use case was not taken into consideration before.

So please let me know if this use case valid, and I'll add test case to it.

src/core/util/options.js
```diff
@@ -248,6 +249,30 @@ function normalizeProps (options: Object) {
 }
 
 /**
+ * Normalize all injections into Object-based format
+ */
+function normalizeInject (options: Object) {
+  const inject = options.inject
+  if (!inject) return
```
yyx990803: The normalization is only needed if `inject` is an Array, so we can just return if it is not an Array.

yyx990803: In addition, with normalization here, we should be able to simplify the `resolveInject` function here since the normalized value will always be objects: https://github.com/vuejs/vue/blob/dev/src/core/instance/inject.js#L38

jkzing: Yes, will look into it later.

dsonet: @jkzing you forgot to simplify the resolveInject.

jkzing:
Any news on this PR? @yyx990803

jkzing:
@yyx990803  thanks 🎉

PR #5857, opened 2017-06-12
Provide an option called plugins to process ast in vue-template-compiler, fix #5703

### LINKED ISSUE #5703: Expose the api for directly generating render function code from element ASTs in createCompiler function
It may sounds awkward when using the modified AST to generate render function instead of using template files as input. But, in the following cases, it might be useful:

What if:
- people want to do some optimized operations on AST like what the `optimizer.js` does before jump into the codegen method in the baseCompile function.
-  a pre-disposed component can be loaded via webpack loader which implements the exposed generate function to get the new render function. In this case, I can initially load some customized vue components with new added dom elements, attributes, events and etc.
- I can visually edit the rendered components by modifying the ASTs instead of the original template which is considered to be much easier.
- After the previous step, I can even generate the new template file like the reverse way of template -> AST -> render function.

I'm not sure whether there is an alternative way of doing the things above, but it might be helpful when it comes to something like visualized vue components development IDE.

Maybe in the createCompilerCreator function, we can add a function as property in compile method like:
```
compile.fromAST = function fromAST(
    ast
  ) {
    return generate(ast, baseOptions)
  };
```
in line 49, src/compiler/create-compiler.js

rainfore:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Plugins to process AST like ones in postcss.

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -21,6 +21,8 @@ declare type CompilerOptions = {
 
   // runtime user-configurable
   delimiters?: [string, string]; // template delimiters
+
+  plugins?: Array<Function>; // customizing plugins to process ast
 };
 
 declare type CompiledResult = {
--- a/src/compiler/index.js
+++ b/src/compiler/index.js
@@ -14,6 +14,9 @@ export const createCompiler = createCompilerCreator(function baseCompile (
 ): CompiledResult {
   const ast = parse(template.trim(), options)
   optimize(ast, options)
+  if (options.plugins) {
+    options.plugins.forEach(plugin => plugin(ast, options))
+  }
   const code = generate(ast, options)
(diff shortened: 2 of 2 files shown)


flow/compiler.js
```diff
@@ -21,6 +21,8 @@ declare type CompilerOptions = {
 
   // runtime user-configurable
   delimiters?: [string, string]; // template delimiters
+
+  plugins? Array<Function>; // customizing plugins to process ast
```
gebilaoxiong: forget to write the colon

Issue #478, opened 2014-10-10
Typescript and Vue in general


indus:
I have a small problem; I'm in love with two things at the same time: Typescript and Vue.
I tried a handfull different approaches to make them work together; using TS modules and/or classes with public/private/static properties to shim vue components is the best i got.
I love the intelisense, autocompletion and error hints i get by doing so!
But after all it is kind of hacky.

I wanted to ask you what you think about Typescript? Do you think it is possible to bring them togeher in a more natural way?
Maybe refactoring some parts of your extend routine (to make some of its internals reachable from a typescript construtor) could make a huge difference.

sl-codenine:
Hi,

does anything changed with Vue 0.12.x introduction? We have a business need to use TypeScript. Application it self have to be made of set of small to medium size independent, decoupled components, so Vue looks like a natural choice here. But lack of TS support is a big blocker for us.

RivetsJS looks like abandoned, so that's not an option neither, however it's way to bind any JS object with any jQuery node object is nice and in fact it solves the problem. Is it possible in Vue to implement (in reasonable time amount) such behavior? I mean implementing it in some `real and vue-native` way, not by embedding custom object inside VueComponent object, or anything like that, which will most likely cause some problems down the road  anyway.

If not, are there any other options? As far as i can tell, by quick looking at Vue's code, the `.extend()` concept goes quite deep. Can't see an easy solution for that.

As to the above solutions from @indus and @diverted247 - not sure i understood it correctly, byt - i must say that we're on a beginning of a big app rewrite and the feelings are strong to avoid hacking and unusual use of libraries, so really - we're not happy to hack Vue extend concept this way.

Anyway @indus and @diverted247, did you manage to solve this some other/better way?

And @yyx990803 how do you think if it's possible/reasonable to alter Vue's extend mechanism? And how hard eventually it would be?

yyx990803:
@sl-codenine now that I think about it, it would be possible to simulate `Vue.extend` using annotations (similar to how Angular 2 does it):

``` js
@Vue.component({
  // Vue.extend() options here...
})
class MyComponent extends Vue {}
```

Or, declare the options as a static property and simply add a decorator:

``` js
@Vue.component
class MyComponent extends Vue {
  static options = {
    // options here...
  }
}
```

Maybe even make it look more "classy" if we do a bit more plumbing in the decorator:

``` js
@Vue.component
class MyComponent extends Vue {
  // data
  msg = 'hello!'
  // method
  greet() {
[...]

PR #5887, opened 2017-06-14
Improve Vue type declarations for more canonical usage

DanielRosenwasser:
This PR improves the Vue `.d.ts` files so that users can get some improved inference when using the `noImplicitThis` compiler option. It takes advantage of many of the features the TypeScript team has been working on over the last few versions. Thus, it requires a minimum TypeScript version of 2.4. You can get this working with the current `typescript@next` (which is `2.5.0-dev.20170614`, but don't let that version number fool you - this will work with the final release of 2.4).

The best part about this change is that it makes using Vue easier when users choose not to use classes and inheritance. In other words, users and can more easily `new` up an instance using the Vue constructor, and can create components using the `Vue.extend` and `Vue.component` APIs a little more naturally. Extending Vue as a class will still work, so if you prefer to use vue-class-component and the rest, that should still work! However, those dependencies may need to be updated for this change.

There's definitely some amount of complexity that's been introduced but I'm willing to iterate on this with the help of others. Given the feedback that we've heard on [TypeScript-Vue-Starter](https://github.com/Microsoft/TypeScript-Vue-Starter), it's been a positive experience for users.

Fixes many of the issues users have asked about in #478.

Tests are probably failing at this point because I need to update the project's dependency on TypeScript.


yyx990803:
Thanks @DanielRosenwasser ! Are there any remaining incompatibilities aside from the need for using latest dependancies, e.g. required change in `tsconfig`?

HerringtonDarkholme:
hmmm, the test is failing. 

It seems ` Base constructor return type 'CombinedVueInstance<Data, Methods, Computed, Record<PropNames, any>>' is not a class or interface type.`. 

So interface augmentation fails.

types/options.d.ts
```diff
@@ -1,43 +1,82 @@
...
+  | ThisTypedComponentOptionsWithArrayProps<Data, Methods, Computed, PropNames>
+  | ThisTypedComponentOptionsWithRecordProps<Data, Methods, Computed, Record<PropNames, PropOptions>>;
+
+
+export interface ComponentOptions<Data, Methods, Computed, Props> {
+  data?: Data;
+  props?: Props;
   propsData?: Object;
-  computed?: { [key: string]: ((this: V) => any) | ComputedOptions<V> };
-  methods?: { [key: string]: (this: V, ...args: any[]) => any };
```
HerringtonDarkholme: We might also need `methods?: ThisType<Methods & Computed & Pops & Data>` to fully capture Vue's `this` instance. I find this 

I haven't try it myself, but will it cause cyclic reference?

DanielRosenwasser: There are a few problems that potentially come up if you try to reference `methods` from `data` in which you'll have to have an explicit annotation on `data()`.

> We might also need methods?: ThisType<Methods & Computed & Pops & Data> to fully capture Vue's this instance.

@HerringtonDarkholme that's not actually necessary. When a method needs its `this `type, it will walk up each contextually typed object literal until it finds one that had a contextual type consisting of `ThisType`. If it does, it uses it. If not, it uses the innermost object literal's contextual type, and if there isn't one it just uses the type of the containing literal.

DanielRosenwasser:
@yyx990803 the fact that `ComponentOptions` used to be generic on 1 type arg, and now has 4 unrelated type arguments, will cause some incompatibilities. That's the main one you'll run into with vue-class-component.

DanielRosenwasser:
The build should now be fixed.

types/options.d.ts
HerringtonDarkholme: Tried it locally. It works like charm!

types/vue.d.ts
```diff
@@ -67,8 +68,51 @@ export declare class Vue {
   $nextTick(callback: (this: this) => void): void;
   $nextTick(): Promise<void>;
   $createElement: CreateElement;
+}
+
+export type CombinedVueInstance<Data, Methods, Computed, Props> = Data & Methods & Computed & Props & Vue;
+export type ExtendedVue<Constructor extends VueConstructor, Data, Methods, Computed, Props> =
+  (new (...args: any[]) => CombinedVueInstance<Data, Methods, Computed, Props>) &
```
ktsn: As `ExtendedVue` is a sub-class of the Vue constructor, we can expect the constructor signature would only receive one argument like below?

```ts
new (options: any) => CombinedVueInstance<Data, Methods, Computed, Props>
```

Not sure whether we could annotate the `options` with `ThisTypedComponentOptionsWithArrayProps` etc. since the type declaration might be so complicated 😂

types/vue.d.ts
```diff
@@ -67,8 +68,51 @@ export declare class Vue {
...
+
+  extend<VC extends VueConstructor, PropNames extends string = never>(this: VC, definition: FunctionalComponentOptions<PropNames[], Record<PropNames, any>>): ExtendedVue<VC, {}, {}, {}, Record<PropNames, any>>;
+  extend<VC extends VueConstructor, Props extends Record<string, PropValidator>>(this: VC, definition: FunctionalComponentOptions<Props, Record<keyof Props, any>>): ExtendedVue<VC, {}, {}, {}, Record<keyof Props, any>>;
+  extend<VC extends VueConstructor, Data, Methods, Computed, PropNames extends string = never>(this: VC, options: ThisTypedComponentOptionsWithArrayProps<Data, Methods, Computed, PropNames>): ExtendedVue<VC, Data, Methods, Computed, Record<PropNames, any>>;
+  extend<VC extends VueConstructor, Data, Methods, Computed, Props extends Record<string, PropValidator>>(this: VC, options?: ThisTypedComponentOptionsWithRecordProps<Data, Methods, Computed, Props>): ExtendedVue<VC, Data, Methods, Computed, Record<keyof Props, any>>;
```
ktsn: In the following code, `this` type in `Child` component options does not seem to have parent's properties while the `new`ed `Child` component has it.

```ts
const Parent = Vue.extend({
  data () {
    return { parent: 'Hello' }
  }
})

const Child = Parent.extend({
  methods: {
    foo () {
      this.parent // Compile error
    }
  }
})

const child = new Child()
child.parent // No error
```

But I'm not sure how we can solve this. It may need more type parameters to extract the parent instance types into `ThisType` of the child component options....

types/vue.d.ts
DanielRosenwasser: > As ExtendedVue is a sub-class of the Vue constructor, we can expect the constructor signature would only receive one argument like below?

@ktsn ExtendedVue creates an intersection with the mixin constructor `(new (...args: any[]) => CombinedVueInstance<Data, Methods, Computed, Props>)`. This means it just absorbs the argument types of the `Constructor` type that's given. See https://github.com/Microsoft/TypeScript/pull/13743 for more details

types/vue.d.ts
DanielRosenwasser: Good catch! I think I can fix this.

types/vue.d.ts
ktsn: Oh, I didn't understand about the mixin constructor correctly. Thank you for the explanation.

DanielRosenwasser:
I just sent out a fix. In the previous iteration, `extend` inherited members on the static side instead of the instance side. I've now done the opposite. Give it another go - I've added the test that @ktsn mentioned above.

DanielRosenwasser:
@yyx990803 I could technically restore the commit if you don't want to fix #5016 right away.

types/vue.d.ts
```diff
@@ -20,30 +24,27 @@ export type CreateElement = {
...
   readonly $parent: Vue;
   readonly $root: Vue;
   readonly $children: Vue[];
-  readonly $refs: { [key: string]: Vue | Element | Vue[] | Element[]};
+  readonly $refs: { [key: string]: Vue | Element | Vue[] | Element[] };
   readonly $slots: { [key: string]: VNode[] };
   readonly $scopedSlots: { [key: string]: ScopedSlot };
   readonly $isServer: boolean;
+  readonly $data: Record<string, any>;
+  readonly $props: Record<string, any>;
```
HerringtonDarkholme: We can also make Vue generic so `$data` and `$props` can be typed. But I wonder if it is an overkill.

types/vue.d.ts
```diff
@@ -67,8 +68,49 @@ export declare class Vue {
...
+  component<PropNames extends string = never>(id: string, definition: FunctionalComponentOptions<PropNames[], Record<PropNames, any>>): ExtendedVue<V, {}, {}, {}, Record<PropNames, any>>;
+  component<Props extends Record<string, PropValidator>>(id: string, definition: FunctionalComponentOptions<Props, Record<keyof Props, any>>): ExtendedVue<V, {}, {}, {}, Record<keyof Props, any>>;
+  component<Data, Methods, Computed, PropNames extends string = never>(id: string, definition: ThisTypedComponentOptionsWithArrayProps<V, Data, Methods, Computed, PropNames>): ExtendedVue<V, Data, Methods, Computed, Record<PropNames, any>>;
+  component<Data, Methods, Computed, Props extends Record<string, PropValidator>>(id: string, definition?: ThisTypedComponentOptionsWithRecordProps<V, Data, Methods, Computed, Props>): ExtendedVue<V, Data, Methods, Computed, Record<keyof Props, any>>;
 
-  static config: {
+  use<T>(plugin: PluginObject<T> | PluginFunction<T>, options?: T): void;
+  mixin(mixin: typeof Vue | ComponentOptions<any, any, any, any>): void;
```
HerringtonDarkholme: `mixin` does not have `ThisTyped` as `component` or `extend`, is this intended?

types/options.d.ts
```diff
@@ -1,43 +1,82 @@
...
 
-export interface ComponentOptions<V extends Vue> {
-  data?: Object | ((this: V) => Object);
-  props?: string[] | { [key: string]: PropOptions | Constructor | Constructor[] };
+/**
+ * This type should be used when an array of strings is used for a component's `props` value.
+ */
+export type ThisTypedComponentOptionsWithArrayProps<Instance extends Vue, Data, Methods, Computed, PropNames extends string> =
+  object &
+  ComponentOptions<Data | ((this: Record<PropNames, any> & Instance) => Data), Methods, Computed, PropNames[]> &
```
HerringtonDarkholme: 👍  for only exposing prop!

DanielRosenwasser: I agree! To be honest this was more a limitation of the type system but it honestly seemed like more of a feature than a bug. 😄

types/options.d.ts
```diff
@@ -50,48 +89,55 @@ export interface ComponentOptions<V extends Vue> {
...
+export interface RenderContext<Props> {
+  props: Props;
   children: VNode[];
   slots(): any;
   data: VNodeData;
   parent: Vue;
   injections: any
 }
 
+export type PropValidator = PropOptions | Constructor | Constructor[];
```
DanielRosenwasser: I originally tried to do that but the more I thought about it, the more time I realized it'd take to finish. I figured it'd be better to put the PR out and try that at a later stage.

types/vue.d.ts
DanielRosenwasser: I originally made it generic but ran into problems.

Namely, `VueConstructor` has two construct signatures. Each of those signatures has two different sets of type parameters. They each constructed, at least in part, a Vue<Data, Methods, Computed, Props>`,  but each gave different type arguments. 

When extending a `Vue<Data, Methods, Computed, Props>, TypeScript tries to find the right construct signature, but it can't because it's not really possible to make a meaningful set of type arguments for `Vue` that would create the same output type for both signatures.

What it comes down to is that TypeScript has always had an assumption that the type parameters of a construct signature have had a direct correlation with the type parameters of the constructed type - but that's not entirely the case here. At least, that's my understanding of the problem.

So I could try to address this, but maybe down the line. Plus, does anyone need to access `$data` and `$props` outside of debugging?

types/vue.d.ts
DanielRosenwasser: This is intended - I don't think `ThisType` would help the common scenarios, and I think don't think we can model the scenario all that well. For example, "get the type of all the `methods` from mixins and intersect them together` isn't possible as far as I know at the moment.

types/options.d.ts
```diff
@@ -1,43 +1,82 @@
...
+  [K in keyof T]: (() => T[K]) | ComputedOptions<T[K]>
+}
 
-export interface ComponentOptions<V extends Vue> {
-  data?: Object | ((this: V) => Object);
-  props?: string[] | { [key: string]: PropOptions | Constructor | Constructor[] };
+/**
+ * This type should be used when an array of strings is used for a component's `props` value.
+ */
+export type ThisTypedComponentOptionsWithArrayProps<Instance extends Vue, Data, Methods, Computed, PropNames extends string> =
```
DanielRosenwasser: Is that a great idea? Does re-exporting mean that this is committed to as part of the public interface?

types/vue.d.ts
HerringtonDarkholme: Thanks for the detailed explanation!

Indeed, `$data` and `$props` are mainly for advanced usage like delegate / wrapper component ([example](https://github.com/vuejs/vue/issues/4571)). We would like to check template code in future. So a `$props` or `$data` might be helpful. 

For ordinary users, `Record` should be enough.

DanielRosenwasser:
> Do you have any suggestion for library author to test compiling error?

We have [dtslint](https://github.com/Microsoft/dtslint#write-tests) which has a testing facility to `$ExpectType` and `$ExpectError`. I could add some tests there if you'd like.

types/options.d.ts
DanielRosenwasser: Also, should I just rename this to `ComponentOptionsWithArrayProps`?

HerringtonDarkholme:
@DanielRosenwasser Thanks for sharing dtslint! Definitely I will try to add it to Vue.

Since this pull request is already big, IMHO it is fine to separate dtslint to another PR.

yyx990803:
@ktsn @HerringtonDarkholme seeing that both of you have approved this, I think the next step should be working out a upgrade guide - i.e. what would be different from the [current recommendation](https://vuejs.org/v2/guide/typescript.html).

@ktsn Does `vue-class-component` need to be upgrade to be compatible with the new types?

ktsn:
@yyx990803 I think we need to update `vue-class-component` since it uses `ComponentOptions` in its declarations.

yyx990803:
Ok - I think I'll hold this off until:

1. TS 2.4 is out
2. `vue-class-component` has been updated.

This most likely means we will ship Vue 2.4 without the new types.

alexdresko:
Okay, lemme put this out there, as I believe this is the correct thread for what I'm dealing with. 

Right now, a basic reproduction of the Vue Getting Started doesn't work in TypeScript. 

image

The only way to make that code work is to import Vue (I'm using this PR's definitions)

image

However, it doesn't appear that modules are a requirement for Vue, given I can do this in a basic HTML file: 

image

We're a bit behind the times in some of our projects, though we do use TS. We just don't use modules. `import`ing Vue forces TS to think we're using modules, and that breaks a lot of our existing code. 

Are the definitions at fault here, or my general lack of understanding otherwise? 

If it's the definitions, maybe that's something you can take into consideration.

ktsn:
@alexdresko
https://forum.vuejs.org/t/typescript-is-there-a-type-definition-for-vue2-and-how-to-use-it/1741/7?u=ktsn

DanielRosenwasser:
@alexdresko what would be ideal is a UMD-style declaration, however, currently we can only say that `Vue` has the shape of `index.ts`, not that it has the shape of the Vue default export. CC @RyanCavanaugh.

alexdresko:
Didn't realize we could just conjure up @RyanCavanaugh from anywhere. Cool. Do we have to @ him three times for it to work?

prashantkhurana:
Once merged, will this allow us to type the props also i.e. props implementing an interface? Run time checking would be nice, but I am more interested in the code completion, compile error etc the ide will provide once the props are typed.

DanielRosenwasser:
@prashantkhurana unfortunately not, but I can try to revisit this. You'll at least have TypeScript aware of your `props` though.

prashantkhurana:
thanks @DanielRosenwasser . yeah that would be the icing on the cake:)

kaorun343:
Hi.

I just wondered how to add Dom references to `$refs`.
Current definitions allow us to write like this, but the future version won't allow to do so.

```ts
interface App extends Vue {
  $refs: {
    target: HTMLElement
  }
}

export default {
  created() {
    this.$refs.target
  }
} as ComponentOptions<App>
```

If many people doesn't care about this, I think there's no need to fix this PR, but if somebody takes care about this, it should be done so.
For me, I haven't written like this yet, so I don't care about it...
How do you think about this?

HerringtonDarkholme:
@kaorun343 I have come up with this: 

Rather than keeping `ComponentOption` as an `interface`, making it a `type` alias seems better choice. With the aid of default generic argument, we can even reduce the impact of breaking change.

Example:

```ts
type ComponentOption<Vue, D={}, M={}, C={}, P={}> = InternalOption<D,M,C,P> & ThisType<D&M&C&P&Vue>

interface InternalOption<Data, Methods, Computed, Props> {
    methods?: Methods
}

interface App {
    $refs: {
        target: HTMLInputElement
    }
}

var a: ComponentOption<App> = { // no breaking change
    methods: {
        test() {
            this.$refs.target
        }
    }
}

interface MyComponentOption extends ComponentOption<App, {foo: number}> { // refine more
}
```

HerringtonDarkholme:
One major issue is typed prop declaration:

```ts
var a = new Vue({
    props: {
        someProp: {
            type: String
        }
    },
    methods: {
        test() {
            this.someProp.toLowerCase() // error
        }
    }
})
```

Another issue is still how to refine `$refs`, `$el` and friends. Ideally, if we can support such usage

```ts
interface App {
    $refs: {
        target: HTMLInputElement
    }
}

new Vue<App>({  // we only provide the first type argument, other type arguments are inferred.
  methods: {
     setVal() {
       this.$refs.target.value = this.someProp
    }
   },
  props: ['someProp']
})
```
However, such API requires optional generic type inference as suggested in https://github.com/Microsoft/TypeScript/issues/14400

prashantkhurana:
@HerringtonDarkholme  I have been thinking about prop typing a bit. One (not so elegant maybe, but quick) way to do prop typing is inferring the types from the default value of the prop.

image

Another is to from the type field but that required me to do ugly casting. 

image

championswimmer:
What's the status with this ? 

I pulled it locally, and tried it out in my project, and everything seems to work pretty well. Almost negates the need of `vue-class-component`

DanielRosenwasser:
@HerringtonDarkholme the problem with your suggestion is that it still breaks people who've done module augmentations (because you're moving from an interface to a type alias.

Would it be okay to pull this in and then revisit that change instead? It might be easier for the core team to gauge the impact of that.

package.json
```diff
@@ -118,7 +118,7 @@
     "selenium-server": "^2.53.1",
     "serialize-javascript": "^1.3.0",
     "shelljs": "^0.7.8",
-    "typescript": "^2.3.4",
+    "typescript": "2.5.0-dev.20170615",
```
znck: Shouldn't it be depending on a stable version?

nickmessing: It depended on some experimental changes but afaik they landed in 2.4, didn't they @DanielRosenwasser?

HerringtonDarkholme:
@DanielRosenwasser  Indeed.

After several trials, I think changing to `interface ComponentOptions<V, D=any, M=any, C=any, P=any>` might introduce fewer upgrading issues.

package.json
loopArray: Is there any progress? Looking forward to it.

blake-newman:
@HerringtonDarkholme @DanielRosenwasser Anything that still needs to be done, can i help out with everything?

HerringtonDarkholme:
@blake-newman I think this pull request needs update dependency and tweak `Props` option type.

@DanielRosenwasser If you're busy with other things, I'm glad to take this over.

loopArray:
Hi, @DanielRosenwasser When I clone your branch [vue-router](https://github.com/DanielRosenwasser/vue-router.git)，It seems not work. It took me a day to find out why.Fortunately, I found it.

This is not sure of the return value。
typescript_2
typescript_3

This is the right to return to rewrite.
typescript_1
typescript
I just started to learn Vue recently, do not know to rewrite the right.
That branch, I can't open an issue, and that's relevant, so I submitted it here.

DanielRosenwasser:
Just pushed a few new commits, but not sure why it's failing.

```
SUMMARY:
✔ 969 tests completed
✖ 1 test failed

FAILED TESTS:
  Transition mode
    ✖ transition out-in on async component (resolve after leave complete)
      PhantomJS 2.1.1 (Linux 0.0.0)
    Expected 1 to be 0.
    webpack:///test/unit/features/transition/transition-mode.spec.js:480:44 <- index.js:149624:44
    shift@webpack:///test/helpers/wait-for-update.js:24:32 <- index.js:90813:36
```

DanielRosenwasser:
Just amended and re-pushed. Looks like there is a flaky test.

DanielRosenwasser:
@HerringtonDarkholme if you can merge and take over, that would be absolutely fantastic. I'd be glad to help however I can.

loopArray:
Hi, @DanielRosenwasser When I create a component with `Vue.extend` or `Vue.component `and I refer to this component in other ts file, I want to assign a type to the $refs or member of data but the `ts` tell me `the name is not found`.
```
export default Vue.extend({...})

export default Vue.component({...})
```
in other file
```
import VueECharts from 'lib/vue-echarts';
... ...
Vue.extend({
    data: {
        center: <IBaiduMapPoint>{
            lng: 103.388611,
            lat: 35.563611
        },
        echartsInstance: <VueECharts | null >null,   ~~~ `[ts] Cannot find name 'VueECharts'.`

    }
});
```
How do I use this type? Or I'm using it wrong.
qq 20170817174700

yyx990803:
Closing with the work being continued in #6391

types/options.d.ts
pkf1994: comment it

pkf1994:
comment it.

chenyayong:
>

PR #5889, opened 2017-06-15
Be able to use string type index in array, fix #5884

### LINKED ISSUE #5884: Vue.set Api strange behavior if path is a numerical string
### Version
2.3.4

### Reproduction link
[https://jsfiddle.net/v0dbnvv0/1/](https://jsfiddle.net/v0dbnvv0/1/)

### Steps to reproduce
Vue.set(array, "indexInTheArrayInString", value) is not working as expected compare to Vue.set(array, "indexInTheArrayInNumericForm", value)

### What is expected?
Both version working the same i suppose?

Watcher on the value are not triggered even if the value is correctly changed in the model

---
My currently fixed in my code:
export function AdvVueSet(obj, path, value) {
    if (!isNaN(path)) {
        path =  +path;
    }
    Vue.set(obj, path, value);
}

Peter-WF:
https://github.com/vuejs/vue/issues/5884

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -189,7 +189,7 @@ export function defineReactive (
  * already exist.
  */
 export function set (target: Array<any> | Object, key: any, val: any): any {
-  if (Array.isArray(target) && typeof key === 'number') {
+  if (Array.isArray(target) && (typeof key === 'number' || /^\d+$/.test(key))) {
     target.length = Math.max(target.length, key)
     target.splice(key, 1, val)
     return val
--- a/test/unit/features/global-api/set-delete.spec.js
+++ b/test/unit/features/global-api/set-delete.spec.js
@@ -96,5 +96,23 @@ describe('Global API: set/delete', () => {
         expect(vm.$el.innerHTML).toBe('')
       }).then(done)
     })
+
+    it('be able to use string type index in array', done => {
+      const vm = new Vue({
+        template: '<div><p v-for="obj in lists">{{obj.name}}</p></div>',
+        data: {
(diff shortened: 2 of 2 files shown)


src/core/observer/index.js
```diff
@@ -189,7 +189,7 @@ export function defineReactive (
  * already exist.
  */
 export function set (target: Array<any> | Object, key: any, val: any): any {
-  if (Array.isArray(target) && typeof key === 'number') {
+  if (Array.isArray(target) && !isNaN(parseFloat(key))) {
```
posva: I think something like a double `typeof...` (+ `key !== ''`) or `key !== '' && typeof Number(key) === 'number'` would be better. `parseFloat` looks like an overkill and `isNaN` is not supported on IE. wdyt @defcc ?

yyx990803: `isNaN` is supported in IE. The problem is that `parseFloat` would turn strings like `"123ABC"` into `123`, which should probably be treated as string keys instead.

I think we should only support strings that are integers, something like `typeof key === 'number' || /^\d+$/.test(key)`

posva: I see, `isNaN` exists as a global function but `Number.isNaN` does not exist 😨

Peter-WF: Thank you all! 
It'd be a good idea to replace `parseFloat` with `regular`

PR #5890, opened 2017-06-15
Special attributes are invalid when using object syntax in v-bind, fix #5881

### LINKED ISSUE #5881: key attribute set by `v-bind="obj"` is ignored by transition-group
### Version
2.3.4

### Reproduction link
[https://codepen.io/Nirazul/pen/RgGRjm](https://codepen.io/Nirazul/pen/RgGRjm)

### Steps to reproduce
1. Click on the trigger button
2. There is only one list visible instead of two

### What is expected?
According to the [documentation](https://vuejs.org/v2/api/#key) I was expecting that I can use the v-bind object notation to set the key attribute.

The list is not rendered when the `key` attribute is not directly present on the element.

---
I'm aware that `key` is a special attribute and that there may be some sort of "hen and egg" problem at work when rendering items in a transition-group.
If the behaviour is intentional, there's maybe some space for improvement in the documentation.

gebilaoxiong:
https://github.com/vuejs/vue/issues/5881

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/render-helpers/bind-object-props.js
+++ b/src/core/instance/render-helpers/bind-object-props.js
@@ -1,7 +1,13 @@
 /* @flow */
 
 import config from 'core/config'
-import { isObject, warn, toObject } from 'core/util/index'
+
+import {
+  warn,
+  isObject,
+  toObject,
+  isReservedAttribute
+} from 'core/util/index'
 
 /**
  * Runtime helper for merging v-bind="object" into a VNode's data.
@@ -24,7 +30,11 @@ export function bindObjectProps (
       }
       let hash
       for (const key in value) {
-        if (key === 'class' || key === 'style') {
+        if (
+          key === 'class' ||
+          key === 'style' ||
+          isReservedAttribute(key)
+        ) {
           hash = data
         } else {
           const type = data.attrs && data.attrs.type
--- a/flow/vnode.js
+++ b/flow/vnode.js
@@ -35,6 +35,7 @@ declare interface VNodeData {
   key?: string | number;
   slot?: string;
   ref?: string;
+  is?: string;
   pre?: boolean;
   tag?: string;
   staticClass?: string;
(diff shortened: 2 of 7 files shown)


gebilaoxiong:
ah... 

originally I wanted to change `v-bind` in compile-time, but  it will be a  breaken change.

because the attribute `is` need special treatment.

so I try to add a attribute (`is`) in `VNodeData`

can someone help me see to review it?

I will add unit tests tomorrow.

thx.

src/core/instance/render-helpers/bind-object-props.js
```diff
@@ -1,8 +1,11 @@
 /* @flow */
 
 import config from 'core/config'
+import { makeMap } from 'shared/util'
 import { isObject, warn, toObject } from 'core/util/index'
 
+const isSpecialAttributes = makeMap('key,ref,slot,is')
```
yyx990803: There is a similar util [here](https://github.com/vuejs/vue/blob/dev/src/core/instance/state.js#L57-L61)

Let's extract it into `shared/util` as `isReservedAttribute` to avoid duplication.

yyx990803: Also could you add a test case for each of these attributes?

gebilaoxiong: @yyx990803 

sorry,  It was too late yesterday, I will finish after work today

Do not we need to consider `is`?

yyx990803: `is` should be included. I missed it earlier.

gebilaoxiong:
@yyx990803  

the `props` test case is incomplete, I improve it in the code

orennitzan:
I know it's a bit old, but the documentation recommend to add :key='item.someid' but doe to the fact that key is special it doesn't render to the dom. Is the any solution with that?
Thanks.

PR #5904, opened 2017-06-17
Allow white space when v-binding to an array or object (fix #4881)

### LINKED ISSUE #4881: Chaining multiple filters without spaces doesn't work

JimRottinger:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Not really a _bug_ but prevents a compilation error when you have white space in array or object bindings which is a common way to write them.

- [ ] Yes
- [x] No

**Other information:**

When binding a value using array or object syntax, it is much more human readable and human writable to allow white space within it. This PR makes a change to the 5th capture group of the HTML parser regex to not break the match when a white space appears, thus allowing you to have spaces in your array/object syntax.

For example, if you have a template like one of these:
```js
template: `<p :class=[color, decoration]>{{message}}</p>`,
//or
template: `<p :class={enabled: isEnabled, error: hasError}>{{message}}</p>`
```
The current parser will throw a template compilation error `- invalid expression: :class="[color,"`. [Here is a simple jsfiddle showing the issue](https://jsfiddle.net/jimrottinger/50wL7mdz/40869/). The changes in this PR correctly grabs the entire array or object, white spaces included.

### DIFF
--- a/test/unit/modules/compiler/parser.spec.js
+++ b/test/unit/modules/compiler/parser.spec.js
@@ -123,7 +123,7 @@ describe('parser', () => {
   it('not warn 3 root elements with v-if, v-else-if and v-else', () => {
     parse('<div v-if="1"></div><div v-else-if="2"></div><div v-else></div>', baseOptions)
     expect('Component template should contain exactly one root element')
-        .not.toHaveBeenWarned()
+      .not.toHaveBeenWarned()
   })
 
   it('not warn 2 root elements with v-if and v-else on separate lines', () => {
@@ -142,7 +142,7 @@ describe('parser', () => {
       <div v-else></div>
     `, baseOptions)
     expect('Component template should contain exactly one root element')
-        .not.toHaveBeenWarned()
+      .not.toHaveBeenWarned()
 
     parse(`
       <div v-if="1"></div>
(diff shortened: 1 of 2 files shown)


test/unit/modules/compiler/parser.spec.js
```diff
@@ -123,7 +123,7 @@ describe('parser', () => {
   it('not warn 3 root elements with v-if, v-else-if and v-else', () => {
     parse('<div v-if="1"></div><div v-else-if="2"></div><div v-else></div>', baseOptions)
     expect('Component template should contain exactly one root element')
-        .not.toHaveBeenWarned()
```
JimRottinger: these lines were throwing `Expected indentation of 6 spaces but found 8 indent`.

yyx990803:
Hmm, I don't think this is a good idea - this makes the template invalid HTML, thus breaking compat with other HTML-based tooling. Thanks for the PR though!

Issue #5937, opened 2017-06-20
Add .sync modifier to v-bind directive


maurop123:
### What problem does this feature solve?
Currently, this syntax is supported,

```
<component v-bind="{propA, propB}" />
```

However, if you want to use the `.sync` modifier on the props, then you can't use the ES6 object destructuring syntax anymore.

```
<component :propA.sync="propA" :propB.sync="propB" />
```

### What does the proposed API look like?
```
<component v-bind.sync="{propA, propB}" />
```

PR #5943, opened 2017-06-21
add sync modifier with bind object

Kingwl:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
#5937

### DIFF
--- a/flow/component.js
+++ b/flow/component.js
@@ -120,7 +120,7 @@ declare interface Component {
   // renderSlot
   _t: (name: string, fallback: ?Array<VNode>, props: ?Object) => ?Array<VNode>;
   // apply v-bind object
-  _b: (data: any, value: any, asProp?: boolean) => VNodeData;
+  _b: (data: any, tag: string, value: any, asProp: boolean, isSync?: boolean) => VNodeData;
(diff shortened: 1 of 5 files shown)


src/compiler/parser/index.js
```diff
@@ -483,7 +479,13 @@ function processAttrs (el) {
       } else if (onRE.test(name)) { // v-on
         name = name.replace(onRE, '')
         addHandler(el, name, value, modifiers, false, warn)
-      } else { // normal directives
+      } else {
+        // bind update event for sync modifier and bind object
+        if (modifiers && modifiers.sync && name === 'v-bind') {
+          addSyncUpdateHandler(el, value, value)
```
posva: I think it's a bit more complicated than that as we need to add a handler for every property in the `v-bind` object but I haven't researched much.

Also, imagine the case where `value` is directly an object: `{ foo, bar }`, we really want to add `update:foo` and `update:bar`

Kingwl: fixed😃

PR #5993, opened 2017-06-28
added Object key sort prior to JSON.stringify comparison (fix #5908)

### LINKED ISSUE #5908: JSON.stringify in looseEqual does not have necessary keys in the same order
I see that in [looseEqual](https://github.com/vuejs/vue/blob/dev/src/shared/util.js#L238) function you are using `JSON.stringify`, but `JSON.stringify` has no quarantines on key order when serializing. [See here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description):

> Properties of non-array objects are not guaranteed to be stringified in any particular order. Do not rely on ordering of properties within the same object within the stringification.

drewsmith:
#5908

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe: Improvement

- [ ] Yes
- [X] No

**Other information:**

### DIFF
--- a/test/unit/modules/shared/key-sort.spec.js
+++ b/test/unit/modules/shared/key-sort.spec.js
@@ -1,0 +1,55 @@
+import { keySort, looseEqual } from 'shared/util'
+
+describe('keySort', () => {
+  const chars = '0123456789abcdefghijklmnopqrstuvwxyz!@#$%^&*()_+./'
+  const randomWord = () => {
+    const letters = []
+    for (let cnt = 0; cnt < 10; cnt++) {
+      letters.push(chars.charAt(Math.floor(Math.random() * chars.length)))
+    }
+    return letters.join('')
+  }
+  const unsortedObject = () => {
+    const unsorted = {}
+    chars.split('').map(c => {
+      unsorted[c + '' + randomWord()] = randomWord()
+    })
+    return unsorted
+  }
+
+  it('returns Object for empty Object argument', () => {
+    expect(looseEqual(keySort({}), {})).toBe(true)
+  })
+
+  it('returns Object for Array argument', () => {
+    expect(looseEqual(keySort([]), {})).toBe(true)
+  })
+
+  it('returns Object for String argument', () => {
+    expect(looseEqual(keySort('Test String'), {})).toBe(true)
+  })
+
+  it('returns Object for Number argument', () => {
+    expect(looseEqual(keySort(Number.MAX_SAFE_INTEGER), {})).toBe(true)
+  })
+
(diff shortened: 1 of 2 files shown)


jkzing:
This maybe not working because AFAIK it's `JSON.stringify` that has no guarantee on the order of properties serialized by it. 🤔

drewsmith:
Definitely a somewhat trivial edge case. There are two ways to ensure this AFAIK. This approach has some reusability or you can pass sorted keys to stringify ```JSON.stringify(obj, Object.keys(obj).sort())```

test/unit/modules/shared/key-sort.spec.js
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  it('returns Object for Null argument', () => {
+    expect(looseEqual(keySort(null), {})).toBe(true)
+  })
+
+  it('Sorts an unsorted Object, which are equal', () => {
+    const unsortedTestObj = unsortedObject()
+    const sortedObject = keySort(unsortedTestObj)
+    expect(looseEqual(sortedObject, sortedObject)).toBe(true)
```
znck: Either one should be `unsortedTestObj` here.

znck:
@drewsmith Is this right approach?

If I compare key by key in two objects, it would take `O(N)` comparisons.
If I sort then compare stringified JSON, it would take `O(N logN)` comparisons + overhead of `stringify`.

drewsmith:
@znck Thanks for the comment. Are you referring to passing sorted keys to ```stringify``` or sorting altogether?

The _right_ approach is deep object equality IMO, which ```stringify``` is a low tech mechanism for doing so since it serializes the entire object including values. The fact the order is not deterministic requires some sort of explicit ordering or perhaps ```stringify``` should be replaced with something else altogether?

znck:
I want to say, sorting the keys beats the purpose of `looseEqual`.

Also, `looseEqual` is used only in [model.js](https://github.com/vuejs/vue/blob/master/src/platforms/web/runtime/directives/model.js) for select options, I guess `stringify` serves the purpose there. @yyx990803 would know this better.

drewsmith:
I agree with you depending on semantics. Does loose mean implicit deep copy, or is it a synonym for shallow? ```stringify``` gets you a poor man's deep copy, but either way values need to be compared.

yyx990803:
Thanks for the PR - however, cloning with sorted keys results in extra memory allocation and it's better to just compare in place. See a8ac129

PR #5996, opened 2017-06-29
support scopedSlots on html elements and components

javoski:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -424,8 +424,9 @@ function processSlot (el) {
     if (slotTarget) {
       el.slotTarget = slotTarget === '""' ? '"default"' : slotTarget
     }
-    if (el.tag === 'template') {
-      el.slotScope = getAndRemoveAttr(el, 'scope')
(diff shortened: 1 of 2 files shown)


yyx990803:
While the extra `<template scope="...">` is more verbose, it's also more obvious/explicit that it is a scoped slot. I intentionally only allowed it on `<template>` for this reason.

Also, just the presence of `scope` is ambiguous - e.g. you may want to bind a prop named `scope` on a component.

javoski:
I agree that we should make it more explicit.
Or maybe scoped slot on html element or component should be implemented through a more specific way, like ```v-scope``` or something else. 

BTW, the condition checking in [codegen - genScopedSlot()](https://github.com/vuejs/vue/blob/dev/src/compiler/codegen/index.js#L342) looks unnecessary if scoped slot is only allowed on ```<template>```

PR #6015, opened 2017-07-02
test: test bundle size

ai:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

[size-limit](https://github.com/ai/size-limit) is a tool to show how many kilobytes your JS library increases the user bundle. You can set size budget and size-limit will prevent JavaScript libraries bloat on Travis CI.

It already used in Autoprefixer, PostCSS and JSS. For instance, I reduce Logux Client size 2x because I found some unexpected huge dependencies.

For Vue this tool can be useful to always know project size and prevent unexpected size regression.


posva:
Thanks for the pr Andrey 🙂 
I'm not sure this is useful for Vue because it doesn't have any dependency and also because we use rollup for bundling. The tools looks interesting, really like the fact that you added a way to run the bundle analyzer

ai:
@posva yeap, tool is much effective in project with dependencies :D.

But even without dependencies some wrong `process` usage could increase bundle size for few KB.

Anyway, Vue is already small, so it is up to you, to lint bundle size or not :)

ai:
@posva BTW, React is planning to add this kind of tool too https://github.com/facebook/react/pull/10090

But since you have no dependencies and already build-in file, maybe `bundlesize` by @siddharthkp will be better for you. It doesn’t run webpack, so results is not so accurate, but in other hand tool works faster.

posva:
@ai I see, yeah, bundlesize may be useful since it checks a generated file but lets you generate it as you want

yyx990803:
Hey thanks for the PR @ai - in fact, Vue's build script already logs the size for every dist file when they are built, and we don't have any dependencies, so I believe an extra tool is unnecessary.

PR #6022, opened 2017-07-03
fix: the first LF following pre and textarea tag should be ignored, fix #5992

### LINKED ISSUE #5992: ssr unexpected hydration bailout with <pre> tag
### Version
2.3.4

### Reproduction
Here what happens on F5 (reload) key:
ssr_pre_tag_issue
(It happens only once, the video is looped)

### Steps to reproduce
Add a `<pre>` tag in App.vue of the vue-hackernews-2.0 demo:

```
<template>
  <div id="app">
    <header class="header">
    <pre style="background-color: #aaa">
foobar
    </pre>
      <nav class="inner">

```

### What is expected?
client-side and server-side have the same version for the `<pre>` tag

client and server side versions are not the same, then bailout ?, then `<pre>` tag content is flickering

defcc:

- [X] Bugfix
- [ ] Feature
- [X] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
fix #5992

### DIFF
--- a/src/compiler/parser/html-parser.js
+++ b/src/compiler/parser/html-parser.js
@@ -61,7 +61,7 @@ const encodedAttrWithNewLines = /&(?:lt|gt|quot|amp|#10);/g
 
 // #5992
 const ignoreFirstLFTagList = makeMap('pre,textarea', true)
-const isIgnoreFirstLf = (tag, html) => tag && ignoreFirstLFTagList(tag.toLowerCase()) && html[0] === '\n'
+const isIgnoreFirstLf = (tag, html) => tag && ignoreFirstLFTagList(tag) && html[0] === '\n'
 
 function decodeAttr (value, shouldDecodeNewlines) {
   const re = shouldDecodeNewlines ? encodedAttrWithNewLines : encodedAttr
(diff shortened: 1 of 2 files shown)


src/compiler/parser/html-parser.js
```diff
@@ -59,6 +59,10 @@ const decodingMap = {
 const encodedAttr = /&(?:lt|gt|quot|amp);/g
 const encodedAttrWithNewLines = /&(?:lt|gt|quot|amp|#10);/g
 
+// #5992
+const ignoreFirstLFTagList = makeMap('pre,textarea', true)
+const isIgnoreFirstLf = (tag, html) => tag && ignoreFirstLFTagList(tag.toLowerCase()) && html[0] === '\n'
```
javoski: Not need to call `toLowerCase()` here, and it seems `<textarea>` will never get into this function since `isPlainTextElement('textarea')` equals `true`.

defcc: 1. html5 tag names are case-insensitiv, so we use lower case to compare
2. isIgnoreFirstLf will be invoked when in textarea parse https://github.com/vuejs/vue/pull/6022/files#diff-084898b8c9f6d4d2833b18312180ef8fR172

javoski: 1. It'll call `toLowerCase()` when passing true to `makeMap` as the second param.
2. I didn't notice those changes starting from line 172, my fault😂.

defcc: 👍  Thanks, I missed it, I'll update.

defcc: Done.

defcc:
The naming is much better, thanks :laughing:

PR #6025, opened 2017-07-04
Fix #6008, Add merge strategy for provide option

### LINKED ISSUE #6008: Provide options not merged from mixins
### Version
2.3.4

### Reproduction link
[https://jsfiddle.net/jason_sanjose/00h2moa2/](https://jsfiddle.net/jason_sanjose/00h2moa2/)

### Steps to reproduce
* Define a component A that uses 2 mixins that each define their own `provide` properties, e.g. `foo` and `bar`. 
* Define a component B inside A that uses `inject` to add `foo` and `bar` properties

### What is expected?
Both `foo` and `bar` should be available in component B.

The `provide` defined by the last mixin is the only one available. 

Looks like the standard options merging only accounts for `inject` but not `provide` https://github.com/vuejs/vue/blob/dev/src/core/util/options.js#L185.

---
Update docs if this behavior is expected.

znck:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -63,7 +63,7 @@ function mergeData (to: Object, from: ?Object): Object {
 /**
  * Data
  */
-strats.data = function (
+export function mergeDataOrFn (
   parentVal: any,
   childVal: any,
   vm?: Component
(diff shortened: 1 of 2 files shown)


src/core/util/options.js
```diff
@@ -183,6 +183,7 @@ strats.watch = function (parentVal: ?Object, childVal: ?Object): ?Object {
 strats.props =
 strats.methods =
 strats.inject =
+strats.provide =
```
posva: This is actually not enough because `provide` can also be a function that returns an object.

I think we should make `data` strategy reusable and use it for `provide`
Other plugins like vuefire will also benefit from this 🙂

znck: 😝 I'd add a failing test first.

znck:
Hey @posva! I guess I didn't miss anything this time.

PR #6039, opened 2017-07-06
fix #5997: refactor function registerRef

### LINKED ISSUE #5997: Refs incorrect when using v-for and combining components and elements
### Version
2.3.4

### Reproduction link
[https://jsfiddle.net/floorish/7td2offb/](https://jsfiddle.net/floorish/7td2offb/)

### Steps to reproduce
- Parent with dynamic components, referenced by `ref`: `<component :is="child.type"  v-for="child in children" :key="child.id" ref="children"></component>`
- Multiple `child.type`s where one of them uses a loading component: `<loading v-if="isLoading"></loading> <div v-else>Child B ready</div>`
- In Child switch `isLoading` from `true` to `false` after a timeout/async call

### What is expected?
`parent.$refs.children` equals the displayed child components

`parent.$refs.children` only includes the child component that has the loading mechanism

---
This is very much an edge case, but it took me a while to figure out any solutions. Most trivial solution is to wrap the Child template in another `<div>`, but AFAIK the `<div v-if="val"></div> <div v-else></div>` construct in the root of a template is supported in Vue (there is always a single root element).

Other solutions are provided in reproduction link.

javoski:
fix #5997 .
When registering `ref`, if `refs[key]` is an array and `ref` is already in `refs[key]`, then do nothing.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/core/vdom/modules/ref.js
+++ b/src/core/vdom/modules/ref.js
@@ -32,10 +32,11 @@ export function registerRef (vnode: VNodeWithData, isRemoval: ?boolean) {
     }
   } else {
     if (vnode.data.refInFor) {
-      if (Array.isArray(refs[key]) && refs[key].indexOf(ref) < 0) {
-        refs[key].push(ref)
-      } else {
+      if (!Array.isArray(refs[key])) {
         refs[key] = [ref]
+      } else if (refs[key].indexOf(ref) < 0) {
+        // $flow-disable-line
+        refs[key].push(ref)
       }
     } else {
       refs[key] = ref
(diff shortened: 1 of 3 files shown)


src/core/vdom/modules/ref.js
```diff
@@ -32,10 +32,11 @@ export function registerRef (vnode: VNodeWithData, isRemoval: ?boolean) {
...
   } else {
     if (vnode.data.refInFor) {
-      if (Array.isArray(refs[key]) && refs[key].indexOf(ref) < 0) {
-        refs[key].push(ref)
-      } else {
+      if (!Array.isArray(refs[key])) {
         refs[key] = [ref]
+      } else if (refs[key].indexOf(ref) < 0) {
+        // $flow-disable-line
+        refs[key].push(ref)
```
javoski: It's seems that `flow` is a little weak to check array type, even codes below couldn't pass.
```js
if (Array.isArray(refs[key])) {
  if (refs[key].indexOf(ref)) {
    refs[key].push(ref)
  }
}
```
So I added a suppress_comment option to `.flowconfig`.

Kingwl:
please add the test case😁

javoski:
Will do it soon
wenlu.wang <notifications@github.com>于2017年7月6日 周四19:12写道：

wuchuguang:
upgrade vue2.4.4
parent1
     child ref="child"
         loop child

parent1.$refs.child  is ok
parent2.$refs.child is undefined

PR #6060, opened 2017-07-09
fix(core): injections should not always reactive #(5913)

### LINKED ISSUE #5913: injected objects are reactive (despite documentation saying the opposite)
### Version
2.3.3

### Reproduction link
[https://jsfiddle.net/50wL7mdz/41000/](https://jsfiddle.net/50wL7mdz/41000/)

### Steps to reproduce
Provide (from parent component) an object with a pre-existing key. Inject that object into child component. Either console.log the object to see that it is reactive, or modify the value of the pre-existing key to see that the value also changes on screen automatically.

### What is expected?
According to vue's documentation, injected objects are not reactive: "Note: the provide and inject bindings are NOT reactive.". The value should not change on the screen after 3 seconds, and console.log should not display the word Getter when logging the injected object.

The injected object is reactive, and so the message changes on screen after 3 seconds, and console.log'ging the object shows it is reactive (displays the word Getter)

---
I use inject to inject commonly used libraries in objects, like moment, which I can then use in the templates to display nicely formatted dates. If the entire moment library object becomes reactive, then this will probably cause an unnecessary performance hit to the browser when using the moment library.

Kingwl:
fix #5913

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/inject.js
+++ b/src/core/instance/inject.js
@@ -1,8 +1,8 @@
 /* @flow */
 
 import { hasSymbol } from 'core/util/env'
-import { warn } from '../util/index'
-import { defineReactive } from '../observer/index'
+import { warn, defWithGetterSetter } from '../util/index'
+import { defineReactive, isObserver } from '../observer/index'
 import { hasOwn } from 'shared/util'
 
 export function initProvide (vm: Component) {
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks - see https://github.com/vuejs/vue/commit/8d66691ee264969447390822971b8caa029cac3e we can use the `observerState.shouldConvert` flag to disable recursive conversion when using `defineReactive` (also used when initializing props).

PR #6072, opened 2017-07-11
perf: remove src directory from npm module

KELiON:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Breaking change is only for users, who uses undocumented requires/imports from `vue/src`. After this PR all scripts from `src` folder are not included in npm-package

**Other information:**

`src` directory is not used by node package. Removing this directory reduces size of `.tgz` package by 112kb and saves 840kb in each node project, where vue.js is installed

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
   "unpkg": "dist/vue.js",
   "typings": "types/index.d.ts",
   "files": [
-    "src",
     "dist/*.js",
     "types/*.d.ts"
   ],


andreiglingeanu:
Agreed, almost everyone who want to build Vue themselves are using `vue/dist/vue.esm.js` outputs

rigor789:
Hi, this change is a good performance improvement, however it is a breaking change for https://github.com/rigor789/nativescript-vue 

Due to the nature of `nativescript-vue` we have to use some code from the `src` directory. 

I am here to ask if there is a better approach to our problem, than importing directly from `src`? 

As for this change, I have tried installing directly through GitHub, using either a tag, or a commit hash, but in all cases I got the version without the `src` directory.  Do you have any suggestions how to move forward with this change?

yyx990803:
Ah, yeah I realized that problem. Will revert it in a later release.

PR #6098, opened 2017-07-13
fix #6093, inject should not be merged with Object.create

### LINKED ISSUE #6093: Inject option in mixin no longer works
### Version
2.4.1

### Reproduction link
[https://jsfiddle.net/d92sL0zy/2/](https://jsfiddle.net/d92sL0zy/2/)

### Steps to reproduce
1. Create mixin that sets the `inject` option
2. Create component that uses said mixin
3. Use component in context with a corresponding `provide`

### What is expected?
The data provided in the context is correctly injected into the component that uses the mixin with the `inject` option set.

The data provided is _not_ injected into the component that uses the mixin, unless the `inject` option is specified directly on the component and not in the mixin.

---
I use the pattern of specifying `provide` and `inject` options on mixins extensively, as this provides some really useful [mixins for specifying relationships between different components](https://github.com/kasperisager/vue-relation). A component using a `hasOne()` mixin would for example `provide` itself to a child which would then use a `memberOf()` mixin to `inject` the parent. As of Vue 2.4.0 , this however no longer works as the `inject` option of the `memberOf()` mixin is ignored.

jkzing:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ x When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**
fix [https://github.com/vuejs/vue/issues/6093](https://github.com/vuejs/vue/issues/6093)

`Object.create` is not adding 'own property' to the object, which caused injections not injected properly.😭

### DIFF
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -195,11 +195,16 @@ strats.watch = function (parentVal: ?Object, childVal: ?Object): ?Object {
  */
 strats.props =
 strats.methods =
-strats.inject =
 strats.computed = function (parentVal: ?Object, childVal: ?Object): ?Object {
   if (!childVal) return Object.create(parentVal || null)
   if (!parentVal) return childVal
   const ret = Object.create(null)
+  extend(ret, parentVal)
+  extend(ret, childVal)
+  return ret
+}
+strats.inject = function (parentVal: ?Object, childVal: ?Object): ?Object {
+  const ret = {}
   extend(ret, parentVal)
   extend(ret, childVal)
   return ret
--- a/test/unit/features/options/inject.spec.js
+++ b/test/unit/features/options/inject.spec.js
@@ -442,4 +442,25 @@ describe('Options provide/inject', () => {
     expect(isObserver(child.bar)).toBe(false)
     expect(isObserver(child.baz)).toBe(false)
   })
+
+  // #6093
+  it('should merge inject from mixins properly', () => {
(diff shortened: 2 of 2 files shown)


javoski:
Thanks for the PR.
Actually there's no need to add another merge strategy function, it's an issue about using `Object.keys()` to iterate inject object while using `for...in` on other options. (`Object.keys()` will not enumerates properties in the prototype chain, but `for...in` does).

So my suggestion is:
```js
strats.props =
strats.methods =
strats.inject =
strats.computed = function (parentVal: ?Object, childVal: ?Object): ?Object {
  if (!parentVal) return childVal
  const ret = Object.create(null)
  extend(ret, parentVal)
  if (childVal) extend(ret, childVal)
  return ret
}
```
Just delete this line will fix #6093 :
```js
if (!childVal) return Object.create(parentVal || null)
```

jkzing:
@javoski You are right, I did think about that too. But I just didn't have the confident to remove that line directly. (I was afraid it's using `Object.create` for a reason)

Anyway, If it works totally OK without `Object.create`, it's better to merge your PR, and feel free to close this one when merging it.

yyx990803:
Thanks, #6093 seems to be a simpler fix.

PR #6194, opened 2017-07-23
fix #6193, change event should not be triggered if each value still matches an option

### LINKED ISSUE #6193: Select with v-model triggers change event when its options are changed
### Version
2.4.2

### Reproduction link
[https://jsfiddle.net/floorish/e2h2vwc2/](https://jsfiddle.net/floorish/e2h2vwc2/)

### Steps to reproduce
Dynamically change the `<option>`s in a `<select v-model="val">` tag and the `v-model` triggers a change event, but the actually chosen option hasn't changed.

### What is expected?
`v-model` does not trigger a change event

`v-model` triggers a change event

---
Change of behaviour intruduced in https://github.com/vuejs/vue/commit/c90b140e80f1bd5d01c733a5bd9fa6e9cb3c2b4d

jkzing:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
[https://github.com/vuejs/vue/issues/6193](https://github.com/vuejs/vue/issues/6193)

---
It seems we still need the value matching check, only detect options change is not enough.

### DIFF
--- a/src/platforms/web/runtime/directives/model.js
+++ b/src/platforms/web/runtime/directives/model.js
@@ -60,7 +60,14 @@ export default {
       const prevOptions = el._vOptions
       const curOptions = el._vOptions = [].map.call(el.options, getValue)
       if (curOptions.some((o, i) => !looseEqual(o, prevOptions[i]))) {
-        trigger(el, 'change')
+        // trigger change event if
+        // no matching option found for at least one value
+        const needReset = el.multiple
+          ? binding.value.some(v => hasNoMatchingOption(v, el.options))
+          : binding.value !== binding.oldValue && hasNoMatchingOption(binding.value, el.options)
+        if (needReset) {
+          trigger(el, 'change')
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/platforms/web/runtime/directives/model.js
```diff
@@ -101,6 +108,15 @@ function setSelected (el, binding, vm) {
   }
 }
 
+function hasNoMatchingOption (value, options) {
```
Kingwl: how about Array.prototype.some😅

jkzing: Yes, it would be better to use [].some. 
Actually I just did revert changes made in [https://github.com/vuejs/vue/commit/c90b140e80f1bd5d01c733a5bd9fa6e9cb3c2b4d](https://github.com/vuejs/vue/commit/c90b140e80f1bd5d01c733a5bd9fa6e9cb3c2b4d), and didn't think about it. 😅
Will do it later.

cgarnier:
Is it released ?

LinusBorg:
It hasn't been merged yet, so: no.

PR #6244, opened 2017-07-28
close #6241 : support .sync modifier on prop object properties in deep

### LINKED ISSUE #6241: Sync Modifier support object properties
Sync Modifier support object properties. 
For example, I have a myObj object, it has a foo and a bar property.
```javascript
myObj: {
  foo: 'foo',
  bar: 'bar'
}
```
This feature could use `update:myObj.foo` and `update:myObj.bar` to update the value.
```javascript
this.$emit('update:myObj.foo', 'new-foo');
this.$emit('update:myObj.bar', 'new-bar');
```

Currently, I need to use below format
```html
<MyComponent :foo.sync="myObj.foo" :bar.sync="myObj.bar"></MyComponent>
```
```javascript
props: ['foo', 'bar']
...
this.$emit('update:foo', 'new-foo');
this.$emit('update:bar', 'new-bar');
```
But with this new feature:
```html
<MyComponent :myObj.sync="myObj"></MyComponent>
```
```javascript
props: ['myObj']
...
this.$emit('update:myObj.foo', 'new-foo');
this.$emit('update:myObj.bar', 'new-bar');
```
It becomes more simple and clear, especially when the object has many properties.

javoski:
close #6241 .

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/core/instance/events.js
+++ b/src/core/instance/events.js
@@ -5,7 +5,8 @@ import {
   toArray,
   hyphenate,
   handleError,
-  formatComponentName
+  formatComponentName,
+  parsePath
 } from '../util/index'
 import { updateListeners } from '../vdom/helpers/index'
 
@@ -20,6 +21,8 @@ export function initEvents (vm: Component) {
 }
 
 let target: Component
+
+const updateEventRE = /^update:([\w.$]+)\.([\w$]+)$/
 
 function add (event, fn, once) {
   if (once) {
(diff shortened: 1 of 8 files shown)


src/core/instance/events.js
```diff
@@ -21,6 +22,8 @@ export function initEvents (vm: Component) {
 
 let target: Component
 
+const updateEventRE = /^update:([\w.$]+)\.([\w$]+)$/
```
posva: Should we do `^update:([\w.$]+[^.])\.([\w$]+)$` to prevent things like `update:foo..bar`

posva: hmm, nvm, it doesn't seem to be necessary

jpsouzasilva: @javoski  `update:.foo.bar` would still pass through the regex and produce the same matching length

isn't it undesirable, since you're getting the first split chunk? removing the first dot from your expression would deal with this

javoski: @jpsouzasilva Thanks for pointing this out, actually the event `'update:.foo.bar'` will not get into the Regex validation process unless there is an event named `'update:'` exists in the instance. However, we can modify the regex to `/^update:([\w$][\w.$]+)\.([\w$]+)$/` to make it more robust.

yyx990803:
See https://github.com/vuejs/vue/issues/6241#issuecomment-326688288

PR #6322, opened 2017-08-08
close #6097: Allow defining optional inject dependency with default values

### LINKED ISSUE #6097: Allow defining optional inject dependency with default values
In case of components that can work both as standalone and as children, it would be nice to be able to provide default values for the not provided dependencies.
It would also avoid the warning for missing inject in case this is one of the expected use case.

inject: Array<string> | { [key: string]: string | Symbol | { name: string | Symbol, default: any } }

pdanpdan:
Allow to define inject as object with name and default properties. The default is used if no provide
is found, and no warning is emitted.

close #6097

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

In case of components that can work both as standalone and as children, it would be nice to be able to provide default values for the not provided dependencies.
It would also avoid the warning for missing inject in case this is one of the expected use case.

inject: Array | { [key: string]: string | Symbol | { name?: string | Symbol, default?: any } }

### DIFF
--- a/src/core/instance/inject.js
+++ b/src/core/instance/inject.js
@@ -49,7 +49,7 @@ export function resolveInject (inject: any, vm: Component): ?Object {
 
     for (let i = 0; i < keys.length; i++) {
       const key = keys[i]
-      const provideKey = inject[key]
+      const provideKey = inject[key].name
       let source = vm
       while (source) {
         if (source._provided && provideKey in source._provided) {
@@ -58,8 +58,15 @@ export function resolveInject (inject: any, vm: Component): ?Object {
         }
         source = source.$parent
       }
-      if (process.env.NODE_ENV !== 'production' && !source) {
-        warn(`Injection "${key}" not found`, vm)
+      if (!source) {
+        if ('default' in inject[key]) {
+          const provideDefault = inject[key].default
+          result[key] = typeof provideDefault === 'function'
+            ? provideDefault.call(vm)
+            : provideDefault
+        } else if (process.env.NODE_ENV !== 'production') {
+          warn(`Injection "${key}" not found`, vm)
+        }
       }
     }
     return result
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -270,12 +270,21 @@ function normalizeProps (options: Object) {
  */
 function normalizeInject (options: Object) {
   const inject = options.inject
+  const normalized = {}
+  let val
   if (Array.isArray(inject)) {
-    const normalized = options.inject = {}
     for (let i = 0; i < inject.length; i++) {
-      normalized[inject[i]] = inject[i]
+      normalized[inject[i]] = { name: inject[i] }
+    }
+  } else if (isPlainObject(inject)) {
+    for (const key in inject) {
+      val = inject[key]
+      normalized[key] = isPlainObject(val)
+        ? extend({ name: key }, val)
+        : { name: val }
     }
   }
+  options.inject = normalized
 }
 
 /**
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -270,21 +270,19 @@ function normalizeProps (options: Object) {
  */
 function normalizeInject (options: Object) {
   const inject = options.inject
-  const normalized = {}
-  let val
+  const normalized = options.inject = {}
   if (Array.isArray(inject)) {
     for (let i = 0; i < inject.length; i++) {
       normalized[inject[i]] = { name: inject[i] }
     }
   } else if (isPlainObject(inject)) {
     for (const key in inject) {
-      val = inject[key]
+      const val = inject[key]
       normalized[key] = isPlainObject(val)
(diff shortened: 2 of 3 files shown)


src/core/util/options.js
```diff
@@ -270,12 +270,21 @@ function normalizeProps (options: Object) {
  */
 function normalizeInject (options: Object) {
   const inject = options.inject
+  const normalized = {}
+  let val
```
posva: no need to declare val here

src/core/util/options.js
```diff
@@ -270,12 +270,21 @@ function normalizeProps (options: Object) {
...
   if (Array.isArray(inject)) {
-    const normalized = options.inject = {}
     for (let i = 0; i < inject.length; i++) {
-      normalized[inject[i]] = inject[i]
+      normalized[inject[i]] = { name: inject[i] }
+    }
+  } else if (isPlainObject(inject)) {
+    for (const key in inject) {
+      val = inject[key]
```
posva: Can be a const

test/unit/features/options/inject.spec.js
```diff
@@ -362,6 +362,23 @@ describe('Options provide/inject', () => {
     expect(`Injection "baz" not found`).not.toHaveBeenWarned()
   })
 
+  // Github issue #6097
+  it('should not warn when injections cannot be found but have default value', () => {
+    const vm = new Vue({})
```
posva: Can you add also an injection with a value and check that it takes precedence?

src/core/util/options.js
```diff
@@ -270,12 +270,21 @@ function normalizeProps (options: Object) {
  */
 function normalizeInject (options: Object) {
   const inject = options.inject
+  const normalized = {}
+  let val
   if (Array.isArray(inject)) {
-    const normalized = options.inject = {}
```
posva: you can keep this line (and move it outside of the if) and remove the last line `options.inject = normalized`

pdanpdan:
Fixed. I added a separate test for provided value taking precedence over default.

posva:
Thanks, we'll wait for @yyx990803  to review it as well 🙂 
Others may chime in in the meantime as well

src/core/instance/inject.js
```diff
@@ -55,8 +55,15 @@ export function resolveInject (inject: any, vm: Component): ?Object {
         }
         source = source.$parent
       }
-      if (process.env.NODE_ENV !== 'production' && !source) {
-        warn(`Injection "${key}" not found`, vm)
+      if (!source) {
```
jkzing: Isn't `if (!result[key])` easier to understand here?

pdanpdan:
No, it's not the same. You don't want to check the value of the injection, but it's presence. It would work `!key in result`

jkzing:
@pdanpdan Oh, yes I missed falsy values. NVM

rstoenescu:
@yyx990803 Hi, any chance you can take a look at this? Quasar Framework developers cannot upgrade to Vue 2.4.x until this PR gets in. Thanks!

ariesjia:
waiting for this RP for upgrade to 2.4

guyulmaz:
Waiting this PR

brucelane:
waiting for this PR to upgrade to 2.4, please merge...

posva:
By chime in, I meant to give further feedback 😉 , upvoting a comment is enough to say _+1_, no need to comment _I need this to upgrade_ , _+1_ or similar

I don't remember seeing weex files getting updated, any idea why that weex PR got in the git history of this PR @pdanpdan ?

pdanpdan:
I don't understand where did the weex commit come from. I'm on vacation till next Monday and I'll have to wait till then to get to a computer.

javoski:
Seems every PR are in the similar situation.

pdanpdan:
I just changed the PR to exclude the weex commit.

src/core/instance/inject.js
```diff
@@ -49,7 +49,7 @@ export function resolveInject (inject: any, vm: Component): ?Object {
 
     for (let i = 0; i < keys.length; i++) {
       const key = keys[i]
-      const provideKey = inject[key]
+      const provideKey = inject[key].name
```
hirokiosame: I'm speculating this will introduce a breaking change for cases where `inject` is an object where the key is the property the injection should be assigned to and the value is the key:
```
const s = Symbol()

const Provider = {
  provide () {
    return {
      [s]: 'foo'
    }
  }
}

const Child = {
  inject: {
    fromProvider: s
  },
  created() {
  	console.log('This is from the provider:', this.fromProvider);
  },
  // ...
}
```

I believe you should be checking if `inject[key]` is a String or Symbol before accessing the `name` property.

PR #6327, opened 2017-08-08
Allow v-bind="$attrs" with v-on="$listeners" to work with v-model

chrisvfritz:

- [x] Bugfix?
- [x] Feature

- [x] No, unless someone was using `v-model` to purposefully bind an element's value directly to a DOM Event, which I think would be strange.

I brought this up in #6216 a couple weeks ago and it didn't encounter any opposition, so here we go! 

__TLDR__: Since `v-bind="$attrs"` and `v-on="$listeners"` were added, a lot of people have been assuming that combining these should be sufficient to allow `v-model` on the component. This PR makes `v-model` work as users expect.


chrisvfritz:
I just noticed some Weex tests are failing. I could easily make them pass, but due to the nature of the tests and my own lack of experience with Weex, I'm not confident this would actually guarantee working behavior. 

@Jinjiang Would you mind taking a look? Is it safe to update the tests, or is it possible my changes would break Weex?

javoski:
Seems `$event.target.value` will break weex, browser-based code shouldn't exist in cross-platform part.

chrisvfritz:
@javoski Are you sure this would break Weex? I only try to access `$event.target.value` if the `$event` is an object and `$event.type === 'input'`. In cases where a DOM Event (or something resembling a DOM Event) isn't attached to the component event, Weex would see the normal behavior.

Jinjiang:
@chrisvfritz sorry it's still loading status in "$ bash build/ci.sh" part of the ci report (maybe my network problem). I will have a look in the deep. Just a moment please :-)

Jinjiang:
@chrisvfritz This commit is fine. Currently you can just skip these failed test cases in weex. And @Hanks10100 and I will fix them later.
Thanks.

Hanks10100:
@chrisvfritz After I confirmed that this feature will not destroy weex, I will modify those failed test cases and add the corresponding case in Weex.

Hanks10100:
@chrisvfritz  I fixed the test case in Weex and send a PR (https://github.com/chrisvfritz/vue/pull/1) to your branch. The CI is passed.

Your modification didn't break Weex, but not consider the difference of it.

As I write in that PR:

> In Weex, the structure of the `<input>` element has a slight difference with Web, it doesn't have the `value` property but has `attr.value`.

SergejKasper:
Can this be merged? It would be a great help for wrapping inputs!

yyx990803:
See https://github.com/vuejs/vue/issues/6216#issuecomment-333648093

chrisvfritz:
Yeah, I agree it felt a little hacky and certainly didn't cover every use case. Part of me was hoping you'd think of some small refactor that would simplify things, but we weren't lucky this time. 🙂

PR #6367, opened 2017-08-14
Uses reactive set to change a synced property

sunyatasattva:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

Small improvement

- [ ] Yes
- [x] No

**Other information:**

I'm using the reactive setter to set a property with the `sync` modifier. My case is that since the `sync` modifier is basically just syntactic sugar to avoid some boilerplate, it is useful for most case to set a property through `Vue.set`.

I might be wrong, but instead of opening an issue I thought doing a pull request would be faster, since the change is just a line of code.

### DIFF
--- a/build/release-weex.sh
+++ b/build/release-weex.sh
@@ -31,7 +31,6 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
   cd -
 
   # commit
-  git add src/entries/weex*
   git add packages/weex*
   git commit -m "[release] weex-vue-framework@$NEXT_VERSION"
 fi
--- a/packages/weex-template-compiler/build.js
+++ b/packages/weex-template-compiler/build.js
@@ -2,12 +2,16 @@
 
 Object.defineProperty(exports, '__esModule', { value: true });
 
-var he = require('he');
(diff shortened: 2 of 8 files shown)


posva:
Hello thanks for your PR but using `Vue.set` has an extra cost because it forces Vue to add watchers whereas doing a plain assignment don't. This is intentional, remember every data property should be declared in `data`

sunyatasattva:
Hey! Thanks for answering. I was just wondering if basically anything with `.sync` isn't practically expected to be reactive, and that would save some boilerplate.

But if you disagree, feel free to close of course :)

posva:
Sync is for two way binding. You shouldn't do Vue. Set everyone to se something, it's for Vey specific cases. Look for caveat in the documentation. That would hopefully clarify many things

yyx990803:
This seems unnecessary to me, since the event listener would only be created if the key already exists in that object. This would only apply when the key exists and is not already reactive, which sounds like quite an unlikely case for `v-bind` object. Closing but feel free to provide a use case where this would prove necessary.

PR #6450, opened 2017-08-25
close #6447: fix type checking for primitive wrapper objects

### LINKED ISSUE #6447: Change prop type checking to use instanceof instead of typeof
Currently prop type checking fails when passing in an object that extends the expected type.
This would allow anything inheriting from the expected type to pass the check.

https://github.com/vuejs/vue/blob/8d66691ee264969447390822971b8caa029cac3e/src/core/util/props.js#L148

If this line were
```javascript
valid = value instanceof expectedType.toLowerCase()
```

Then child classes would also pass. In my mind if a component is checking for `String` and I pass in an object that extends string this should pass the typecheck.

This was born out of an issue that one of my users has passing an String castable object which extends string as a prop into a Vue component.

If you're interested you can check that out here: 

https://github.com/tightenco/ziggy/pull/64

javoski:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
close #6447 .

### DIFF
--- a/build/release-weex.sh
+++ b/build/release-weex.sh
@@ -31,7 +31,6 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
   cd -
 
   # commit
-  git add src/entries/weex*
   git add packages/weex*
   git commit -m "[release] weex-vue-framework@$NEXT_VERSION"
 fi
--- a/packages/weex-template-compiler/build.js
+++ b/packages/weex-template-compiler/build.js
@@ -2,12 +2,16 @@
 
 Object.defineProperty(exports, '__esModule', { value: true });
 
-var he = require('he');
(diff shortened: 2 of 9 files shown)


DanielCoulbourne:
I strongly support this PR. (And thanks for making it. I would have done myself but you beat me to it 😄) It solves a pretty big use case for a few users of my package Ziggy as outlined in https://github.com/tightenco/ziggy/pull/64

JesseLeite:
^ I'd like to note that this PR has way more upvotes than downvotes.

yyx990803:
Merging since this is a somewhat implicit change and we probably will not document this behavior so that we don't "encourage code smell" for the users :)

PR #6453, opened 2017-08-25
fixes #6361: check if component is custom

### LINKED ISSUE #6361: <input :is="textarea"> broken
### Version
2.4.2

### Reproduction link
[https://jsfiddle.net/pi0/qnvcgujb](https://jsfiddle.net/pi0/qnvcgujb)

### Steps to reproduce
This is working with 2.3 but not 2.4 anymore: (see provided fiddles)
```html
<input v-model="value" is="textarea"/>
```

### What is expected?
This was the behavior of 2.3.x:

https://jsfiddle.net/pi0/yL4rwesh

When using `:is="textarea"` to dynamically change input type to a text area neither `:value` or `v-model` are not working anymore. (There is also a common case for both which `<component :is="..."` is not working) 

---
Using dynamic components is a common usage for component libraries like BootrapVue which user can dynamically change elements behavior. [here](https://github.com/bootstrap-vue/bootstrap-vue/blob/v0.21.0/lib/components/form-input.vue#L4) is the current implementation we are falling back to `v-if`/`v-else` as a fix, But is it an intentional breaking change?

cwang22:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
fixes #6361

### DIFF
--- a/src/platforms/web/compiler/directives/model.js
+++ b/src/platforms/web/compiler/directives/model.js
@@ -40,7 +40,7 @@ export default function model (
     }
   }
 
-  if (el.component) {
+  if (el.component && el.component !== '"input"' && el.component !== '"select"' && el.component !== '"textarea"') {
     genComponentModel(el, value, modifiers)
     // component v-model doesn't need extra runtime
     return false
--- a/src/platforms/web/compiler/directives/model.js
+++ b/src/platforms/web/compiler/directives/model.js
@@ -40,7 +40,7 @@ export default function model (
     }
   }
 
-  if (el.component && !isInputTag(el.attrsMap.is)) {
+  if (el.component && (!el.attrsMap.is || !isInputTag(el.attrsMap.is))) {
     genComponentModel(el, value, modifiers)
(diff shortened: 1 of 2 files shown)


src/platforms/web/compiler/directives/model.js
```diff
@@ -40,7 +40,7 @@ export default function model (
     }
   }
 
-  if (el.component) {
+  if (el.component && (!el.attrsMap.is || !isInputTag(el.attrsMap.is))) {
```
posva: I wonder if we shouldn't use ` && !isInputTag(el.attrsMap.is)` and call `makeMap` with `false` (or nothing). If someone uses an capitalized name for the component, like `Textarea` or `Input`, they probably want to use a component instead of the HTML native tag

yyx990803:
Thanks for the PR! See https://github.com/vuejs/vue/issues/6361#issuecomment-327319470

PR #6622, opened 2017-09-16
WIP: implement Weex native directive compiler for <recycle-list>

Hanks10100:
+ [x] `v-bind` & `:`
+ [x] `v-if`
+ [x] `v-else`
+ [x] `v-else-if`
+ [x] `v-for`
+ [ ] `v-on` & `@`

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -72,8 +72,8 @@ declare type ASTNode = ASTElement | ASTText | ASTExpression;
 declare type ASTElement = {
   type: 1;
   tag: string;
-  attrsList: Array<{ name: string; value: string }>;
-  attrsMap: { [key: string]: string | null };
+  attrsList: Array<{ name: string; value: any }>;
+  attrsMap: { [key: string]: any };
   parent: ASTElement | void;
   children: Array<ASTNode>;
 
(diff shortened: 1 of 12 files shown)


Hanks10100:
In order to support the binding arguments of the inline event handler, I think the binding argument list should also pass to Weex runtime.

The expected compilation results are as follows.

**1. Normal handler** (no change).

```html
<div @click="doSomething"></div>
```

```js
on: {
  click: vm.doSomething
}
```

**2. Inline handler**.

```html
<div @click="doSomething(index, 'xxx')"></div>
```

```js
on: {
  click: {
    handler: function (index, 'xxx', $event) {
      vm.doSomething(index, 'xxx')
    },
    params: [{ '@binding': 'index' }, 'xxx']
  }
}
```

**3. Inline handler with `$event`**.

```html
<div @click="doSomething(index, $event, 35)"></div>
```

```js
on: {
  click: {
    handler: function (index, $event, 35) {
      vm.doSomething(index, $event, 35)
    },
    params: [{ '@binding': 'index' }, { '@binding': '$event' }, 35]
  },
}
```

The logic of binding events also needs to be modified. I have a try: https://github.com/vuejs/vue/compare/dev...Hanks10100:weex-event

PR #6631, opened 2017-09-17
add hasCrashed function to determin whether the code has minified or not

loatheb:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Sometimes I minified the code without `NODE_ENV=production`, It still warned me `You are running Vue in development mode.`, I don't think it's rigorous enough.

To determine whether the code runs in the development mode or in production mode, I think it should be judged from the behavior of the code, not just from the `NODE_ENV` flag when building.

So I add `function hasCrashed () {}`, when someone minified the code without flag. But the code here will know, "Oh, You are No Longer Development!"

Also, I find the [doc](https://vuejs.org/v2/guide/deployment.html) of VUE are still using the vue.js uncompressed version. 

(ง •̀_•́)ง

### DIFF
--- a/src/platforms/web/runtime/index.js
+++ b/src/platforms/web/runtime/index.js
@@ -42,6 +42,12 @@ Vue.prototype.$mount = function (
   return mountComponent(this, el, hydrating)
 }
 
+/*
+  determin wheather the code has minified
+  if the code has minified, then `hasCrashed.name !== 'hasCrashed'`
+*/
+function hasCrashed () {}
+
 // devtools global hook
 /* istanbul ignore next */
 setTimeout(() => {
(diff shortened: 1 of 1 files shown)


posva:
You can remove the production tip by setting an option, but as you can imagine, we cannot add that hack to check if the code is minified.
Using the non-minified version is intentional to allow using the devtools to play around in the docs

FYI, if you minify the code without `NODE_ENV=production`, you'll end up with a bigger bundle

yyx990803:
Development mode simply means all the extra runtime checks and warnings are still in your bundle - even if it's minified, it is still technically in development mode. If you want to deploy for production, **you should never minify your code without `NODE_ENV=production`**. The fact that the warning is still there actually helps you discover the problem.

Issue #5415, opened 2017-04-11
Provide Server-Side-Rendering helper using more purely JavaScript


denghongcai:
### What problem does this feature solve?
Current Server-Side-Rendering implementation is bond to Node.js, makes it hard to use in Nashorn (lightweight high-performance JavaScript runtime in Java with a native JVM) or other JavaScript runtime.

Foundations needed by Vue like `vm` and `fs` modules may be passed into renderer or runner as parameter which implements some interfaces.

posva:
Added a note about Nashorn

wojtask9:
I would love to see that happen.

I'm trying to use Vue.js with nashorn too.
It will be great if Vue.js SSR added interfaces such FileSystem or Engine to externalize 
some functionalities that are bind to node.js

I've made some research:

1. `path` in all places can be abstracted 
currently used `path.extname`, `path.join`, `path.isAbsolute`, `path.dirname`

2. `fs` in all places can be abstracted too
currently used `fs.existsSync`, `fs.readFileSync`

3. `module` 
only one usage NativeModule.wrap
According to docs and soruce code this only wrap script code with
```
(function (exports, require, module, __filename, __dirname) {
// Your module code actually lives in here
});
```
so this is can be replaced with function if we are not on node.js

4. `resolve` , `vm` used in create-bundle-runner.js
This file can be somehow abstracted :) not sure how and what interface could look

5. `stream`, `source-map-support`  TODO  :)

I made small modifications in source code to get rid of `stream` and `path`
my entry:
```
//import 'core-js/modules/es6.array.find.js';
//import 'core-js/modules/es6.object.assign.js';
//import 'nashorn-polyfill'
//import 'event-loop'
import  Vue from './web-runtime-with-compiler'
export { Vue }
export { createRenderer/*, createBundleRenderer*/} from './web-server-renderer'
```

result sizes: 
1.  only createRenderer() and removed path and stream requires (commented stream functions and replaced path with own simple implementation)
after browserify -> 430KB
2.  only createRenderer() (commented out bundleRenderer form entry)
after browserify run -> 580KB
3. createRenderer() and createBundleRenderer() (not modified at all)
after browserify -> 700KB

denghongcai:
@wojtask9 suprised to see you go further! Would you mind share us a demo code?

wojtask9:
Yes of course. I can share my code. But you must be patient (because of Easter Holiday)
JS part I'll try share tomorrow evening and java part probably on Tuesday.

[...]

PR #6651, opened 2017-09-20
ssr: switch to iterative rendering and don't use process.nextTick in basic renerer (RFC)

wojtask9:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

From comment https://github.com/vuejs/vue/issues/5415#issuecomment-327276515
" in 2.4 we already shipped a pure js build of vue-server-renderer, but we haven't announced it yet. It only supports the base renderer, but is decoupled from Node.js APIs"

That statement is not correct. `basic.js` still use `process.nextTick()` when recursion stack is to high.
With this PR switching to iterative version we can avoid `process.nextTick()` path.
Ideally `createBasicRenderer` should use different `write` function but currently this PR is "Request For Comments"

Also iterative version passes all SSR tests (not only basic if we switch to full server-renderer) and helps a lot for not Node.js env (in my case Nashorn)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,6 +8,7 @@
     "plugin:flowtype/recommended"
   ],
   "globals": {
-    "__WEEX__": true
+    "__WEEX__": true,
+    "__BASIC_RENDERER__": false
   }
 }
--- a/build/config.js
+++ b/build/config.js
@@ -120,6 +120,7 @@ const builds = {
     format: 'umd',
     env: 'development',
     moduleName: 'renderVueComponentToString',
+    basicRenderer: 'true',
     plugins: [node(), cjs()]
   },
   'web-server-renderer-webpack-server-plugin': {
@@ -169,6 +170,7 @@ function genConfig (opts) {
     moduleName: opts.moduleName || 'Vue',
     plugins: [
       replace({
+        __BASIC_RENDERER__: !!opts.basicRenderer,
         __WEEX__: !!opts.weex,
         __WEEX_VERSION__: weexVersion,
         __VERSION__: version
--- a/src/platforms/web/entry-server-basic-renderer.js
+++ b/src/platforms/web/entry-server-basic-renderer.js
@@ -1,5 +1,7 @@
 /* @flow */
 
+process.env.BASIC_RENDERER = 'true'
+
 import modules from './server/modules/index'
 import directives from './server/directives/index'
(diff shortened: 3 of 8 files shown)


yyx990803:
Thanks! This is indeed something we overlooked. The basic renderer was still a proof of concept which is why we didn't include it in the 2.4 release notes. Glad that this is noticed before we make it public :)

However, if we were to make it truly env-agnostic, the basic renderer should not be even able to access `process`, which is a Node-specific global (maybe it also exists in Nashorn, but not in every environment). The `process.env.BASIC_RENDERER` needs to be converted into constants during the compilation phase. I'll take a look at this as part of 2.5 features, but feel free to try addressing that.

wojtask9:
There isn't `process` in nashorn. I'm all for env-agnostic basic renderer.

I'll try switch from `process.env....` to constants

yyx990803:
FYI env vars to constant replacement can be configured [here](https://github.com/vuejs/vue/blob/dev/build/config.js#L172-L174)

wojtask9:
@yyx990803 
Something like this?
Another question. How to fix flow 
```
src/server/render-context.js:64
 64:     if (__BASIC_RENDERER__ === true) {
             ^^^^^^^^^^^^^^^^^^ identifier `__BASIC_RENDERER__`. Could not resolve name
```
looking for __WEEX__ I can't find anything about in config files

yyx990803:
Thanks for looking into this! Instead of making the whole rendering synchronous, we can just use a platform specific deferring function instead. In c5d0fa05 I've made it preferring `process.nextTick` -> `Promise.then` -> `setTimeout`.

In php-v8js Promise is already supported; in Nashorn, you can [polyfill setTimeout](https://blogs.oracle.com/nashorn/setinterval-and-settimeout-javascript-functions) using Java timers.

wojtask9:
Just for curiosity. Is there anything wrong with synchronous approach?
In my free time I'm working on Vue SSR in pure java (core still use Nashorn) and if there are some issues with synchronous approach it is better to me know this early :)

yyx990803:
Vue's SSR is designed to be async first, because there are many operations that may be async during render:

- async components
- external cache calls
- unwinding deep call stacks

Making things sync is just fundamentally against the design.

PR #6713, opened 2017-09-28
fix(v-html): added warning when v-html used with child content present

callumacrae:
v-html and v-text both overwrote the content of the element if anything
was present: this change checks if there's content and throws an warning
if there is.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

Improvement.

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/platforms/web/compiler/directives/html.js
+++ b/src/platforms/web/compiler/directives/html.js
@@ -2,7 +2,14 @@
 
 import { addProp } from 'compiler/helpers'
 
-export default function html (el: ASTElement, dir: ASTDirective) {
+export default function html (el: ASTElement, dir: ASTDirective, warn: Function) {
+  if (el.children.length) {
+    warn(
+      `<${el.tag} v-html="${dir.value}">:\n` +
+      'Element content should be empty with v-html, because it will be overwritten.'
+    )
+  }
+
   if (dir.value) {
     addProp(el, 'innerHTML', `_s(${dir.value})`)
   }
--- a/src/platforms/web/compiler/directives/text.js
+++ b/src/platforms/web/compiler/directives/text.js
@@ -2,7 +2,14 @@
 
 import { addProp } from 'compiler/helpers'
 
-export default function text (el: ASTElement, dir: ASTDirective) {
+export default function text (el: ASTElement, dir: ASTDirective, warn: Function) {
+  if (el.children.length) {
+    warn(
+      `<${el.tag} v-text="${dir.value}">:\n` +
+      'Element content should be empty with v-text, because it will be overwritten.'
+    )
+  }
+
   if (dir.value) {
     addProp(el, 'textContent', `_s(${dir.value})`)
   }
--- a/test/unit/features/directives/html.spec.js
+++ b/test/unit/features/directives/html.spec.js
@@ -64,4 +64,18 @@ describe('Directive v-html', () => {
       expect(vm.$el.innerHTML).toBe('')
     }).then(done)
   })
+
+  it('should warn when element has children', done => {
+    new Vue({
+      template: '<div v-html="a">test</div>',
+      data: { a: 'hello' }
+    }).$mount()
+
+    setTimeout(() => {
+      expect(`<div v-html="a">:
+Element content should be empty with v-html, because it will be overwritten.`)
+        .toHaveBeenWarned()
+      done()
+    }, 0)
+  })
 })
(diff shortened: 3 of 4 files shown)


callumacrae:
The failing test case is specifically testing that what I wrote isn't happening - should I just remove that case or something else?

it's this one: https://github.com/vuejs/vue/blob/dev/test/unit/modules/vdom/patch/hydration.spec.js#L301

src/platforms/web/compiler/directives/html.js
```diff
@@ -2,7 +2,14 @@
 
 import { addProp } from 'compiler/helpers'
 
-export default function html (el: ASTElement, dir: ASTDirective) {
+export default function html (el: ASTElement, dir: ASTDirective, warn: Function) {
+  if (el.children.length) {
```
blake-newman: Add `process.env.NODE_ENV !== 'production'` to ensure that this will only warn in non production enviroments.

src/platforms/web/compiler/directives/text.js
```diff
@@ -2,7 +2,14 @@
 
 import { addProp } from 'compiler/helpers'
 
-export default function text (el: ASTElement, dir: ASTDirective) {
+export default function text (el: ASTElement, dir: ASTDirective, warn: Function) {
+  if (el.children.length) {
```
blake-newman: Add `process.env.NODE_ENV !== 'production'` to ensure that this will only warn in non production enviroments

chrisvfritz:
@callumacrae Let's keep that failing test, as it seems there are a couple possible exceptions we didn't think of before:

- Server-side rendering, where we want some starting content that will later be replaced by client-side rendering
- In-DOM template rendering, where a user may provide default content until Vue takes over

callumacrae:
@chrisvfritz I guess in both of those cases the behaviour we're warning about is exactly the desired behaviour. Maybe this is a documentation issue instead of something that should throw a warning?

chrisvfritz:
It depends on how much of a pain it'll be to make exceptions for those two environments. @blake-newman might be able to provide more information here.

blake-newman:
@chrisvfritz I believe those are valid cases. For the SSR we can detect this environment with https://github.com/vuejs/vue/blob/55816543c46e75aa53481ac95a89ff6f87a2d704/src/core/util/env.js#L37

I'm not totally sure about the 'in browser template', also if we made conditions for different enviroments, this will be very confusing for the user. IMO it should remain consistent behaviour, and we should allow default values. 

Instead of placing this as a check the docs should probably notify the user of the replacement behaviour of this directive.

chrisvfritz:
@blake-newman But content with `v-html` wouldn't actually act as a default value in the contexts we want to warn about, right? In string templates, that content will simply be ignored during compilation.

callumacrae:
Perhaps adding a modifier to disable the warning would be the way to go - maybe `v-html.force` or `v-html.nowarn`. Users with those two specific cases can use the modifier, users who accidentally or unknowingly use both will be warning.

I guess this is a breaking change, though.

blake-newman:
I think having to declare modifiers to remove warnings, is not right. It adds a layer of additional complication, that we would also need to document. 

I do believe that this is down to the user to understand, and the warning will not add any value. Then having to add modifiers to disable is just more to learn.

chrisvfritz:
@blake-newman 

> I think having to declare modifiers to remove warnings, is not right. It adds a layer of additional complication, that we would also need to document.

I agree here.

> I do believe that this is down to the user to understand, and the warning will not add any value.

I disagree here. 🙂  It's not about learning, but about making a mistake that could be easy to make. These are the facts as I understand them (please correct me if you have a different understanding):

- Unless server-side rendering or in-DOM, there is _never_ a case where an element with `v-html` should have content. That can always be considered a mistake.
- To make this warning possible, we need to be able to confirm that we are not in-DOM or SSRing.
- We can easily detect SSR with `isServerRendering()`.
- We're not sure how to easily detect in-DOM compilation.

If all of the above are true, then there _is_ value to adding the warning and the only question is whether we can feasibly detect in-DOM compilation. Does that sound right?

yyx990803:
Technically we *can* detect in-dom template, but I don't think it's worth adding the extra check for this specific warning. The point is that this behavior does not result in potential malfunctioning of the app logic, so it's largely harmless, not to mention it even has legit use in some cases. As a rule of thumb, we should only warn about things that would lead to bugs if you neglect it.

PR #6734, opened 2017-10-03
feat(v-model): warn for undefined binding values (#5932)

### LINKED ISSUE #5932: Warn when `v-model` is bound on non-existent key
Considering this binding:

```<input type="text" v-model="data.val1">```

This works fine if there is for example { data:{ val1:null }} defined in component-data, but if only { data:{} } is defined it doesn't work correctly because while val1 is created by v-model, it's not made reactive.

By using Vue.set for setting values, this wouldn't be a problem which would make live easiert for certain usecases.

No visible change.

But to maintain the actual behaviour this could also be implemented using a modifier like 'set' for example:

```<input type="text" v-model.set="data.val1">```

This would use Vue.set(data, 'val1', ...) instead of simply generating data.val1 = ...

### PR DESCRIPTION
**Pull request by chriscasola**
Warn when the v-model binding value is undefined so users are aware that the binding may not be
reactive.

Fixes #5932

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

I'm not sure if checking if the value is undefined is the right thing to do. I was hoping to do some kind of `hasOwnProperty`-like check to see if the key in the object exists, but I couldn't figure out how to do that. If that's what makes more sense a pointer in the right direction would be awesome.

### DIFF
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -291,6 +291,7 @@ function genDirectives (el: ASTElement, state: CodegenState): string | void {
       needRuntime = !!gen(el, dir, state.warn)
     }
     if (needRuntime) {
+      let warning
       hasRuntime = true
       res += `{name:"${dir.name}",rawName:"${dir.rawName}"${
         dir.value ? `,value:(${dir.value}),expression:${JSON.stringify(dir.value)}` : ''
@@ -298,6 +299,8 @@ function genDirectives (el: ASTElement, state: CodegenState): string | void {
         dir.arg ? `,arg:"${dir.arg}"` : ''
       }${
         dir.modifiers ? `,modifiers:${JSON.stringify(dir.modifiers)}` : ''
+      }${
+        process.env.NODE_ENV !== 'production' && (warning = genValidateValue(dir.value)) ? `,warn:${warning}` : ''
       }},`
     }
   }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by yyx990803** [comment]
Thanks for the PR - unfortunately, this will also warn even if the value is explicitly declared as `undefined`:

``` js
data () {
  return {
    foo: undefined
  }
}
```

chriscasola:
Yeah...I noticed that but was unsure how to check properly:

> I'm not sure if checking if the value is undefined is the right thing to do. I was hoping to do some kind of hasOwnProperty-like check to see if the key in the object exists, but I couldn't figure out how to do that. If that's what makes more sense a pointer in the right direction would be awesome.

Any suggestions on how to check for the property existing?

posva:
Shouldn't the `in` operator be enough? `if (key in object)`

chriscasola:
The `in` operator isn't enough, because at this point in the code, I think the value of the binding has already been evaluated. So the `binding` only has two properties, `expression` and `value`. And if the key doesn't exist in the data of the component then `binding.value` is present but the value is undefined.

I think the check needs to be further up the stack but I'm not sure where. I'll keep digging.

chriscasola:
@posva @yyx990803 I think I have a solution that works correctly. I'm not sure it's the best way to go about this, but I did a lot of digging and I couldn't figure out any other way to be able to check properties in the second-to-last object in the binding expression.

yyx990803:
@chriscasola thanks a lot for looking into this - I just made a comment here: https://github.com/vuejs/vue/issues/5932#issuecomment-334040085

I also implemented a proper warning check strategy in https://github.com/vuejs/vue/commit/0fbad1e62bdd9f2ff2ca0a42d109face302ecd87 - but we will probably not use that. Feel free to compare that to your implementation - sorry for your wasted effort, hopefully this can at least help you understand the codebase better and make it easy for further contributions! 🙇

chriscasola:
@yyx990803 no problem - thanks for your commit, that definitely clarifies things.

PR #6735, opened 2017-10-03
Improve error msg for non-reactive properties

### LINKED ISSUE #6657: improve error message for non-reactive properties
current error message is non helpful for vue-class-component (js or ts):

if you forget to initialize a property then you get this error:

`Property or method "xxx" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option.`

So of course with class based components there is no 'data option', so this isn't helpful.

Proposal is to change the error message:

`Property or method "xxx" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property.  See:  https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties`

no changes.

kutyel:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

Improved error message.

- [ ] Yes
- [x] No

**Other information:**

Fixes #6657

### DIFF
--- a/src/core/instance/proxy.js
+++ b/src/core/instance/proxy.js
@@ -15,11 +15,11 @@ if (process.env.NODE_ENV !== 'production') {
 
   const warnNonPresent = (target, key) => {
     warn(
-      `Property or method "${key}" is not defined on the instance but
-      referenced during render. Make sure that this property is reactive,
-      either in the data option, or for class-based components, by
-      initializing the property.
-      See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.`,
+      `Property or method "${key}" is not defined on the instance but` +
+      'referenced during render. Make sure that this property is reactive, ' +
+      'either in the data option, or for class-based components, by ' +
+      'initializing the property.' +
+      'See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.',
       target
     )
   }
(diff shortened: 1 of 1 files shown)


src/core/instance/proxy.js
```diff
@@ -42,7 +44,7 @@ if (process.env.NODE_ENV !== 'production') {
   }
 
   const hasHandler = {
-    has (target, key) {
```
yyx990803: Can we remove the code style changes?

src/core/instance/proxy.js
```diff
@@ -15,9 +15,11 @@ if (process.env.NODE_ENV !== 'production') {
...
     warn(
-      `Property or method "${key}" is not defined on the instance but ` +
-      `referenced during render. Make sure to declare reactive data ` +
-      `properties in the data option.`,
+      `Property or method "${key}" is not defined on the instance but
+      referenced during render. Make sure that this property is reactive,
+      either in the data option, or for class-based components, by
+      initializing the property.
+      See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.`,
       target
```
yyx990803: We need to use concatenations for each line to avoid the whitespace between the lines.

kutyel:
Thanks, @yyx990803 should be ok right now 👌

posva:
@kutyel you can run `npm run lint -- --fix` to fix the lint problems

PR #6736, opened 2017-10-04
Add slot v-bind warning (fix #6677)

### LINKED ISSUE #6677: v-bind with something else than an object should warn you
### Version
2.4.4

### Reproduction link
[https://jsfiddle.net/57508hba/4/](https://jsfiddle.net/57508hba/4/)

### Steps to reproduce
1. Create a component with a slot
2. Bind a reference to a string to the slot via `v-bind`
3. Utilize said component and define a template for the defined slot, giving it a local scope parameter with which to output the bound string.
4. Output the scoped string in your template which is passed to the component created in step 1

### What is expected?
The string bound to the slot should be passed into the template passed for the slot, as a string.

The string is being exploded into an object where the properties are the string indexes for each character in the string.

---
I'm not sure if this is intended behavior or not. The default value for the defined slot outputs the string as expected, but as soon as it's passed to the consumer defined template, the string is altered to an object.

The temporary work-around for this is to simply pass an actual object to v-bind on the slot, and expect an object as the scoped value in your defined template.

It seems that if this is intended behavior, there should be at least a Vue warning about the value being mangled and that v-bind expects an object and not string.

jamOne-:
v-bind with not object warning

close #6677 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


src/core/instance/render-helpers/render-slot.js
```diff
@@ -15,6 +15,12 @@ export function renderSlot (
   if (scopedSlotFn) { // scoped slot
     props = props || {}
     if (bindObject) {
+      if (!isObject(bindObject)) {
+        process.env.NODE_ENV !== 'production' && warn(
```
posva: can you move the `process.env.NODE_ENV !== 'production'` into the if condition so it gets stripped off in production mode?

jamOne-: Oh, of course I can move it!
Sure, I will add the test.

jamOne-:
Ready to merge :)

PR #6740, opened 2017-10-05
fix(v-model): update model correctly when input event is bound #6690

### LINKED ISSUE #6690: Select value is not updated correctly when input handler triggers class change
### Version
2.4.4
Current chrome, safari

### Reproduction link
[https://jsfiddle.net/6dupwfpf/6/](https://jsfiddle.net/6dupwfpf/6/)

### Steps to reproduce
- open provided JSFiddle
- change select box value to "Alpha"

### What is expected?
Value should be changed and "green" class applied.

"green" class is applied, but value stays unchanged.

---
Provided fiddle runs on version 2.4.0, but I tested it locally on 2.4.4 and it behaves exactly the same.

This report is linked to https://github.com/monterail/vuelidate/issues/192

chriscasola:
Use the "input" event instead of "change" when listening for
changes to the selection state since the "input" event will
always fire before the "change" event.

This avoids an issue where a user binds to the "input" event
with "v-bind" and causes the component to update and set its
model value back to the value of the select before it has
received the new selection value.

Fixes #6690

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

There is a detailed description of what causes this bug in https://github.com/vuejs/vue/issues/6690#issuecomment-333734327

### DIFF
--- a/src/platforms/web/compiler/directives/model.js
+++ b/src/platforms/web/compiler/directives/model.js
@@ -126,7 +126,7 @@ function genSelect (
   const assignment = '$event.target.multiple ? $$selectedVal : $$selectedVal[0]'
   let code = `var $$selectedVal = ${selectedVal};`
   code = `${code} ${genAssignmentCode(value, assignment)}`
-  addHandler(el, 'change', code, null, true)
+  addHandler(el, 'input', code, null, true)
 }
 
 function genDefaultModel (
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks again for the PR - however, I don't think switching events would solve the root problem, which is that microtasks (which Vue uses for async deferring) can fire in between these supposedly sequential events.

I managed to find a nextTick implementation that fires after all sequential events in 6e41679a96582da3e0a60bdbf123c33ba0e86b31.

Issue #6659, opened 2017-09-21
v-if影响预期定义的组件


dohkoxiaozu:
### Version
2.3.3

### Reproduction link
[https://jsfiddle.net/vitorDeng/50wL7mdz/63147/](https://jsfiddle.net/vitorDeng/50wL7mdz/63147/)

### Steps to reproduce
打开浏览器控制台，点击Run

### What is expected?
其他v-if的元素不应该影响互不相干的组件

### What is actually happening?
同级下，若前后使用了与包裹组件的元素一样的，并使用了v-if指令，则v-if指令会影响该组件不断创建和销毁

Jinjiang:
主要是你那个发生改变的组件和 `v-if` 里的东西完全一样，diff 的时候难免会判断错误，但最起码也可以保持渲染出来的结果是一样的。临时解决办法就是在这些标签上加一些不一样的 attribute，就立刻工作正常了。

----

Translate into English:

Temporarily you can easily change:

```html
<div><custom-component ref="componentPlugin"></custom-component></div>
```

into:

``` html
<div marked><custom-component ref="componentPlugin"></custom-component></div>
```

to avoid the re-render of the custom component in the middle.

Because the `v-if` component and the element above is all the same tag + without any props/attrs. So It's hard to the detect update details very exactly. But you can add special attribute to solve this.

Thanks.

yyx990803:
For now you can use a `key` on the div to mark it: https://jsfiddle.net/50wL7mdz/63293/

However this can be improved for sure.

dohkoxiaozu:
谢谢

looser098:
看不懂哟

14780062668:
key is importatnt

chriscasola:
I'll take a stab at this.

PR #6750, opened 2017-10-07
perf(vdom): don't move plain elements containing components (#6659)

chriscasola:
Do not move a plain element during patch if it contains any
components. This will prevent re-creating the component in a
new element after it's previous parent is moved.

#6659

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This is my naive attempt at improving the logic in the diff process to prevent moving a plain element that contains components. I'm definitely open to better options and feedback, I'm not sure this is the right direction to go in. Thanks!

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -55,6 +55,16 @@ function sameInputType (a, b) {
   return typeA === typeB || isTextInputType(typeA) && isTextInputType(typeB)
 }
 
+function plainNodeWithComponent (a) {
+  let i
+  const { children } = a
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks for the PR and sorry for the late review.

While this is addressing a real problem, this solution may not be worth the tradeoff. The current implementation does not take into account potential deeply nested components. However accounting for all nested components will be prohibitively expensive. So, this is more or less a band-aid solution and I'd rather resolve it in a major refactor of the core algorithm in the future.

PR #6781, opened 2017-10-12
Changed discord widget

Samuell1:
Changed to official discord widget with number of online people

<a href="https://chat.vuejs.org/">Chat


posva:
While I like the widget because it also displays the online people, current widget is super clear about what discord is because it says _chat on discord_ but this one isn't, meaning people not knowing discord will probably ignore it

yyx990803:
I agree with @posva - not everyone know what Discord is and we should definitely keep the "chat" part. Thanks for the PR though.

Issue #6850, opened 2017-10-19
Export Prop type in main types file (vue/types/index.d.ts)


ferdaber:
### What problem does this feature solve?
When attempting to declare a type for a complex object-type prop for a component, it's difficult to understand (without knowing exactly how the Props typing works) how to cast that specific prop to the interface you create.

The best way of doing it currently is to do this:
```ts
import Vue from 'vue'
import { Prop } from 'vue/types/options'

interface FooBar {
  foo: string,
  bar: string
}

export default Vue.extend({
  props: {
    foobar: Object as (() => FooBar)
  }
})
```

### What does the proposed API look like?
It would be nice to expose the Prop generic type in vue/types/options.d.ts in the default types export so that we can simply use this:
```ts
import Vue, { Prop } from 'vue'

// boilerplate
  props: {
    foobar: Object as Props<FooBar>
}
```

This also extends to complex callback types if need be so that whenever the regular prop types via the primitive type constructors don't work, we can create an interface and cast it as `Prop<ICustomType>` and it just works.

sandangel:
Did you tried vue-property-decorator? 
```ts
import Vue from 'vue'
import { Prop, Component } from 'vue-property-decorator'

interface FooBar {
  foo: string,
  bar: string
}

@Component
export default class App extends Vue {
  @Prop() foobar: FooBar;
}
```

ktsn:
I like the idea. `Prop` type would hide the internal detail of type declaration and looks more intuitive syntax.

@sandangel This proposal is not related with decorators. It would help us declare `props` type easier with canonical API.

ferdaber:
Yep, it's not obvious from that the type declaration uses generic constructors or factory functions to tease out the generic type, and exposing the `Prop` type makes it much easier to use custom interfaces

ferdaber:
I can do a PR if it's easier since this would be a very simple change.

PR #6856, opened 2017-10-19
feat: add Prop to main type declaration file [#6850]

ferdaber:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

See #6850.

**Other information:**


HerringtonDarkholme:
This should also fix https://github.com/vuejs/vue/issues/6841.

I wonder how it would impact case like `union: [User, Number]`. 

Of course, exporting more types means more compatibility for us to maintain. If it can also fix `union` case I think we should ship this feature.

ferdaber:
@HerringtonDarkholme it seems to work fine with union types where you want the constructor instead of the primitive:
union

ktsn:
To deal with union type, I guess it would be more convenient to declare another type (e.g. `PropType` in the following example) to unify the casting syntax.

```ts
type PropType<T> = Prop<T> | Prop<T>[]

interface Foo {
  a: string
}

Vue.extend({
  props: {
    // Both looks the same syntax
    // (we don't need to concern the `type` is an array or not)
    foo: Object as PropType<Foo>
    union: [Object, String] as PropType<Foo | string>
  }
})
```

HerringtonDarkholme:
`Prop<string | Foo>[]` is enough for union case. This is both concise and saves us one exported type.

ferdaber:
@HerringtonDarkholme, I think that @ktsn has a good point. That is technically the purpose of `PropValidator<T>` which serves as the all-encompassing union of `PropOptions<T> | Prop<T> | Prop<T>[]`. Would it be better to export `PropValidator<T>` instead so that users don't have to worry about conforming to the Vue prop validator syntax, and just do:
```js
props: {
  myProp: {/* constructor, array, or prop options */} as PropValidator<IMyProp>
```

ktsn:
@HerringtonDarkholme 
I didn't mean we should export both `Prop` and `PropType` - I think we should minimize exposing types as possible too.

But the `Prop` might be a little low level to be used in user land because we need to consider whether the props `type` is an array or not, even though we just want to focus its type parameter in that case.
So I would suggest to add a thin wrapper type to intended to be used in the user land.

@ferdaber 
I'd say `PropValidator` shouldn't be exposed since we just want to cast `type` here. It seems a little overkill for me.

ferdaber:
The only difference between `PropType<T>` (which currently does not exist in options.d.ts), and `PropValidator<T>` is the additional union of the customized object syntax of the prop validator API. If someone were to use the customized object syntax then they would still need to do something like:
```ts
props: {
  union: [Object, string] as PropType<Foo | string>,
  complexUnion: {
    type: [Object, String] as PropType<Foo | string>
    // ... other stuff
  }
}
```
versus
```ts
props: {
  union: [Object, String] as PropValidator<Foo | string>,
  complexUnion: {
    type: [Object, String]
    // ... other stuff
  } as PropValidator<Foo | string>
}
```

I guess to me it makes more sense to have the typecast be in the same level of each prop definition, but it doesn't really make much of a difference so I'm fine with whichever you all decide. The other benefit I see too would be that `PropValidator` is the more abstract definition of props in a Vue constructor, so if in the future the lower-level API changes `PropValidator` is the one that's least likely to change

yyx990803:
I think `PropType` is more pragmatic than `Prop` and more intuitive than `PropValidator`.

ferdaber:
Great! I will modify the PR to create a new type called `PropType` and export that in place of `Prop`

types/options.d.ts
```diff
@@ -122,9 +122,11 @@ export interface RenderContext<Props=DefaultProps> {
...
-export type Prop<T> = { (): T } | { new (...args: any[]): T & object }
+export type Prop<T> = { (): T } | { new(...args: any[]): T & object }
 
-export type PropValidator<T> = PropOptions<T> | Prop<T> | Prop<T>[];
+export type PropType<T> = Prop<T> | Prop<T>[];
+
+export type PropValidator<T> = PropOptions<T> | PropType<T>;
 
 export interface PropOptions<T=any> {
   type?: Prop<T> | Prop<T>[];
```
thatguystone: Should the type here be `PropType<T>`?

alexsasharegan:
Does this PR solve the prop compile error: `TS2339: Property 'x' does not exist on type 'CombinedVueInstance<Vue, ...>`?

**Compiler Error**

```sh
ERROR in src/js/components/Manage/DataViz/HdSalesGraph.vue.ts
[tsl] ERROR in src/js/components/Manage/DataViz/HdSalesGraph.vue.ts(71,26)
      TS2339: Property 'currentMoment' does not exist on type 'CombinedVueInstance<Vue, { styles: { width: string; height: string; position: string; }; }, { for...'.

ERROR in src/js/components/Manage/DataViz/HdSalesGraph.vue.ts
[tsl] ERROR in src/js/components/Manage/DataViz/HdSalesGraph.vue.ts(75,26)
      TS2339: Property 'previousMoment' does not exist on type 'CombinedVueInstance<Vue, { styles: { width: string; height: string; position: string; }; }, { for...'.
```

**Abbreviated Source**

```html
<script lang="ts">
import Vue from "vue";
import { Prop } from "vue/types/options";
import moment from "moment";

export default Vue.extend({
  name: "HdSalesGraph",

  props: {
    currentMoment: {
      type: Object as Prop<moment.Moment>,
      required: true,
    },

    previousMoment: {
      type: Object as Prop<moment.Moment>,
      required: true,
    },
  },

  data() {
    return {
      styles: {
        width: "100%",
        height: "500px",
        position: "relative",
      },
    };
  },

  computed: {
    curMoment(): moment.Moment {
      return moment(this.currentMoment);
      //                 ^^^^^^^^^^^^^ 71,26 
    },

    prevMoment(): moment.Moment {
      return moment(this.previousMoment);
      //                 ^^^^^^^^^^^^^^ 75,26
    },
  },
});
</script>
```

KindWizzard:
Any updates on this?

ascii-soup:
Any movement on this? Keen to see it land!

Grandizer:
Do we know when this will be released?  I am sure "when it's ready" but just looking for a guestimate?

pushkarskiy:
Do we know when this will be released? This very important for us.

sjmcdowall:
Gang -- when will this happen?!  This is causing us a major issue in our large Vue project.. it looks like everything is done and ready to go? Why isn't it merged?

ffxsam:
Why is this hanging around in limbo?

ffxsam:
@yyx990803 @ktsn @sodatea Anyone? This bug is causing unnecessary TypeScript errors in our projects.

yyx990803:
Once we finish CLI release this will be the first to be merged in core, please bear with us.

RehanSaeed:
Just want to echo the question by @alexsasharegan. Will it solve that problem?

r0skar:
@RehanSaeed Yes there is. Check out [this](https://frontendsociety.com/using-a-typescript-interfaces-and-types-as-a-prop-type-in-vuejs-508ab3f83480) great article by Mitchell Garcia.

ffxsam:
@yyx990803 It's been a month since your last comment. Any hope of getting this merged soon? I'm glad there's a workaround (by simply excluding `type: Array`), but it's still a nuisance to not have proper type checking for props.

ky-is:
@ffxsam Looking forward to this as well, but is the function interface syntax not a viable workaround? Types are working for me with:
```js
type: Array as () => string[],
```

ffxsam:
@ky-is Nice trick, I'll give it a try. Thanks!

ffxsam:
The solution is this PR. I'm not sure what the holdup is.

kjleitz:
@yyx990803 @ferdaber can this be merged in? It's a big roadblock for our TypeScript codebase, too!

cesalberca:
I think what's happening is that they won't focus on this kind of issues as they are rewritting Vue in TypeScript (as announced here: https://medium.com/the-vue-point/plans-for-the-next-iteration-of-vue-js-777ffea6fabf).

sjmcdowall:
@cesalberca -- that is all well and good, but realistically it may be a year before adoption of V3.0 etc. meanwhile people everyday are going through this pain -- which is FIXED -- and an easy patch.  It doesn't make sense.

wanton7:
Maybe just just make a local copy and patch it yourself and not wait forever? That what we did.

cesalberca:
> @cesalberca -- that is all well and good, but realistically it may be a year before adoption of V3.0 etc. meanwhile people everyday are going through this pain -- which is FIXED -- and an easy patch. It doesn't make sense.

I know @sjmcdowall , I'm just saying that this is what may be occurring. I wouldn't agree too if this is whats truly happening.

japboy:
will this be merged so soon?

abelgoodwin1988:
Fantastic!
Is there some additional documentation on this somewhere so I can double make sure I'm implementing this correctly? I have:
```
<script lang="ts">
import Vue from 'vue';
import { Prop } from 'vue/types/options';
import { Education } from '@/assets/resume/resume.model';

export default Vue.extend({
    props: {
        education: {
            type: Object as Prop<Education>,
            required: false,
        },
    },
});
```
Is this the appropriate way to do it?

LissetteIbnz:
@abelgoodwin1988, I think you should use it like that
```
<script lang="ts">
import Vue, { PropType } from 'vue';
import { Education } from '@/assets/resume/resume.model';

export default Vue.extend({
    props: {
        education: Array as PropType<Education[]>,
    },
});
```

The PropType interface definition is `export type PropType<T> = Prop<T> | Prop<T>[];`

> Fantastic!
> Thanks for all the hard work and merging this.
> Is there some additional documentation on this somewhere so I can double make sure I'm implementing this correctly? I have:
> 
> ```
> <script lang="ts">
> import Vue from 'vue';
> import { Prop } from 'vue/types/options';
> import { Education } from '@/assets/resume/resume.model';
> 
> export default Vue.extend({
>     props: {
>         education: Array as Prop<Education[]>,
>     },
> });
> ```
> 
> Is this the appropriate way to do it?

alexsasharegan:
I've been having good luck with PropOptions as an alternative approach.

```html
<script lang="ts">
import Vue, { PropOptions } from 'vue';
import { Education } from '@/assets/resume/resume.model';

export default Vue.extend({
    props: {
        education: Array,
    } as PropOptions<Education[]>,
});
```

The name might be singular, I can check now from my phone. I like this approach because it also hints the validator function parameters.

liulei237136:
ok

PR #6885, opened 2017-10-23
fix #6864: don't set ignoredElements' attributes to their default values

### LINKED ISSUE #6864: Overwrite of element attributes with default value
For custom HTML elements Vue overwrites some attributes with default values. One example is the `selected` attribute which will always be set to `selected=selected` for truthy values. This absolutely makes sense for HTML elements that are defined in the HTML standard, but not for custom elements. Custom elements are not yet a web standard but there are already ways to build and use them. We experienced the issue when experimenting with a custom element for a tab bar which has the selected tab stored in a selected attribute.

Find a code example in:
https://jsfiddle.net/1pc2e9yj/2/
(The code example is currently limited to Chrome and Safari, because of missing polyfills for other browsers)

Vue already has a config setting for `ignoredElements`. My proposal would be, that attributes are not set to their default values for elements in the `ignoredElements` array. If that approach is valid, then the example code in the jsfiddle would stay the same, but the output would change to "selected attribute value: tab1"

imyzf:
Fix #6864

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


yyx990803:
Thanks for the PR. A few issues:

1. In 2.5 `ignoredElements` also support regular expressions.
2. Placing this check in `setAttr` can be too expensive. Ideally we don't want to affect performance of the entire framework for such a niche use case.

imyzf:
@yyx990803 Thanks for your suggestions. I have a new idea:

imyzf:
@yyx990803 Thanks for your suggestions. I have a new idea:

1. Add a flag `isIgnoredElement` to `VNode`
2. Judge and set `isIgnoredElement` in [createElm](https://github.com/vuejs/vue/blob/v2.5.2/src/core/vdom/patch.js#L107)
3. In `setAttr`,  strip the default attribution process if `isIgnoredElement` is `true`

Would this be better?

fenggenggeng:
.

fenggenggeng:
这是对话吗？

yyx990803:
Thanks a lot for the PR. After re-consideration, I think the easier way to deal with this is simply checking if the element could be a custom one when setting the attribute. This avoids the overhead of adding flags to the vnode.

PR #6892, opened 2017-10-23
[weex] Support to compile `v-on` into the weex native directive

Hanks10100:
Finish #6622. Generate the binding params of the event handler.

Only effect the Weex platform within the `<recycle-list>` component.


Hanks10100:
The `ci:test-ssr-weex` is failed because the version of `weex-js-runtime` is mismatch. It has already been fixed in the dev branch. (https://github.com/vuejs/vue/commit/f38d44e23bc9f3eda950a9436205f42be006abfc)

Hanks10100:
The `test-ssr-weex` is failed because the version of `weex-js-runtime` is mismatch with the `weex-vue-framework`.

It has already been fixed in the dev branch. (https://github.com/vuejs/vue/commit/f38d44e23bc9f3eda950a9436205f42be006abfc)

PR #6932, opened 2017-10-26
fix #6931: Clean up target variables to avoid memory leaks

### LINKED ISSUE #6931: Memory leak in updateDOMListeners and updateComponentListeners
### Version
2.5.2

### Reproduction link
[https://jsfiddle.net/opzzsudp/](https://jsfiddle.net/opzzsudp/)

### Steps to reproduce
1. Create a Vue instance that has a reference to a large object in its data
1. Render either a component or an element with any event handler
1. Destroy the Vue instance, remove its element, null out its reference

### What is expected?
The Vue instance and the references it contains should be freed for garbage collection

The `target` variables [here](https://github.com/vuejs/vue/blob/dev/src/platforms/web/runtime/modules/events.js#L29) or [here](https://github.com/vuejs/vue/blob/dev/src/core/instance/events.js#L22) hang onto references and prevent garbage collection

---
I have a fix ready for this and will be filing a PR shortly

fenduru:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
The development environment on Windows is flaky and crashes a lot. I'm also seeing intermittent failures that seem unrelated to my change (they also occur on the `dev` branch with no changes). Filing the PR now to get feedback and to see if CI has the same failures.

### DIFF
--- a/src/core/instance/events.js
+++ b/src/core/instance/events.js
@@ -19,7 +19,7 @@ export function initEvents (vm: Component) {
   }
 }
 
-let target: Component
+let target: any
 
 function add (event, fn, once) {
   if (once) {
@@ -40,6 +40,7 @@ export function updateComponentListeners (
 ) {
   target = vm
   updateListeners(listeners, oldListeners || {}, add, remove, vm)
+  target = undefined
 }
 
 export function eventsMixin (Vue: Class<Component>) {
--- a/src/platforms/web/runtime/modules/events.js
+++ b/src/platforms/web/runtime/modules/events.js
@@ -26,7 +26,7 @@ function normalizeEvents (on) {
   }
 }
 
-let target: HTMLElement
+let target: any
 
 function createOnceHandler (handler, event, capture) {
   const _target = target // save current target element in closure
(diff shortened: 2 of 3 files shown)


src/core/instance/events.js
```diff
@@ -19,7 +19,7 @@ export function initEvents (vm: Component) {
   }
 }
 
-let target: Component
+let target: any
```
fenduru: I don't love removing the type annotation here, but if I left it as `Component` then Flow would see the assignment to undefined and error due to "Possible undefined value".

In typescript you can tell the type checker to trust you by doing `target!.method()` but I couldn't find an equivilent in Flow. The Flow docs recommend `if (target !== undefined)`, however I didn't think we'd want to add a branch in the VDom code.

chriscasola: What about just binding the `add` and `remove` methods to the `vm` so you don't need this closure variable at all?

dsonet: @chriscasola You won't wanna do it that way. function.bind is just create a closure.

PR #6934, opened 2017-10-26
Properly set inherited setters

eschalks:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Encountered an issue where v-model would not try to call a setter on an instance of a class.
A basic example:

```
    <div>
        <input v-model="person.fullName">
        <table>
            <tr>
                <td>First Name</td>
                <td>{{ person.firstName }}</td>
            </tr>
            <tr>
                <td>Last Name</td>
                <td>{{ person.lastName }}</td>
            </tr>
            <tr>
                <td>Full Name</td>
                <td>{{ person.fullName }}</td>
            </tr>
        </table>
    </div>
```
```
export class Person {
    constructor() {
        this.firstName = 'First'
        this.lastName = 'Last'
    }

    get fullName() {
        if (!this.lastName) {
            return this.firstName;
        }
        return this.firstName + ' ' + this.lastName
    }

    set fullName(v) {
        const split = v.split(' ', 2)
        this.firstName = split[0]
        this.lastName = split.length === 2 ? split[1] : ''
    }
}
```

In the above example the firstName and lastName don't change when the value of the input changes because the fullName setter is never called.

This PR fixes this by checking if we're trying to set a property that is defined as a setter in an object's prototype chain.
Because this explicitly checks for the presence of a .set on the property descriptor it will not try to set normal properties, to keep it more in line with how set has worked so far.

I'm guessing simply doing a (key in obj) check followed by a obj[key] = value would be more performant, but this being my first time looking at the Vue source code I'm not sure what this might break.

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -199,14 +199,6 @@ export function set (target: Array<any> | Object, key: any, val: any): any {
     target[key] = val
     return val
   }
-  if (!Array.isArray(target)) {
-    const propertyDescriptor = getInheritedPropertyDescriptor(target, key)
-    if (propertyDescriptor && propertyDescriptor.set) {
-      propertyDescriptor.set.call(target, val)
-      return val
-    }
-  }
-
   const ob = (target: any).__ob__
   if (target._isVue || (ob && ob.vmCount)) {
     process.env.NODE_ENV !== 'production' && warn(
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks for the PR - this is actually `v-model` related and is already addressed in 83ed92608d81349e1cac2e481ed079e51a490b2b

PR #6940, opened 2017-10-27
feat(v-model): add lazy modifier for v-model on components(close #6914)

### LINKED ISSUE #6914: Request: add v-model.lazy for components
v-model for components already supports .trim and .number.
Lazy modifier support for components would make it more uniform across the platform.

The only change (backwards compatible) would be to add a new property to model definition:
```
model: {
  prop: 'value',
  event: 'input',
  eventLazy: 'change'
}
```

javoski:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
close #6914 .

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -131,6 +131,7 @@ declare type ASTElement = {
     value: string;
     callback: string;
     expression: string;
+    event: string;
   };
 
   directives?: Array<ASTDirective>;
--- a/flow/vnode.js
+++ b/flow/vnode.js
@@ -62,6 +62,7 @@ declare interface VNodeData {
   model?: {
     value: any;
     callback: Function;
+    event: string;
   };
 };
 
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -255,6 +255,8 @@ export function genData (el: ASTElement, state: CodegenState): string {
       el.model.callback
     },expression:${
       el.model.expression
+    },event:${
+      el.model.event
     }},`
   }
   // inline-template
--- a/src/compiler/directives/model.js
+++ b/src/compiler/directives/model.js
@@ -8,7 +8,7 @@ export function genComponentModel (
   value: string,
   modifiers: ?ASTModifiers
 ): ?boolean {
-  const { number, trim } = modifiers || {}
+  const { number, trim, lazy } = modifiers || {}
 
   const baseValueExpression = '$$v'
   let valueExpression = baseValueExpression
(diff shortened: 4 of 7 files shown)


rstoenescu:
Any news on this? Is it going to get merged, and if so, when is it scheduled to get into an official release? Thanks.

posva:
@javoski  I think adding this feature should allow **both** using `v-model` and `v-model.lazy`, and therefore, a component could potentially have 2 different events for model

---

On a side note, this is possible with a prop on the component

```vue
<my-comp v-model="value" lazy/>
```

```vue
// my-comp.vue
<input :value="value" @inpu="$emit('input', $event.target.value)" v-if="lazy" />
<input :value="value" @change="$emit('input', $event.target.value)" v-else />
```

or with a render function as well http://jsfiddle.net/dazvtwat/

I do agree to have `.lazy` makes it more succinct though 🙂

javoski:
@posva Yeah, it's reasonable for components to aware of if they are using `.lazy` modifier on `v-model`, and prop is a proper way. 
But in the case of single input element wrapper component, I prefer to do like this:
```js
const MyComp = {
  ...
  template: `<input :value="value" @input="onUpdate" @change="onUpdate">`,
  methods: {
    onUpdate (e) {
      this.$emit(e.type, e.target.value)
    }
  }
}
```

posva:
That's much better, indeed, completely forgot about `event.type` 😄 
But we should allow the user to customize both, `input` and `change` events name emitted by the component because the component can be used in both scenarios but only one event name can be provided, and if it is provided, it will end up used by both `v-model.lazy` and `v-model`

tmorehouse:
It would be nice for a component to actually know if the `v-model` is requesting a `lazy` (in some cases this may be required).

Native inputs will have `element._vModifiers` set with an object defining what modifiers are in use (i.e. `{ lazy: true, trim: true }`, while the root element of a component will have `this.$el.vModifiers` set to an empty object, regardless of modifiers in use.  It would be very handy to have these values made available to components (or maybe exposed as a readonly `this.$vModifiers` injection, although the former would be easier to implement)

Tantawi:
@javoski Can you please look at the merge requests and try to get this merged? thanks a lot!

Sshuichi:
@PhilipJohnBasile Please mege? :3

Tantawi:
@javoski Awesome now that it is approved, can you please look at the conflicts and merge? thanks a lot!

PhilipJohnBasile:
There's merge conflics @Tantawi

cn-xufei:
Exciting. When can we merge?

javoski:
This PR is too old, so I opened a new one based on the latest codebase. #11830.

Issue #6938, opened 2017-10-27
keep-alive include will also cache anonymous components


PanJiaChen:
### Version
2.5.2

### Reproduction link
[https://jsfiddle.net/L613xva0/13/](https://jsfiddle.net/L613xva0/13/)

### Steps to reproduce

triggle the route 

### What is expected?
the second route component not be cached

### What is actually happening?
the second route component  be cached

---
this [demo](https://jsfiddle.net/L613xva0/12/) tow router-view component both have name, the result is right , 
but in this [demo](https://jsfiddle.net/L613xva0/13/), the second route have no name ,but also be cached

>The match is first checked on the component’s own name option, then its local registration name (the key in the parent’s components option) if the name option is not available. Anonymous components cannot be matched against.

  ```
<keep-alive :include="cachePages">
    <router-view></router-view>
 </keep-alive>
```
when l keep-alive router-view , every anonymous component page will be cached , i don't think is reasonable.

yuu2lee4:
@yyx990803   Now keep-alive will not cache any anonymous components without include prop.

devinRex:
###  The same as @yuu2lee4.
### You can see the reproduction in the console panel of this link [https://jsfiddle.net/tqfc2edq/](https://jsfiddle.net/tqfc2edq/)
### The vue version 2.5.6, and you can see without "include", all anonymous components can't cahce. only mounted , without activated.

yyx990803:
Fixed in a23b913796a7d18e76185607f250655e18a390c8

Alendorff:
@yyx990803 but what if I need to cache anonymous component, how can I do that? There is no way? Why not to stick with the key passed to it? For example for that case I would expect this component working under keep-alive:
![image](https://user-images.githubusercontent.com/4989157/60678551-229c0500-9e85-11e9-83f3-efd0b19adbc9.png)

But it doesn't. Because it has no name.

PR #6985, opened 2017-11-01
fix:#6938 <keep-alive> should not cache anonymous components

cuteyumiko:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
anonymous components aslo have tag , so func getComponentName can not be used anymore.

### DIFF
--- a/src/core/components/keep-alive.js
+++ b/src/core/components/keep-alive.js
@@ -85,7 +85,7 @@ export default {
     const componentOptions: ?VNodeComponentOptions = vnode && vnode.componentOptions
     if (componentOptions) {
       // check pattern
-      const name: ?string = componentOptions && componentOptions.Ctor.options.name
+      const name: ?string = componentOptions.Ctor.options.name
       if (!name || (
         (this.exclude && matches(this.exclude, name)) ||
         (this.include && !matches(this.include, name))
--- a/src/core/components/keep-alive.js
+++ b/src/core/components/keep-alive.js
@@ -85,8 +85,8 @@ export default {
     const componentOptions: ?VNodeComponentOptions = vnode && vnode.componentOptions
     if (componentOptions) {
       // check pattern
-      const name: ?string = getComponentName(componentOptions)
-      if (name && (
+      const name: ?string = componentOptions && componentOptions.Ctor.options.name
+      if (!name || (
         (this.exclude && matches(this.exclude, name)) ||
         (this.include && !matches(this.include, name))
       )) {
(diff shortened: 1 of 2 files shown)


cuteyumiko:
@jkzing  unit test added, please review

src/core/components/keep-alive.js
```diff
@@ -85,8 +85,8 @@ export default {
     const componentOptions: ?VNodeComponentOptions = vnode && vnode.componentOptions
     if (componentOptions) {
       // check pattern
-      const name: ?string = getComponentName(componentOptions)
-      if (name && (
+      const name: ?string = componentOptions && componentOptions.Ctor.options.name
```
dsonet: why do you change the way of getting name? Plus doesn't need `componentOptions &&`  since it's already in the if branch.
IMO it should keep original logic ` opts.Ctor.options.name || opts.tag`

cuteyumiko: Thanks for review , already modified.

docs descripition:
The match is first checked on the component’s own name option, then its local registration name (the key in the parent’s components option) if the name option is not available. Anonymous components cannot be matched against.

Anonymous components always have a registered name(  tag in code) , i think the descripition of docs may cause logical hole.

jkzing: @cuteyumiko IMO we may need to keep `opts.tag`. For example:
``` js
const foo = {
  template: `<div>foo</div>`
}

new Vue({
  el: '#app',
  components: {
    foo
  }
})
``` 

In above case, you'll find `Ctor.options` doesn't has a `name` property, but there actually is `tag` in it.

cuteyumiko: ehhh..i see 
i have done another test for caching anonymous components , it turns out caching logic is ok

 #6938  it's problem may be casued by vue-router , i tracked this issue in a wrong way

PanJiaChen:
@cuteyumiko why close the pr ?

cuteyumiko:
@PanJiaChen  it works fine in offical unit test .your issue may be casued by vue-router .

PanJiaChen:
may be is `vue-router` bug or `vue`. but the problem is objective existence。
if a router has not set name :
```js
function getComponentName (opts) {
  return opts && (opts.Ctor.options.name || opts.tag)
}
```
this will return undefined, so has some problems when use `keep-alive`

detail:https://github.com/vuejs/vue/issues/6938
@yyx990803

PR #6987, opened 2017-11-02
feat($compiler): support to generate "@render" function for weex recycle-list

Hanks10100:

- [x] Feature

- [x] No

**Other information:**

Compile the template twice with different options for weex platform if the `recyclable` flag is passed. Generate both normal render function and `@render` function for `<recycle-list>`.

Function names and arguments are also adjusted in `<recycle-list>` compiler.

If a component has the `recyclable` attribute on the template (`<template recyclable="true">`), weex-vue-loader will pass the `recyclable` option to weex-template-compiler. (https://github.com/weexteam/weex-vue-loader/commit/2b3ed319fa40bdde20f5a284399c1268a4af689d)

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -17,6 +17,9 @@ declare type CompilerOptions = {
   shouldDecodeNewlines?:  boolean;
   shouldDecodeNewlinesForHref?: boolean;
 
+  // support <recycle-list> in weex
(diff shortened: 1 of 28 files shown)


Hanks10100:
I also add many test cases for `<recycle-list>`, which is located in  `test/weex/cases/recycle-list/`.

PR #7023, opened 2017-11-08
fix #6687 out-in transition getting stuck with v-if

### LINKED ISSUE #6687: Transitioning between components fails when component's root node has v-if="false"
### Version
2.4.4

### Reproduction link
[https://jsfiddle.net/by4bfduo/](https://jsfiddle.net/by4bfduo/)

### Steps to reproduce
Click radio "B".

### What is expected?
Should switch to component "B".

Doesn't switch to "B".

---
In real scenario, when a user navigates to a new view, I want to load the remote data (in created hook) and set the root node's v-if to true to switch the component in. But when data hasn't been load for some reason, the dynamic component is stuck and can't go anywhere.

neelance:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

An out-in transition could get stuck if the child with v-if was not yet shown and then replaced with a different child. In this case, "afterLeave" would never get fired, thus "_leaving" stayed true indefinitely. This was likely to happen with vue-router and asynchronous data loading before the new page is shown.

Fixes #6687.

### DIFF
--- a/src/platforms/web/runtime/components/transition.js
+++ b/src/platforms/web/runtime/components/transition.js
@@ -160,7 +160,7 @@ export default {
     if (
       oldChild &&
       oldChild.data &&
-      oldChild.elm.nodeType !== Node.COMMENT_NODE &&
+      oldChild.elm.nodeType !== 8 && // Node.COMMENT_NODE
       !isSameChild(child, oldChild) &&
       !isAsyncPlaceholder(oldChild)
     ) {
--- a/src/platforms/web/runtime/components/transition.js
+++ b/src/platforms/web/runtime/components/transition.js
@@ -160,6 +160,7 @@ export default {
     if (
       oldChild &&
       oldChild.data &&
+      oldChild.elm.nodeType !== Node.COMMENT_NODE &&
       !isSameChild(child, oldChild) &&
       !isAsyncPlaceholder(oldChild)
     ) {
--- a/src/platforms/web/runtime/components/transition.js
+++ b/src/platforms/web/runtime/components/transition.js
@@ -160,10 +160,10 @@ export default {
     if (
       oldChild &&
       oldChild.data &&
-      oldChild.elm &&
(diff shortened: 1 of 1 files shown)


neelance:
Flow is unhappy with `Node.COMMENT_NODE`. I don't have that much experience with flow. What's the best way to fix this?

dsonet:
I guess just use the constant value 8 would be fine.

neelance:
I've changed it to the constant value. One of the tests is still failing. I don't know enough about this code to say if this failure is really something bad or if the behavior is still fine this way and the test could be changed. Can someone with more knowledge on this please help me here?

dsonet:
Seems you have to update the test to reflect your fix.
The failed test is 'async components with transition-mode out-in'.

src/platforms/web/runtime/components/transition.js
```diff
@@ -160,6 +160,7 @@ export default {
     if (
       oldChild &&
       oldChild.data &&
+      oldChild.elm.nodeType !== 8 && // Node.COMMENT_NODE
```
jkzing: `oldChild` could be `undefined`, so check existence should be necessary.
```js
oldChild.elm && oldChild.elm.nodeType !== 8
```

neelance: In which situation would this happen?

jkzing: This is why 'async components with transition-mode out-in' test failing. 🙂

neelance:
@jkzing You are right, this indeed fixed the test. Thanks!

I'm a bit surprised, because the log output of the test failure didn't show any error related to `undefined`.

Is this good to merge now?

neelance:
Any update?

yyx990803:
Thanks for the PR @neelance ! I tweaked the implementation a little bit because the transition component is reused in Weex and should ideally be decoupled from the DOM.

Issue #7048, opened 2017-11-12
[template-compiler] Using both v-bind:value and v-model on an input generates code that'S invalid in IE


LinusBorg:
### Version
2.5.3

### Reproduction link
[https://vuejs.org/v2/guide/render-function.html#Template-Compilation](https://vuejs.org/v2/guide/render-function.html#Template-Compilation)

(Alternative reproduction on jsfiddle: https://jsfiddle.net/Linusborg/pqbppg7a/)

### Steps to reproduce
1. Paste the following piece of HTML into the textarea of the linked documentation chapter above.
```html
<input type="text" v-bind:value="message" v-model="message">
```
2. Inspect the code of the created render function, especially the `domProps` options:
```javascript
domProps: {
  "value": message,
  "value": (message)
},
```

### What is expected?
The code should run in all browsers, or throw an error or warning.

### What is actually happening?
IE11 throws an error: "duplicate properties not allowed in strict mode".

---
I realize that using both v-bind:value and v-model on an input  is not recommended or even sensible, but it doesn't throw an error or warning when done in modern browsers, and doesn't create any problems if the expression is the same for both, so it can easily be overlooked when added accidentally.

Uglify also doesn't help - it doesn't remove the first occurrence that would otherwise be "overwritten" by the second in modern browsers.

So what would be a sensible solution? I'm not sure.

* We could adjust the [codegen of domProps](https://github.com/vuejs/vue/blob/dev/src/compiler/codegen/index.js#L488-L495) to check for a double occurence of the `value` prop on elements of type "input" and drop all but the last occurrence. Possible Problem: performance?
* We could log a warning to console (how to deal with that in vue-loader?)
* We could throw an error, forcing the user to choose between the two (breaking change)
* we could improve awareness by adding a warning to docs and a rule to eslint-plugin-vue?

yyx990803:
I think this can be dealt with by a compile-time warning at the parsing phase (when `v-model` and `v-bind:value` are detected on the same element)

justsrik:
I am getting compiling errors not warnings when I do 'npm run dev' on my code base. @yyx990803 @ @rpemberton @LinusBorg

yyx990803:
You should fix them because they are not supposed to work in the first place.

justsrik:
Thanks for your reply @yyx990803, removed :value. Thanks

PR #7056, opened 2017-11-14
test($compile): #7048 warn if both v-model and v-bind:value used on same element

rpemberton:
Address issue #7048 

I don't have much experience with contributing so any feedback would be very welcome.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -598,6 +598,15 @@ function makeAttrsMap (attrs: Array<Object>): Object {
     }
     map[attrs[i].name] = attrs[i].value
   }
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    map['v-model'] &&
+    (map['v-bind:value'] || map[':value']) &&
+    (!map['type'] || map['type'] === 'text')
+  ) {
+    var vBind = map['v-bind:value'] ? 'v-bind:value' : ':value'
+    warn('v-model and ' + vBind + ' used on the same text input')
+  }
   return map
 }
 
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -51,7 +51,7 @@ export function createASTElement (
     type: 1,
     tag,
     attrsList: attrs,
-    attrsMap: makeAttrsMap(attrs),
+    attrsMap: makeAttrsMap(attrs, tag),
     parent,
     children: []
   }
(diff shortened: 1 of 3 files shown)


src/compiler/parser/index.js
```diff
@@ -598,6 +598,15 @@ function makeAttrsMap (attrs: Array<Object>): Object {
...
     map[attrs[i].name] = attrs[i].value
   }
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    map['v-model'] &&
+    (map['v-bind:value'] || map[':value']) &&
+    (!map['type'] || map['type'] === 'text')
+  ) {
+    var vBind = map['v-bind:value'] ? 'v-bind:value' : ':value'
+    warn('v-model and ' + vBind + ' used on the same text input')
```
yyx990803: We could definitely use a more detailed message here - something like "v-bind:value conflicts with v-model on the same element because the latter already expands to a value binding internally"

src/compiler/parser/index.js
```diff
@@ -598,6 +598,15 @@ function makeAttrsMap (attrs: Array<Object>): Object {
     }
     map[attrs[i].name] = attrs[i].value
   }
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    map['v-model'] &&
+    (map['v-bind:value'] || map[':value']) &&
+    (!map['type'] || map['type'] === 'text')
```
yyx990803: This warning should apply to anything that is not:
- type=checkbox
- type=radio
- select

src/compiler/parser/index.js
```diff
@@ -598,6 +598,16 @@ function makeAttrsMap (attrs: Array<Object>): Object {
...
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    map['v-model'] && (map['v-bind:value'] || map[':value']) &&
+    map['type'] !== 'checkbox' &&
+    map['type'] !== 'radio' &&
+    tag !== 'select'
+  ) {
+    var vBindValue = map['v-bind:value'] ? 'v-bind:value' : ':value'
+    warn(vBindValue + ' conflicts with v-model on the same element because the latter already expands to a value binding internally')
+  }
```
yyx990803: Check looks good now - but I'm not sure if `makeAttrsMap` is the best place to do this check. Let's place it [here](https://github.com/vuejs/vue/blob/dev/src/platforms/web/compiler/directives/model.js#L25) since this warning is about `v-model`.

rpemberton: Sure, it's now been moved

Issue #6877, opened 2017-10-22
Mouse button modifiers for "v-on" should have more precise aliases


st-sloth:
### What problem does this feature solve?
Currently mouse button modifiers, introduced in #4866, are `left`, `middle`, `right` for `$event.button` values `0`, `1`, `2` correspondingly. But that is only correct for the "default" right-handed layout. In the left-handed mode (when the buttons are switched in the OS) the right button has `0` value and the left button - `2`. Besides, also some right-handed users may reverse or remap the mouse buttons. And working with code with these names for mouse buttons may lead to confusion and bugs among developers not familiar with the "default" right-handed mode being the target for these modifiers.

### What does the proposed API look like?
Keep `left` and `right` as deprecated and add `main` and `secondary` as their aliases.

st-sloth:
I am currently working on it. What names should the modifiers really have?

1. Since the "mouse" modifiers for `left` and `right` have already clashed with keyboard modifiers `left` and `right` (#5046, #5060), and to avoid such future name issues, should we perhaps include something pointer device / mouse related like `ptrMain`? Introducing two-word modifiers doesn't seem like a good idea though because of all the case related stuff.
2. What should the main signifiers be?
  * for 0 (now `left`) - `main` / `primary` / your suggestions
  * for 1 (now `middle`) - `aux` (but might be confusing since `auxclick` [is fired for all the mouse buttons except the primary one](https://developers.google.com/web/updates/2016/10/auxclick)) / your suggestions
  * for 2 (now `right`) - `secondary` / your suggestions
  
Previous names `left`, `right`, `middle` of course are kept but deprecated with warnings.

dsbert:
If using `primary` and `secondary`, then `tertiary` would be the consistent name for the third, typically middle, button.

st-sloth:
The discussion kinda moved to the pull request, and [Evan okayed the naming](https://github.com/vuejs/vue/pull/7069#issuecomment-344938670).

PR #7069, opened 2017-11-16
feat(v-on): add more precise names for pointer modifiers

ghost:
re #6877

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Having a warning "uncaught" with `toHaveBeenWarned` fails otherwise valid tests. Adding assertions about warnings seems excessive in those tests. What is the proper way to do it?

Also questions posed in https://github.com/vuejs/vue/issues/6877#issuecomment-344174202 are relevant. Currently names are `primary`, `secondary`, `auxiliary` for `left`, `right`, `middle` correspondingly.

### DIFF
--- a/src/compiler/codegen/events.js
+++ b/src/compiler/codegen/events.js
@@ -3,6 +3,12 @@
 const fnExpRE = /^\s*([\w$_]+|\([^)]*?\))\s*=>|^function\s*\(/
 const simplePathRE = /^\s*[A-Za-z_$][\w$]*(?:\.[A-Za-z_$][\w$]*|\['.*?']|\[".*?"]|\[\d+]|\[[A-Za-z_$][\w$]*])*\s*$/
 
+const keyboardEvents = [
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks for the PR!

I think it makes sense to assert that the warning has indeed been warned whenever it should be. That's the point of the warn check.

Naming looks good.

Final note: we probably won't merge/ship this until 2.6 since we want to avoid introducing deprecations in patch releases.

st-sloth:
Thank you!

I'll add the warn checks in a bit.

2.6 is totally fine. But if the new API is welcome in a patch release while deprecation warnings are not, and if it would be beneficial, I can move deprecation warnings into another PR that would wait for the minor version bump and keep new names with working old ones here for the patch.

PR #7082, opened 2017-11-17
Fix unnecessary conditional

nuoyz:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

https://github.com/vuejs/vue/blob/dev/src/platforms/web/runtime/components/transition.js#L34-L41
If an if block contains a return statement, the else block becomes unnecessary


yyx990803:
Thanks for the PR, but the change is rather stylistic and doesn't have any real benefits.

PR #7089, opened 2017-11-20
chore #7070: warn properties that starts with $ or _

### LINKED ISSUE #7070: Warn when a data property name starts with $ or _
### Version
2.5.3

### Reproduction link
[http://jsfiddle.net/p48cLbe3/3/](http://jsfiddle.net/p48cLbe3/3/)

### Steps to reproduce

```js
new Vue({
  el: '#app',
  data: {
    // invalid, not proxy by vm
    $a: 'a'
  },
  computed: {
    // valid
    $b() {
      return 'b'
    }
  },
  methods: {
    // valid
    $c() {
      console.log('c')
    }
  }
})
```

### What is expected?
Property name starts with $ or _ identifier should all valid or invalid whether defined in data/computed/methods option, i think these options's behaviors should be consistent

It's valid to define $xxx property in computed/methods option, but fall to define $xxx in data option

guieiras:
close #7070

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
As the docs says:

> Properties that start with _ or $ will not be proxied on the Vue instance because they may conflict with Vue’s internal properties and API methods. You will have to access them as vm.$data._property.`

But, this is a little bit hidden during development. It would be better if a warn could be raised to show developer that the data property will not be proxied on the Vue instance.

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -142,12 +142,14 @@ function initData (vm: Component) {
         vm
       )
     } else if (isReserved(key)) {
-      warn(
-        `Property "${key}" will not be proxied on the instance because it ` +
-        `may conflict with Vue's internal properties. You will have to access it as ` +
-        `"vm.$data.${key}". Avoid defining component properties that start with _ or $.`,
-        vm
-      )
+      if (process.env.NODE_ENV !== 'production') {
+        warn(
+          `Property "${key}" will not be proxied on the instance because it ` +
+          `may conflict with Vue's internal properties. You will have to access it as ` +
+          `"vm.$data.${key}". Avoid defining component properties that start with _ or $.`,
+          vm
+        )
+      }
     } else {
       proxy(vm, `_data`, key)
     }
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -141,7 +141,14 @@ function initData (vm: Component) {
         `Use prop default value instead.`,
         vm
       )
-    } else if (!isReserved(key)) {
+    } else if (isReserved(key)) {
+      warn(
+        `Property "${key}" will not be proxied on the instance because it ` +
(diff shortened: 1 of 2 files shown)


src/core/instance/state.js
```diff
@@ -141,7 +141,14 @@ function initData (vm: Component) {
         `Use prop default value instead.`,
         vm
       )
-    } else if (!isReserved(key)) {
+    } else if (isReserved(key)) {
+      warn(
```
posva: warn must be shown only in development mode (process.env.NODE_ENV !== production)

guieiras:
@posva How about this broken spec? And, as you said on the issue, isn't a little bit risky to proceed with this change as some users/plugins uses internally this kind of data?

src/core/instance/state.js
guieiras: fixed

posva:
The broken test is using `$foo` and btw, is wrong because it should ready `$foo` from `this.$data` and not from `this` but it doesn't change anything having `undefined` or `{}` as the injected value.

I think I went too fast saying we could safely add a warning for this, sorry about that.
If users want to add properties to `data` that start with `$` or `_` they're free to do so, they only need to know that in order to access them, they have to use `this.$data`
I don't have any better idea that just drop this 🙁
@chrisvfritz @sdras @LinusBorg do you have any feedback on how to improve this in case you have seen people falling under this _caveat_?

guieiras:
@posva I was willing to contribute and I didn’t think about the consequences or how it could impact users. I’m sorry 😕.

Well, I think we can close this. Thanks for your review and the discussion.

posva:
No worries, as I said, I should have payed more attention 🙂

posva:
Let's close this for the moment

PR #7110, opened 2017-11-22
feat: add browser ESM build

yyx990803:
Adding a new dist file `vue.esm.browser.js` which enables direct import in ESM-enabled browsers:

``` html
<script type="module">
import Vue from 'https://unpkg.com/vue/dist/vue.esm.browser.js'

new Vue({
  el: '#app',
  data: {
    msg: 'hello!'
  }
})
</script>
```

### DIFF
--- a/build/config.js
+++ b/build/config.js
@@ -58,11 +58,21 @@ const builds = {
     format: 'es',
     banner
   },
-  // Runtime+compiler CommonJS build (ES Modules)
+  // Runtime+compiler ES modules build (for bundlers)
   'web-full-esm': {
     entry: resolve('web/entry-runtime-with-compiler.js'),
     dest: resolve('dist/vue.esm.js'),
     format: 'es',
+    alias: { he: './entity-decoder' },
+    banner
+  },
+  // Runtime+compiler ES modules build (for direct import in browser)
+  'web-full-esm-browser': {
+    entry: resolve('web/entry-runtime-with-compiler.js'),
+    dest: resolve('dist/vue.esm.browser.js'),
+    format: 'es',
+    transpile: false,
+    env: 'development',
     alias: { he: './entity-decoder' },
     banner
   },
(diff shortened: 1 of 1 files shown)


tvanier:
Thanks @yyx990803 , the new `vue.esm.browser.js` looks good, should it be mentioned in [dist/README.md](https://github.com/vuejs/vue/blob/dev/dist/README.md)?
Is it expected the new dist file is already included in the vue package (before this PR is merged)?
https://unpkg.com/vue@2.5.16/dist/vue.esm.browser.js
`dist/vue.esm.browser.js` seems to exist since version `2.5.9`, and its internal version is always `2.5.8`?

damianpeterson:
Is this meant to be for development only? I see things like `productionTip: "development" !== 'production'` in the compiled file.

simonhaenisch:
Having the same question as @damianpeterson: will you add a `vue.esm.browser.min.js` production build as well?

from https://github.com/vuejs/vue/tree/dev/dist#development-vs-production-mode:

> CommonJS and ES Module builds are intended for bundlers, therefore we don't provide minified versions for them. You will be responsible for minifying the final bundle yourself.

But with ESM in the browser we won't necessarily need bundlers anymore (depending on the project and its audience). For what I'm working on, I don't need template pre-compilation and runtime-only, but I would prefer a minified version (`vue.min.js` is 84 KB compared to `vue.js` with 284 KB).

TimvdLippe:
I am already using this bundle in my project and can say that is working as expected. I am not sure why this file is already available after `npm install vue`, but I can't complain :smile:

megacromulent:
Is there a way we could get a beta build of the esm browser version of this for internal testing purposes? (ie, as beta testers perhaps?)

yyx990803:
Landed as f2fd44b7

zigomir:
Seems like latest versions aren't published as ES modules anymore. Opening Vue.js https://unpkg.com/vue@2.5.21/dist/vue.esm.browser.js says it's a `v2.5.8` but should be `2.5.21`, right?

Or is it only wrong comment "banner"? I believe using this build with dev tools also picked `2.5.8`.

yyx990803:
@zigomir these files were actually published by mistake - they were built on a different branch and accidentally published. Actual ESM browser build won't land until 2.6.

PR #7121, opened 2017-11-23
feat(core): support KeyboardEvent.key for built in keyboard event modifiers, fix #6900

### LINKED ISSUE #6900: Native event and VueJs Event modifier
### Version
2.5.2

### Reproduction link
[https://jsfiddle.net/klesk/0x4q5kup/1/](https://jsfiddle.net/klesk/0x4q5kup/1/)

### Steps to reproduce
A/ Press enter in the input

B/ Check the checkbox (that emultae keyboard event)

### What is expected?
A/  the classList should appear

B/ the classList should appear

A/  the classList of it appear

B/ nothing appear

---
Remove the modifier in the template (eg: @keyup) and the checkbox raise event.

Is that normal? Or a bug ?

jkzing:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
#6900

### DIFF
--- a/src/compiler/codegen/events.js
+++ b/src/compiler/codegen/events.js
@@ -16,18 +16,6 @@ const keyCodes: { [key: string]: number | Array<number> } = {
   'delete': [8, 46]
 }
 
-const keyNames: { [key: string]: string | Array<string> } = {
-  esc: 'Escape',
-  tab: 'Tab',
-  enter: 'Enter',
-  space: ' ',
-  up: 'ArrowUp',
-  left: 'ArrowLeft',
-  right: 'ArrowRight',
-  down: 'ArrowDown',
-  'delete': ['Backspace', 'Delete']
-}
-
 // #4868: modifiers that prevent the execution of the listener
 // need to explicitly return null so that we can determine whether to remove
 // the listener for .once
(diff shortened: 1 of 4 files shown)


src/compiler/codegen/events.js
```diff
@@ -16,6 +16,18 @@ const keyCodes: { [key: string]: number | Array<number> } = {
   'delete': [8, 46]
 }
 
+const keyNames: { [key: string]: string | Array<string> } = {
```
dsonet: Why not just move the keyNames and keyCodes to runtime instead of in the compiler? Since it leads sooo much repetition and should reduce the code size.

jkzing: Actually in my first version, I did define `keyNames` in `check-keycodes.js`. After that I realize this would increase much runtime bundle size, so I got it moved into compiler...

jkzing: I guess the reason `keyCodes` stays in compiler is that it simply generate numbers rather strings. Agreed for moving `keyNames` into runtime.🙂

dsonet: Considering all compiled templates size together, it could reasonable to move them into runtime. BTW, seems it has not that much different to deal with keyCodes and keyNames.

PR #7151, opened 2017-11-29
(keep-alive): do not use the cache which has been destroyed

wanyaxing:


wanyaxing:
Sorry, I don't know what happened in the test of 'lint-flow-types '.

I can't fix this fail test.

posva:
Please, do not remove the template of the PR 🙁 
Could you add it back, it really helps us. (if you cannot find it, you can try closing this pr and creating a new one, you may need to `git push -f` to create a different commit and allow github to create a new PR). The checklist there is also necessary. For example, we need a test for every fix or feature that is added. For instance, here we don't clearly see what is the bug you're trying to fix

wanyaxing:
I only added some codes in src/core/components/keep-alive.js, I don't know what happened in the template as you said.
I will try to do the new PR .
Thanks.

leixu2txtek:
@wanyaxing  PR try?

wanyaxing:
@leixu2txtek  https://github.com/vuejs/vue/issues/6509#issuecomment-403693256

PR #7198, opened 2017-12-07
fix(props): handle misspelled keys on prop validation object

Alex-Sokolov:

- [x] Bugfix

- [ ] Yes
- [x] No

**Other information:**

When using object for validating props, there are no checks about object content. This may take a long time before noticing that you forget to pass required prop, but there are no warnings with a misprint.

Fail cases:
- Typo in key name `deafult: false` or `require: true`
- Using non-english letter in key name (for example, russian)

P.S.: Will be great to add instance sections names validation, but I have no idea how to solve this (once I spent a lot of time before see that writed `component: {}` section instead `components: {}`).

### DIFF
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -94,7 +94,7 @@ function getPropDefaultValue (vm: ?Component, prop: PropOptions, key: string): a
 /**
  * Assert whether a prop object keys are valid.
  */
-function assertPropObject (
+export function assertPropObject (
   prop: Object,
   key: string,
   vm: ?Component
@@ -119,7 +119,6 @@ function assertProp (
   vm: ?Component,
   absent: boolean
 ) {
-  assertPropObject(prop, name, vm)
   if (prop.required && absent) {
     warn(
       'Missing required prop: "' + name + '"',
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -18,6 +18,13 @@ type PropOptions = {
   validator: ?Function
 };
 
+const propOptionsNames = [
+  'type',
+  'default',
+  'required',
+  'validator'
+]
+
 export function validateProp (
   key: string,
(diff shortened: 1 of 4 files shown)


src/core/util/props.js
```diff
@@ -94,6 +119,7 @@ function assertProp (
   vm: ?Component,
   absent: boolean
 ) {
+  assertPropObject(prop, name, vm)
```
sqal: Here you need to add production environment check so this function can be removed by minifier, [like here](https://github.com/Alex-Sokolov/vue/blob/bb8a124214b2d2d19162f65e903b6e42df57e0dc/src/core/util/props.js#L55)

yyx990803:
Thanks for the PR. Placing the check in `assertProp` incurs more runtime cost than necessary (even only in dev mode) as `assertProp` is also called on every re-render of the component. This check only needs to be done once in `initProps`.

Alex-Sokolov:
@yyx990803 Thanks for info — moved to `initProps`!

yyx990803:
Thanks for the update - as I was reviewing again I realized the runtime cost can be further reduced by placing the check [here](https://github.com/vuejs/vue/blob/dev/src/core/util/options.js#L297) when the options are normalized... (performed once per constructor)

Also, even though this is dev-only code, I'd prefer to refactor to use a `for...in` loop to get rid of the iteration functions.

Alex-Sokolov:
@yyx990803 done

mattixittam:
This warns about anything which relies on adding custom attributes to the prop object. For instance this: https://github.com/KoRiGaN/Vue2Leaflet/issues/109.

In another use-case, we add a `note` attribute to props for generating documentation.

I would suggest making this warning configurable somehow.

JounQin:
A custom key `meta` like `vue-router` would be enough.

PR #7224, opened 2017-12-12
fix(compiler): fix double escaping of ssrNode attribute values (fix #7223)

### LINKED ISSUE #7223: vue-template-compiler double-escaping newlines in _ssrNode text string templates
### Version
2.5.9

### Reproduction link
[https://gist.github.com/2b3c1a683b334022c626917b0abb2adc](https://gist.github.com/2b3c1a683b334022c626917b0abb2adc)

(I used a Gist, because ssrCompile is not exposed by the globals build of Vue.)

### Steps to reproduce
```javascript
compiler = require('vue-template-compiler')
template = "<div id=\"foo,\nbar\">\n<textarea id=\"foo,\nbar\">foo\nbar</textarea>\n</div>"
result = compiler.ssrCompile(template)
console.log(result.render)
// output: with(this){return _c('div',{attrs:{"id":"foo,\nbar"}},[_ssrNode("<textarea id=\"foo,\\nbar\">foo\nbar</textarea>")])}
```

### What is expected?
vue-template-compiler should not double-escape newlines in attributes for _ssrNode string templates.

vue-template-compiler double-escapes newlines in attributes of _ssrNodes.

---
If you look at the output from the reproduction steps, you will see that the newline character '\n' is only double escaped, if it is part of an attribute in an ssrNode, if is nor optimized to a string template and stored in the attrs property, is it fine and other newlines that are outside attributes are also fine.

The bug was observed with vue-template-compiler v2.5.9 and node.js v8.9.1.

While the above example seems silly, newlines in attribute are very common when using responsive images, with multiple sizes in the srcset attribute, which are usually broken onto one line per size/url for readability and this is how a colleague of mine ran into this bug, where responsive images were working differently between ssr and rehydration due to the problem, because the browser aborted srcset parsing when finding the "\n" literal string.

felixbuenemann:
This fixes a double escaping of attribute values in the SSR optimizing
compiler, because literal attribute values get escaped early during the
call to `addAttr` inside `processAttrs` before it is known, if this
attribute will be serialized to an _ssrNode string template.

Later on the _ssrNode string template gets escaped again to preserve
whitespace and this causes double escaping of the whitespace inside the
attribute value.

This approach fixes the problem by undoing the escaping by parsing the
attribute value JSON in `genAttrSegment`, which is the easiest fix to
the problem, but probably not the best.

This fixes #7223 at least for the cases where I encountered the problem.

See #7223 for an in-depth description of the problem and a test-case.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

I would love to add a breaking test, that shows that the fix is working, but I'm not sure were to add it with the large test suite, so some advise on that would be appreciated.

### DIFF
--- a/src/server/optimizing-compiler/modules.js
+++ b/src/server/optimizing-compiler/modules.js
@@ -77,7 +77,7 @@ function genAttrSegment (name: string, value: string): StringSegment {
         ? ` ${name}="${name}"`
         : value === '""'
           ? ` ${name}`
-          : ` ${name}=${value}`
+          : ` ${name}="${JSON.parse(value)}"`
     }
   } else {
     return {


yyx990803:
Thanks and good catch!

felixbuenemann:
Thanks for taking care of the missing test in 86e4d7566e685f1a98ffab1777fa7d07603c994a.

PR #7272, opened 2017-12-19
feat(weex): adjust framework entry APIs and add flow annotations

Hanks10100:

- [x] Refactor

- [x] No

**Other information:**

### Remove legacy framework APIs

Remove the `init`, `reset`, `getRoot` and `receiveTasks` APIs, which are already implemented in the new version of `weex-js-runtime`.

Moreover, the deprecated timer APIs should also be removed when they were supported by native render engines (C++ timers).

### No longer execute the js code at front-end

Use `createInstanceContext` instead of `createInstance` to create the Vue module instance for each page. The Vue framework only needs to prepare the instance context and no longer execute the js code. The js code will be executed by native render engines.

### Flow type declarations

Add more reliable flow type declarations of Weex instance APIs and DOM APIs, and add type annotations for the entry file.

### DIFF
--- a/flow/weex.js
+++ b/flow/weex.js
@@ -1,6 +1,79 @@
 // global flag to be compiled away
 declare var __WEEX__: boolean;
 
+declare type Weex = {
+  config: Object;
+  document: WeexDocument;
+  requireModule: (name: string) => Object | void;
+  supports: (condition: string) => boolean | void;
+  isRegisteredModule: (name: string, method?: string) => boolean;
+  isRegisteredComponent: (name: string) => boolean;
+};
+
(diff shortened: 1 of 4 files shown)


Hanks10100:
This feature is used to achieve multi-instance isolation of Weex. Weex is planning to execute each page's code in different js context. Source codes and js contexts will be managed by the customed js engine within Weex core.

See the discussions on Weex's mailing list: https://lists.apache.org/thread.html/87bf83253ed7ee318a8cf98d693bc534b3e08b4d28655a1290f2e0c7@%3Cdev.weex.apache.org%3E

PR #7291, opened 2017-12-20
feat(weex): support to parse object literal in binding attributes or styles

Hanks10100:
- [x] Feature

- [x] No

**Other information:**

Required `acorn` and `escodegen` in `weex-template-compiler` to compile expression strings into specified weex native directives. (https://github.com/Hanks10100/weex-native-directive/issues/11)

### DIFF
--- a/src/platforms/weex/util/parser.js
+++ b/src/platforms/weex/util/parser.js
@@ -1,0 +1,60 @@
+/* @flow */
+
+// import { warn } from 'core/util/index'
+
+// this will be preserved during build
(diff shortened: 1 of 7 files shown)


src/platforms/weex/util/parser.js
```diff
@@ -0,0 +1,60 @@
+/* @flow */
+
+// import { warn } from 'core/util/index'
```
Hanks10100: This code is commented because jasmine seems not support using alias.

PR #7302, opened 2017-12-21
fix(core): fix defineReactive unnecessary getters call fix #7280

### LINKED ISSUE #7280: Data getters are being invoked on component initialisation
### Version
2.5.11

### Reproduction link
[https://jsfiddle.net/7gmh5wL8/1/](https://jsfiddle.net/7gmh5wL8/1/)

### Steps to reproduce
* Run JSFiddle snippet 
* Open browser console
* You'll see a warning from property getter

### What is expected?
I expect, that my data and props enumerable properties will not be invoked, without my explicit call for them. 

Seems, that `defineReactive` function in `/src/core/observer/index.js` works that way, that it accepts an object property value, instead of getting it from the getter on demand. Despite the fact, that it uses `getOwnPropertyDescriptor` to respect user-defined getters and setters. 

---
I faced this problem in a context of lazy-loaded backend entities. I have some redux store selectors, that returns RESTfull models. Some of the models nested properties are wrapped into getters, so I can call them, if needed and trigger a background api request for additional info. Now, all of the getters are invoked immediately and a bunch of requests is sent after initial rendering.

I can do a pull request, to fix this, but may be I do not understand some concept under this behaviour. 

Sorry for bad english.

DeyLak:
The defineReactive gets the property of the object, only if it has no getter

fix #7280

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -145,7 +145,7 @@ export function defineReactive (
 
   // cater for pre-defined getter/setters
   const getter = property && property.get
-  if (!getter && arguments.length <= 2) {
+  if (!getter && arguments.length === 2) {
     val = obj[key]
   }
   const setter = property && property.set
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -61,7 +61,7 @@ export class Observer {
   walk (obj: Object) {
     const keys = Object.keys(obj)
     for (let i = 0; i < keys.length; i++) {
-      defineReactive(obj, keys[i], obj[keys[i]])
+      defineReactive(obj, keys[i])
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/core/observer/index.js
```diff
@@ -145,6 +145,9 @@ export function defineReactive (
 
   // cater for pre-defined getter/setters
   const getter = property && property.get
+  if (!getter && arguments.length <= 2) {
```
dsonet: If so, it's better to be `arguments.length === 2`.

PR #7356, opened 2018-01-02
added support for parsing of space delimitted paths

lopugit:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/src/core/util/lang.js
+++ b/src/core/util/lang.js
@@ -23,7 +23,7 @@ export function def (obj: Object, key: string, val: any, enumerable?: boolean) {
 /**
  * Parse simple path.
  */
-const bailRE = /[^\w.$]/
+const bailRE = /[^\w.\s$]/
 export function parsePath (path: string): any {
   if (bailRE.test(path)) {
     return


HerringtonDarkholme:
Thanks for your contribution.

However, as its name indicates, `parsePath` only handle path name delimited by dot, like `obj.pathName.noSpace`. Bracket indexing expression `ob['some pathh']` isn't supported. So we cannot accept space in path.

lopugit:
@HerringtonDarkholme how is it not supported though? How do you access the dot delimited properties? Isn't bracket index expression the only way? I've been using it for a few hours and it's working fine for me, I can imagine there might be some breaking bugs but I thought that since you're parsing a string into paths all you do is split on a dot, I wrote a library that uses the same principle, I can't see any reason to be splitting on anything but a dot and doing anything else with the property strings, split or not split, do they get processed at all besides splitting on dot?

HerringtonDarkholme:
`grep -R parsePath src -A 9`

```
src/core/util/lang.js:export function parsePath (path: string): any {
src/core/util/lang.js-  if (bailRE.test(path)) {
src/core/util/lang.js-    return
src/core/util/lang.js-  }
src/core/util/lang.js-  const segments = path.split('.')
src/core/util/lang.js-  return function (obj) {
src/core/util/lang.js-    for (let i = 0; i < segments.length; i++) {
src/core/util/lang.js-      if (!obj) return
src/core/util/lang.js-      obj = obj[segments[i]]
src/core/util/lang.js-    }
--
src/core/observer/watcher.js:      this.getter = parsePath(expOrFn)
src/core/observer/watcher.js-      if (!this.getter) {
src/core/observer/watcher.js-        this.getter = function () {}
src/core/observer/watcher.js-        process.env.NODE_ENV !== 'production' && warn(
src/core/observer/watcher.js-          `Failed watching path: "${expOrFn}" ` +
src/core/observer/watcher.js-          'Watcher only accepts simple dot-delimited paths. ' +
src/core/observer/watcher.js-          'For full control, use a function instead.',
src/core/observer/watcher.js-          vm
src/core/observer/watcher.js-        )
src/core/observer/watcher.js-      }
```

lopugit:
@HerringtonDarkholme 

`obj = obj[segments[i]]` So it's using bracket notation?

HerringtonDarkholme:
> The expression only accepts simple dot-delimited paths. For more complex expression, use a function instead.

Let me clarify what does this mean. `parsePath` is only used in `$watch` and `$watch`'s string argument only accepts simple dot-delimited path as convienence.

Dot-delimited path is a simple and valid expression in JavaScript, while `obj.some path` isn't.
Whether it works coincidentally with current implementation or not doesn't matter whether it is appropriate to be accepted as a public API.

Consider, if space can be accepted as object path, why minus, divide cannot? I can conceive a lot of CSS-in-JS users will use `background-color` as key name. Accepting non identifier symbol just breaks the simple concept of dot-delimiteed path.

Please do use function as watcher expression. Or, you can use your own wrapper function in your application.

```ts
function p(path) {
   const segments = path.split('.')
   return function () {
     const obj = this
     for (let i = 0; i < segments.length; i++) {
       if (!obj) return
       obj = obj[segments[i]]
     }
     return obj
   }
}

this.$watch(p('my space.my api.my data'), handler)
// only three more characters, two more key strokes if a decent editor is used
```

lopugit:
@HerringtonDarkholme ohh you can actually pass the object straight through, that works well, I'm using a get function like that already to build objects out of strings, hence why I was surprised the watch function didn't allow spaces.

In my opinion having spaces is pretty awesome for readability and in general. 

I don't believe it's a non-identifier symbol is it? I just read up on the variable name javascript docs a little

yyx990803:
`$watch`'s expression parse is intentionally limited - it's meant as a convenience API only, so anything more complex than dot-delimited paths should use a function getter instead.

lopugit:
@yyx990803 @HerringtonDarkholme just coming back to this, shouldn't watching include any valid javascript property path? expression parse just has to separate on dots, that's literally all, and it would support minuses, plusses, spaces, etc... ? if I can do ```data(){ return { "some property" : true } }``` I should be able to do ```{ watch: { "some property" } }```

come on?

Justineo:
> if I can do `data(){ return { "some property" : true } }` I should be able to do `{ watch: { "some property" } }`

What if you have this:

```js
{
  data () {
    return {
      'some.property': true,
      some: {
        property: false
      }
    }
  },
  watch: {
    'some.property' (val) {
      // ?
    }
  }
}
```

lopugit:
> > if I can do `data(){ return { "some property" : true } }` I should be able to do `{ watch: { "some property" } }`
> 
> What if you have this:
> 
> ```js
> {
>   data () {
>     return {
>       'some.property': true,
>       some: {
>         property: false
>       }
>     }
>   },
>   watch: {
>     'some.property' (val) {
>       // ?
>     }
>   }
> }
> ```

yeah that's a good point but who in their right mind uses dots in their property names :')

Alinawaz-786:
do you have any idea to use the love streaming api in laravel php

lopugit:
@Alinawaz-786 what?

Alinawaz-786:
Live streaming Videos

Anyone have any idea to live streaming videos api

any how to use this API

lopugit:
@Alinawaz-786 this is not the place to ask about that? I don't know sorry.

Issue #5443, opened 2017-04-14
get return value of $emit event's callback


rainjay:
### Version
2.2.6

### Reproduction link
[https://jsfiddle.net/50wL7mdz/27555/](https://jsfiddle.net/50wL7mdz/27555/)

### Steps to reproduce
```
<component @click="callback">
</component>

function callbcak() {
  return new Promise(function(resolve, reject) {
    setTimeout(() => {
      resolve('resolved')
    }, 2000)
  })
}

// in component
<script>
  ...
    methods: {
      handleClick(evt) {
        var promise = this.$emit('click', evt)
        console.log(promise)   // promise is component self, not the return promise
      }
    }
  ...
</script>
```

### What is expected?
expect $emit get the return value of event's callback

### What is actually happening?
get component self

yyx990803:
This would be a breaking change. Just emit a callback.

alexsandro-xpt:
+1

dlongley:
> This would be a breaking change. Just emit a callback.

An alternative would be to use a common pattern from Web APIs where you provide a method on the emitted event called something like `waitUntil` that takes a promise.

For example:

```js
<component @click="callback">
</component>

function callback(evt) {
  evt.waitUntil(new Promise(function(resolve, reject) {
    setTimeout(() => {
      resolve('resolved')
    }, 2000)
  }))
}

// in component
<script>
  ...
    methods: {
      handleClick(evt) {
        var promise = Promise.resolve()
        evt.waitUntil = p => p = promise
        this.$emit('click', evt)
        console.log(promise)   // promise is a Promise
      }
    }
  ...
</script>
```

If desired, you could get more robust and throw an error if `waitUntil` is called more than once.

alexsandro-xpt:
Nice vision @dlongley, you see beyond to the edge!!

sirlancelot:
You could also just pass the `resolve` parameter directly to `$emit` like so:

```js
methods: {
    handleClick(evt) {
        var result = new Promise((resolve) => this.$emit('click', evt, resolve))

        result.then((value) => console.log(value))

        return result
    }
}
```

That way you're not mutating an object that you didn't create ;)

dlongley:
@sirlancelot,

This forces the receiver to call `resolve` in order to achieve proper operation and eliminates the ability for them to reject with sane (and automatic) error propagation.

[...]

PR #7465, opened 2018-01-17
feat(core): add 'Vue.prototype.$notify'

CarterLi:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This PR adds a new function `Vue.prototype.$notify`, which works basically like `$emit`, except it returns an array that collects all return values of its event callbacks.

**Motivation:**

Developers of an component may want to collect data from callbacks of an event. There are a few options to do this:

1. Bind a function directly into the component. But it forces users of the component to write a function or needs the creator of the component writes property type checks to determine whether the prop value is callable or not.
2. Pass a callback parameter for the event. But nobody wants callbacks today.

To solve this issue, I'd like `$emit` or something else returns an array that collects all return values of its event callbacks. In order not to break old code, the behavior of `$emit` was kept ( which returns this ), but I wrote a new function instead. The function name `$notify` is not decided, but I can't come up with a better name.

Related issue: https://github.com/vuejs/vue/issues/5443
Related question: https://segmentfault.com/q/1010000011143372

### DIFF
--- a/flow/component.js
+++ b/flow/component.js
@@ -43,6 +43,7 @@ declare interface Component {
   $on: (event: string | Array<string>, fn: Function) => Component;
   $once: (event: string, fn: Function) => Component;
   $off: (event?: string | Array<string>, fn?: Function) => Component;
+  $notify: (event: string, ...args: Array<mixed>) => Array<mixed>;
   $emit: (event: string, ...args: Array<mixed>) => Component;
   $nextTick: (fn: Function) => void | Promise<*>;
   $createElement: (tag?: string | Component, data?: Object, children?: VNodeChildren) => VNode;
--- a/src/core/instance/events.js
+++ b/src/core/instance/events.js
@@ -111,7 +111,7 @@ export function eventsMixin (Vue: Class<Component>) {
     return vm
   }
 
-  Vue.prototype.$emit = function (event: string): Component {
+  Vue.prototype.$notify = function (event: string): Array<mixed> {
     const vm: Component = this
     if (process.env.NODE_ENV !== 'production') {
       const lowerCaseEvent = event.toLowerCase()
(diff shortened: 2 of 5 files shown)


sirlancelot:
I'm not a huge fan of this because it is impossible to know the order in which the callbacks get executed.

When I encounter a situation where I want to get something back from an `$emit`. I usually pass a callback or a deferred as part of the arguments like so:

```javascript
// callback can be called multiple times by different consumers...
this.$emit("get-something", arg, (value) => {
    // Do stuff with value
})
```

Use it like so:

```javascript
this.$on("get-something", (arg, done) => {
    // Send a value back to the emitting component
    if (done) done("hello from consumer")
})
```

This manner creates a more predictable outcome where the developer decides whether to send a value back. It also allows for an alternative situation where you can only listen for a single value using a deferred like so:

```javascript
var callback
const deferred = new Promise((resolve) => { callback = resolve })

// callback will only save the value of the first caller
this.$emit("get-something", arg, callback)

deferred.then((value) => {
    // Do something with resolved value
})
```

yyx990803:
Thanks for the PR - however, I agree w/ @sirlancelot and I think explicit callbacks is more straightforward.

CarterLi:
It's 2018. I don't think callbacks are a good design anymore.

What I want to write is:

```js
const values = await Promise.all(vm.$emit('event', arg))
```

Instead of

```js
var promises = []
vm.$emit('event', arg, promise => {
  promises.push(promise)
})
const value = await Promise.all(promises)
```

return vm itself for $emit is almost useless. It's not jQuery or lodash and you don't need to chain the operations in 99% cases. But it's a common use case for return values from event handler, for example filtering list remotely. Using return value instead of vender callbacks is much easier to use and a rather modern design nowadays.

> I'm not a huge fan of this because it is impossible to know the order in which the callbacks get executed.

Values in the returned array of `$notify` are still ordered. Besides I think most of the use cases is `@event="handler"` thus no order is needed.

What you get using callback is just `the callback is called BEFORE executing the next handler`, but you get no benefit from this. You can't stop the event bus ( by throwing exceptions or something else ). If you want to return something from the callback, it must be a design error.

@yyx990803 @sirlancelot any thoughts?

yyx990803:
Well... It's 2018. I don't think event-ping-ponging is a good design anymore. **It makes the flow of your application very hard to reason about for those who are not familiar with the codebase**. I don't want to encourage this kind of patterns. If you insist, you don't have to get this into Vue core to use it. [Just write your own plugin](https://vuejs.org/v2/guide/plugins.html).

CarterLi:
https://segmentfault.com/q/1010000012977867

Another one @yyx990803

PR #7491, opened 2018-01-20
fix(core): Use native bind function instead of own (fix #7408)

### LINKED ISSUE #7408: Bind implementation is slower when native in modern browsers

dmitry-vakhnenko:
Simple refactor bind function for easy revert.
Custom bind implementation performance is lower than native. ( https://jsperf.com/vue-bind-perf )

issue link: https://github.com/vuejs/vue/issues/7408

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -174,7 +174,7 @@ export const hyphenate = cached((str: string): string => {
 })
 
 /**
- * Bind funciton context
+ * Bind function context
  */
 export function bind (fn: Function, ctx: Object): Function {
   if(!Function.prototype.bind) {
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -174,20 +174,10 @@ export const hyphenate = cached((str: string): string => {
 })
 
 /**
- * Simple bind, faster than native
+ * Bind funciton context
  */
 export function bind (fn: Function, ctx: Object): Function {
-  function boundFn (a) {
-    const l: number = arguments.length
-    return l
-      ? l > 1
-        ? fn.apply(ctx, arguments)
-        : fn.call(ctx, a)
-      : fn.call(ctx)
-  }
-  // record original fn length
-  boundFn._length = fn.length
-  return boundFn
+  return fn.bind(ctx)
 }
 
 /**
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -174,9 +174,13 @@ export const hyphenate = cached((str: string): string => {
 })
 
 /**
- * Bind function context
+ * Simple bind polyfill for environments that do not support it... e.g.
+ * PhantomJS 1.x. Technically we don't need this anymore since native bind is
+ * now more performant in most browsers, but removing it would be breaking for
+ * code that was able to run in PhantomJS 1.x, so this must be kept for
+ * backwards compatibility.
  */
-function ownBind (fn: Function, ctx: Object): Function {
+function polyfillBind (fn: Function, ctx: Object): Function {
   function boundFn (a) {
     const l = arguments.length
     return l
@@ -190,11 +194,13 @@ function ownBind (fn: Function, ctx: Object): Function {
   return boundFn
 }
 
-function nativeBind(fn: Function, ctx: Object): Function {
+function nativeBind (fn: Function, ctx: Object): Function {
   return fn.bind(ctx)
 }
 
-export const bind = Function.prototype.bind ? nativeBind : ownBind
+export const bind = Function.prototype.bind
+  ? nativeBind
+  : polyfillBind
 
 /**
  * Convert an Array-like object to a real Array.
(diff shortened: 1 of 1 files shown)


yyx990803:
Some browsers in the support range e.g. PhantomJS don't have `Function.prototype.bind`, so we should only use the native version when it exists.

OEvgeny:
Maybe it's better to decide at first module run which implementation to use? e.g.:
```js
export const bind = Function.prototype.bind ? nativeBind : bind
```

dmitry-vakhnenko:
> Maybe it's better to decide at first module run which implementation to use? e.g.:

```JavaScript
const ownBind = (fn: Function, ctx: Object) => {
  function boundFn(a) {
    const l = arguments.length
    return l
      ? l > 1
        ? fn.apply(ctx, arguments)
        : fn.call(ctx, a)
      : fn.call(ctx)
  }

  boundFn._length = fn.length
  return boundFn
}

export const bind = Function.prototype.bind ? Function.prototype.bind : ownBind
```

dmitry-vakhnenko:
> Some browsers in the support range e.g. PhantomJS don't have Function.prototype.bind, so we should only use the native version when it exists.

Why we should carry polyfill for this? Phantomjs is a `virtual` browser that can be used for `tests` and scrapping. We carry polyfill for case when any vuejs site can be scrapped?

We increase bundle size for every build for ^ case?

I just didn't understood.

OEvgeny:
I think
> ```js
> ...
> export const bind = Function.prototype.bind ? Function.prototype.bind : ownBind
> ```

Should be:
```js
function ownBind(fn: function, ctx: object) {
  // internal/current implementation ...
}

function nativeBind(fn: function, ctx: object) {
  return fn.bind(ctx)
}

export const bind = Function.prototype.bind ? nativeBind : ownBind
```

Because of how native `bind` should be called.

dmitry-vakhnenko:
So maybe wrap polyfill version in `if (process.NODE_ENV == 'development')` ? its removes dead code  in production builds and broke dead some e2e tests and scrappers.

I don't see its as as breaking change.

yyx990803:
**It is a breaking change if you remove a polyfill.**

Someone's tests that were able to run in PhantomJS would break suddenly in a patch release because of that.

dmitry-vakhnenko:
> It is a breaking change if you remove a polyfill.

Then its ready for PR.

yyx990803:
I think @OEvgeny 's suggestion should be applied so that we only check native bind once on initialization.

src/shared/util.js
```diff
@@ -174,22 +174,28 @@ export const hyphenate = cached((str: string): string => {
...
       : fn.call(ctx)
   }
-  // record original fn length
+
   boundFn._length = fn.length
   return boundFn
 }
 
+function nativeBind(fn: Function, ctx: Object): Function {
+  return fn.bind(ctx)
```
dsonet: It's missing the optional arguments, cause different behavior to ownBind

OEvgeny: I see no optional arguments in `ownBind`. Are you about `fn._length`? If yes. It shouldn't be a problem because we should check both `length` and `_length` since we don't know exactly was the `fn` bound via our `ownBind` or not. e.g.: https://github.com/vuejs/vue/blob/7cf188e134fe7bfc9e8a16b763fb85b18eff1eac/src/platforms/weex/runtime/modules/transition.js#L73

rlightner:
I think this broke something. if fn.bind is undefined it blows up.

yyx990803:
@rlightner please provide a reproduction. `fn.bind` is only used when `Function.prototype.bind` exists. `fn.bind` cannot be undefined in that case unless `fn` is not a function.

rlightner:
I'll try. The fn being passed in is a method in a template that contains:

```
drawer: {
    get() {
        return this.$store.state.drawer
    }
    , set(value) {
        this.$store.commit('toggleDrawer', value)
    }
}
```
and it's being passed in to `nativeBind` as the get/set. That doesn't have a bind(). This is in a normal vue template.

image

rjoo:
As @rlightner pointed out, this exposed a coding error if whatever's provided to the `bind` helper is not a function.

Version 2.5.13 https://jsfiddle.net/ujxu9mvd/ Renders fine, fails silently
Version 2.5.14 https://jsfiddle.net/bad4a8mb/1/ Does NOT render and it throws an Uncaught TypeError

PR #7529, opened 2018-01-25
Fix: CSS FOUC issues

blake-newman:
## Changes
- Browser like firefox currently render <style> tags after js. Causing
brief FOUC.
- In order to prevent this you can add a `<script> </script>` to ensure
render is blocked until all css has been loaded.

### before
fouc-before
### after
fouc-after


yyx990803:
I'm not sure if this should be injected by default, since it looks like a bug in Firefox. Also looks like it can be done in userland for temporary workaround.

Issue #3934, opened 2016-10-13
Trim whitespaces from texts in vue-template-compiler output


SebastianS90:
I love the feature of single file components. They allow for proper syntax highlighting and we can write HTML without putting everything in awkward javascript strings.
A proper indentation within the template is useful to improve readability of `.vue` files.

Consider this example:
````html
<template>
  <div>
    <p>
      This is my first text
    </p>
    <p>
      This is my second text
    </p>
  </div>
</template>
````

Compile it with webpack. The resulting bundle contains a compiled version of the template, you will find the following line:
```javascript
return _h('div', [_h('p', ["\n    This is my first text\n  "]), " ", _h('p', ["\n    This is my second text\n  "])])
```

There is this `preserveWhitespace` option described [here](https://github.com/vuejs/vue/blob/d362d64633f00f8cbf1613c424d8852fb45c7596/packages/vue-template-compiler/README.md#options). If you configure webpack with `loader: 'vue?preserveWhitespace=false'`, then you get the following:

```javascript
return _h('div', [_h('p', ["\n    This is my first text\n  "]), _h('p', ["\n    This is my second text\n  "])])
```

Note that one array entry `" "` has gone, i.e. the whitespace between `</p>` and `<p>` has been removed.
But how about the leading and trailing whitespaces in the string `"\n    This is my first text\n  "` ?

### My Feature Rrequest
Allow to trim these texts. The compilation result should look like this:
```javascript
return _h('div', [_h('p', ["This is my first text"]), _h('p', ["This is my second text"])])
```
Changing [this line](https://github.com/vuejs/vue/blob/d362d64633f00f8cbf1613c424d8852fb45c7596/src/compiler/parser/index.js#L213) as follows achieves what I want:
```diff
       text = inPre || text.trim()
-        ? decodeHTMLCached(text)
+        ? decodeHTMLCached(inPre || preserveWhitespace ? text : text.trim())
         // only preserve whitespace if its not right after a starting tag
         : preserveWhitespace && currentParent.children.length ? ' ' : '';
```

But I do not know whether it might break more complex layouts, so I posted it as issue here for discussion.

fnlctrl:
https://github.com/vuejs/vue/releases/tag/v2.0.3

I guess it's already added in 2.0.3?

> vue-template-compiler
> Exposed preserveWhitespace option.

SebastianS90:
@fnlctrl please re-read my entry post, the option is mentioned there :wink:

[...]

PR #7598, opened 2018-02-03
feat: implement template compiler options 'trimTextWhitespace' and 'collapseTextWhitespace' (#3934)

cerlestes:
Implements #3934 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The added options have to be made accessible to at least vue-loader in order to enable them for our builds. I'll see if I can implement that in another PR.

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -7,6 +7,8 @@ declare type CompilerOptions = {
   canBeLeftOpenTag?: (tag: string) => ?boolean; // check if a tag can be left opened
   isReservedTag?: (tag: string) => ?boolean; // check if a tag is a native for the platform
   preserveWhitespace?: boolean; // preserve whitespace between elements?
+  trimTextWhitespace?: boolean; // trim whitespaces from text nodes?
+  collapseTextWhitespace?: boolean; // collapse multiple whitespaces into one within text nodes?
   optimize?: boolean; // optimize static content?
 
   // web specific
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -78,12 +78,15 @@ export function parse (
   delimiters = options.delimiters
 
   const stack = []
-  const preserveWhitespace = options.preserveWhitespace !== false
   let root
   let currentParent
   let inVPre = false
   let inPre = false
   let warned = false
+
+  const preserveWhitespace = options.preserveWhitespace !== false
+  const trimTextWhitespace = options.trimTextWhitespace === true
+  const collapseTextWhitespace = options.collapseTextWhitespace === true
 
   function warnOnce (msg) {
     if (!warned) {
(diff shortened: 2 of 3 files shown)


cerlestes:
After turning on those options, my production build sizes have decreased by 4-10% without any negative side effects at all.

kennardconsulting:
Excited to see this! Great work!

econic:
Really looking forward to it!

Just one wish:
I think this should not remove whitespaces, but replace them with a single space.
Like this you don't have issues where the browser renders a whitespace.
E.g. `text node 1 .. inline node .. text node 2` (all as siblings)
Here the browser will render text node 1's trailing whitespace and text node 2's leading whitespace.
If you trim each whitespace to a single space, the behaviour is not changed at all.

cerlestes:
@econic Thanks for the input. You're totally right that trimming would be fatal in that case. If you only enable the collapse setting and not the trim one though, it should work out fine, as the whitespaces around the text nodes will be collapsed down to a single ASCII-space and not be completely removed.

If this is ever accepted, it'd probably be a safe bet to enable the collapse option by default, as this really shouldn't change behaviour, but reduce component size a lot. The trim option might break existing code and should stay off for now.

revolter:
Any news on this one?

yyx990803:
Thanks for the PR and sorry for letting it hang! See https://github.com/vuejs/vue/issues/9208#issuecomment-450012518 for something that achieves similar result with slight differences. We do want to keep the number of options to a minimum and avoid having too many possible different combinations of behavior. The new condense mode essentially combines collapse with a more conservative trim.

afwn90cj93201nixr2e1re:
@yyx990803 its not same, coz it's doesnt remove tabulating, we dont need tabulating in compiled templates for example in frameworks.
изображение

gold:
This overly complicated implementation seems to be based on a slight misunderstand of the default behavior of HTML rendering. Except for <pre> tag rendering, 2+ space characters are automatically collapsed to a single space in display. All developers know this behavior and thus format their HTML tags accordingly, e.g.;

```html
    <span>one</span>
    <span>two</span>
    <span>three</span>
```

knowning that it will render: one two three

One would not expect it to render: onetwothree

I wrote an HTML minifier in python many years ago; and it takes into account that some spacing between tags are significant and thus should never be completely removed. It's not perfect -- but it also does not cause rendering errors.

```python
# release build strips out HTML comments and removes insignificant space
removeHtmlComments_rx = re.compile(r'<!--.+?-->', re.MULTILINE | re.DOTALL)
removeBlankLines_rx = re.compile(r'^\s*$\n', re.MULTILINE)
removeIndents_rx = re.compile(r'^\s+', re.MULTILINE)
```
The regexen are applied in these three sequential steps against HTML content. Done. Simplicity is a virtue.

Could this be improved? Yes, by replacing newlines with a single space in a fourth step perhaps.

(I recall something I recently read: "Perfect is sometimes the enemy of good.")

cerlestes:
@gold Your proposed implementation is exactly identical to the originally proposed "overly complicated" implementation. Even the regular expressions are basically identical.

PR #7702, opened 2018-02-23
shorten font property

jmilanovic:
Shortens the font styling from font-size, font-weight, font-family to font.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/examples/elastic-header/style.css
+++ b/examples/elastic-header/style.css
@@ -1,6 +1,5 @@
 h1 {
-  font-weight: 300;
-  font-size: 1.8em;
+  font: 300 1.8em Arial, sans-serif;
   margin-top: 0;
 }
 a {
@@ -14,10 +13,8 @@ a {
   overflow: hidden;
   margin: 30px auto;
   position: relative;
-  font-family: 'Roboto', Helvetica, Arial, sans-serif;
+  font: 300 14px 'Roboto', Helvetica, Arial, sans-serif;
   color: #fff;
-  font-size: 14px;
-  font-weight: 300;
   -webkit-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;


examples/elastic-header/style.css
```diff
@@ -1,6 +1,5 @@
 h1 {
-  font-weight: 300;
-  font-size: 1.8em;
+  font: 300 1.8em Arial, sans-serif;
```
Justineo: What's wrong with the original declarations? There's no need to define `font-family` here.

jmilanovic: In an actual live implementation of the code we can assume that a font-familly would defined. The current code omits this and leaves the font-family of the h1 tag to be determined by the browser.

e.g.
font-weight: 300;
font-size: 1.8em;
font-family: x;

Thus, we can shorten future logic into:
font: 300 1.8em x;

myst729: This is an example, and its key point is to demonstrate how to create things with Vue. In my opinion this PR is off-topic whatsoever, because it doesn't have much to do with that purpose.

myst729:
This is an example, and its key point is to demonstrate how to create things with Vue. In my opinion this PR is off-topic whatsoever, because it doesn't have much to do with that purpose.

jmilanovic:
All syntax examples should serve as a model for best code practice.

examples/elastic-header/style.css
Justineo: Omitting a property can be intentional. That’s why we have longhand properties.

Justineo:
IMO always using shorthand properties is hardly best practice. This makes you set unnecessary property values, in this case, `font-family`.

yyx990803:
Thanks, we in general do not modify examples unless for behavioral changes.

PR #7735, opened 2018-03-02
fix(v-model): v-model ended with spaces

### LINKED ISSUE #7730: Problem when a v-model end with spaces
### Version
2.5.13

### Reproduction link
[https://codepen.io/leocardoso/pen/OQqNYz?editors=1011](https://codepen.io/leocardoso/pen/OQqNYz?editors=1011)

### Steps to reproduce
Create an input with a v-model like  <input v-model="obj.value " />

And try to change its value

### What is expected?
The value of model change

The value of model doesn't change and it creates a new property to the Object ended with space like obj['value ']

Leocardoso94:
fix #7730

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


src/compiler/directives/model.js
```diff
@@ -69,6 +69,7 @@ type ModelParseResult = {
 
 export function parseModel (val: string): ModelParseResult {
   len = val.length
+  val = val.trim()
```
posva: Shouldn't this be above the previous line 🤔

Leocardoso94: It's true, I fixed this in the last commit

test/unit/features/directives/model-text.spec.js
```diff
@@ -19,6 +19,17 @@ describe('Directive v-model text', () => {
     }).then(done)
   })
 
+  it('should not create a new property ended with spaces when', () => {
```
posva: Could you add a test that fails without the fix above please? This one passes
Also, could you rename the test title to `should work with extra whitespace properties` or something similar, please?

Leocardoso94: I did not understand the part of the test. Do I create another pull request with only the test? Because the current test fails without this fix

posva:
The current test does pass without the patch you added but it should fail. I'm asking if you could add a test case that fails without the `val = val.trim()` but passes with it

Leocardoso94:
Hi @posva , I add a e2e that fails without the `val.trim()`

Apparently, in the unit test, it was not possible to simulate this case, I believe it only creates the new property with extra spaces only when it is typed in the browser.

Q. The extra property is only created if you pass a v-model from an object like  `v-model="obj.value "`

posva:
You needed to add a test with a nested object 😄 
I added the test at #7737 and kept your initial commits

PR #7755, opened 2018-03-06
Style: update variable name

Gcaufy:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

```
return createWatcher(vm, expOrFn, cb, options)
```

They are actually the same thing, but use different variable name.

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -292,7 +292,7 @@ function initWatch (vm: Component, watch: Object) {
 
 function createWatcher (
   vm: Component,
-  keyOrFn: string | Function,
+  expOrFn: string | Function,
   handler: any,
   options?: Object
 ) {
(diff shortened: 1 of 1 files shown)


src/core/instance/state.js
```diff
@@ -292,7 +292,7 @@ function initWatch (vm: Component, watch: Object) {
 
 function createWatcher (
   vm: Component,
-  keyOrFn: string | Function,
+  expOrFn: string | Function,
```
jkzing: It's key path or function, IMO `keyOrFn` is better :)

Gcaufy: but it's using 
```
return createWatcher(vm, expOrFn, cb, options)
```
should we change `expOrFn` to `keyOrFn`, I just get confused when reading the code, exactly the same thing but using different name.

jkzing: Oops...you are right, it is `expOrFn` indeed.😂

PR #7765, opened 2018-03-07
feat(core): support passing down scopedSlots with v-bind

### LINKED ISSUE #7178: Ability to pass scoped slots to descedants in templates
The related issue (https://github.com/vuejs/vue/issues/5151) is closed, so I need to clarify my point.

It should exist convenient way to define template for items of hierarchical component (for example, tree or list with sections).
I know, it can be solved with render functions, but it is not so "native" way of building components in vue as in react. I mean, writing / reading render functions is more uncomfortable for average vue user, because they are used only occasionally. It's even worse, if user need to setup webpack / babel / jsx. So I think, render functions should be avoided for not-so-advanced cases.

Slots can be "passed" to children using templates, like this:
```
<child>
    <template slot-scope="props">
        <slot v-bind="props"></slot>
    </template>
</child>
```
But there are some drawbacks:
1) This creates wrapper function on each level. First, we have wrapper. Then wrapper of wrapper, and so on.
2) This breaks fallback content (default template) feature.

```
<child>
    <slot slot-scope></slot>
</child>
```
### Remark
I propose syntax, which is boiled down version of first markup, and can be supported by compiler. Here is my test implementation (it seems quite easy): https://github.com/shameleo/vue-slot-pass-demo/commit/bdb6cbbf3ab7574d4b2d350f7d866e2bc609a043
If you give me green light, I'll try to prepare PR. If something is missed in implementation, let me know, please.

yyx990803:
Allow passing down scoped slots using `v-bind`:

``` vue
<child v-bind="{ scopedSlots: $scopedSlots }">
</child>
```

close #7178, close #7351

### DIFF
--- a/src/core/instance/render-helpers/bind-object-props.js
+++ b/src/core/instance/render-helpers/bind-object-props.js
@@ -34,6 +34,7 @@ export function bindObjectProps (
         if (
           key === 'class' ||
           key === 'style' ||
+          key === 'scopedSlots' ||
           isReservedAttribute(key)
         ) {
           hash = data
--- a/test/unit/features/component/component-scoped-slot.spec.js
+++ b/test/unit/features/component/component-scoped-slot.spec.js
@@ -593,4 +593,36 @@ describe('Component scoped slot', () => {
       expect(vm.$el.innerHTML).toBe('<p>hello</p>')
     }).then(done)
   })
+
+  it('passing down scoped slots with v-bind', () => {
+    const Bar = {
+      template: `
+        <div class="bar">
+          <slot name="hello" msg="hi" />
+        </div>
+      `
+    }
+
+    const Foo = {
+      components: { Bar },
+      template: `
+        <div class="foo">
+          <bar v-bind="{ scopedSlots: $scopedSlots }"></bar>
+        </div>
+      `
+    }
+
+    const vm = new Vue({
+      components: { Foo },
+      template: `
+        <foo>
+          <template slot="hello" slot-scope="props">
+            {{ props.msg + '!' }}
+          </template>
+        </foo>
+      `
+    }).$mount()
+
+    expect(vm.$el.querySelector('.bar').textContent.trim()).toBe('hi!')
+  })
 })
--- a/test/unit/features/component/component-scoped-slot.spec.js
+++ b/test/unit/features/component/component-scoped-slot.spec.js
@@ -625,4 +625,75 @@ describe('Component scoped slot', () => {
 
     expect(vm.$el.querySelector('.bar').textContent.trim()).toBe('hi!')
   })
+
+  it('passing down merged scoped slots with v-bind', () => {
+    const Bar = {
+      template: `
+        <div class="bar">
+          <slot name="one" msg="1" />
+          <slot name="two" msg="2" />
+        </div>
+      `
+    }
+
+    const Ney = {
+      template: `
+        <div class="ney">
+          <slot name="three" msg="3">
+            3?
+          </slot>
+        </div>
+      `
+    }
+
+    const Foo = {
+      components: { Bar, Ney },
+      template: `
+        <div class="foo">
+          <bar v-bind="{ scopedSlots: $scopedSlots }"></bar>
+          <ney v-bind="{ scopedSlots: $scopedSlots }"></ney>
+        </div>
+      `
+    }
+
+    const vm1 = new Vue({
+      components: { Foo },
+      template: `
+        <foo>
+          <template slot="one" slot-scope="props">
+            {{ props.msg + '!' }}
+          </template>
+          <template slot="two" slot-scope="props">
+            {{ props.msg + '!!' }}
+          </template>
+        </foo>
+      `
+    }).$mount()
+
+    expect(vm1.$el.textContent.trim().replace(/\s+/g, '')).toBe('1!2!!3?')
+
+    const vm2 = new Vue({
+      components: { Foo },
+      methods: {
+        tickle (me) {
+          return me + ' tickled!'
+        }
+      },
+      template: `
+        <foo>
+          <template slot="one" slot-scope="props">
+            {{ props.msg + '!' }}
+          </template>
+          <template slot="two" slot-scope="props">
+            {{ props.msg + '!!' }}
+          </template>
+          <template slot="three" slot-scope="props">
+            {{ tickle( props.msg )}}
+          </template>
+        </foo>
+      `
+    }).$mount()
+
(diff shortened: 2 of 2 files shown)


yyx990803:
Btw, this is still open to discussion, since the syntax isn't entirely consistent.

You can do `v-bind="{ ref: 'foo', slot: 'bar' }"` which makes sense because `ref` and `slot` can be bound individually as attributes, but `scopedSlots` is not.

Internally `ref` and `slot` are treated specially because they are "reserved" attributes and thus merged into the VNode's data as a root-level property.

So alternatively, we can expose a special convention for `v-bind`: if a property starts with `$`, it will always be merged as a root-level property into the VNode's data.

This then allows the usage:

``` vue
<child v-bind="{ $scopedSlots } />
```

In addition, if you know the underlying [VNode data interface](https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth), you can also do low-level stuff without dropping into render functions:

``` vue
<child v-bind="{
  $on: {
    someEvent: someMethod
  },
  $nativeOn: {
    click: someMethod
  }
}"/>
```

ktsn:
What about adding a dedicated directive for merging vnode data? Since `v-bind` is used to be used for passing parent component data to child component attribute/props, it is a bit weird if it can modify underlying vnode data, IMHO.

If we have such directive we can simply write the vnode data as the same shape as in render function (named it `v-vnode` as a placeholder 😉 ):

```vue
<child v-vnode="{
  scopedSlots: $scopedSlots,
  on: {
    someEvent: someMethod
  },
  nativeOn: {
    click: someMethod
  }
}"/>
```

We may write directive argument like `v-bind`:

```vue
<child v-vnode:scoped-slots="$scopedSlots" />
```

This also has a benefit that we may able to write template type checker for that as we can simply utilize Vue's typescript typings. (FYI: The current WIP implementation of template type checker is [here](https://github.com/vuejs/vetur/pull/681))

dennythecoder:
The specific directive proposed by @ktsn seems a little awkward to me, but I think it would make sense to separate the behavior to improve readability.

FrankFang:
```
<child v-bind="{ scopedSlots: $scopedSlots }">
</child>
```

Dost not work for me :(

bigianb:
What's the status of this?

kellym:
So is there a consensus on using a different tag name or not? If so, what should it be called? 

Otherwise, this is clearly a hangup for a lot of people using Vue (googling led me to 5-6 different issues on the same topic) and it'd be nice to either give this a thumbs up or have some direction in what needs to be done to get this shipped.

ClickerMonkey:
I wouldn't recommend this solution, but I needed this functionality now - so I implemented @yyx990803 's change as a plugin:

```javascript
var BindScopedSlotsPlugin = {
  install: function(Vue, options) {
    Vue.prototype._b = function bindObjectProps (
      data,
      tag,
      value,
      asProp,
      isSync
    ) {
      if (value) {
        if (value === null || typeof value !== 'object') {
          warn(
            'v-bind without argument expects an Object or Array value',
            this
          );
        } else {
          if (Array.isArray(value)) {
            var res = {};
            for (var i = 0; i < value.length; i++) {
              if (arr[i]) {
                Vue.util.extend(res, value[i]);
              }
            }
            value = res;
          }
          var hash;
          var loop = function ( key ) {
            if (
              key === 'class' ||
              key === 'style' ||
              key === 'scopedSlots' || // new code
              Vue.config.isReservedAttr(key)
            ) {
              hash = data;
            } else {
              var type = data.attrs && data.attrs.type;
              hash = asProp || Vue.config.mustUseProp(tag, type, key)
                ? data.domProps || (data.domProps = {})
                : data.attrs || (data.attrs = {});
            }
            if (!(key in hash)) {
              hash[key] = value[key];

              if (isSync) {
                var on = data.on || (data.on = {});
                on[("update:" + key)] = function ($event) {
                  value[key] = $event;
                };
              }
            }
          };

          for (var key in value) loop( key );
        }
      }
      return data;
    };
  }
};

Vue.use( BindScopedSlotsPlugin );
```

The syntax is as mentioned initially: `v-bind="{ scopedSlots: $scopedSlots }"`

joernroeder:
@ClickerMonkey thanks for building the plugin! there is a tiny typo: `if (arr[i])` needs to be replaced with `if (value[i])`…

ClickerMonkey:
Thanks! I have a smaller simpler version:

```javascript
    Vue.prototype._b = (function (bind) {
      return function(data, tag, value, asProp, isSync) {
        if (value && value.$scopedSlots) {
          data.scopedSlots = value.$scopedSlots;
          delete value.$scopedSlots;
        }
        return bind.apply(this, arguments);
      };
    })(Vue.prototype._b);
```

And you just need to add `v-bind="{$scopedSlots}"` to the component.

RedShift1:
Check out this solution too: https://stackoverflow.com/a/52823029/2050460

ggirodda:
I edited the @ClickerMonkey script to merge the parent and the current $scopedSlots, maybe it can be useful for someone
```
import Vue from "vue"

Vue.prototype._b = (function(bind) {
  return function(data, tag, value, asProp, isSync) {
    if (value && value.$scopedSlots) {
      data.scopedSlots = {
        ...data.scopedSlots,
        ...value.$scopedSlots
      }
      delete value.$scopedSlots
    }
    return bind.apply(this, arguments)
  }
})(Vue.prototype._b)
```

yyx990803:
So I don't think we will land this, because:

1. I'm not entirely satisfied with the consistency of the change. Feels like a dirty special case.

2. There is [already a way to make it work in userland](https://github.com/vuejs/vue/pull/7765#issuecomment-401985464).

3. This is probably the most important reason: we want to avoid landing changes in v2 that would then have to be broken in v3. In v3, slots and scoped slots are being unified under `$slots`, and `slots` itself is becoming a reserved prop like `key` and `ref`. This means in v3 you can just do:

``` html
<foo :slots="$slots"></foo>
```

Moreover, we have other changes in v3 that makes passing things down to child components easier (RFC in early 2019).

ClickerMonkey:
Glad to hear about v3. Thanks Evan for everything!

shameleo:
> 3\. In v3, slots and scoped slots are being unified under `$slots`, and `slots` itself is becoming a reserved prop like `key` and `ref`. This means in v3 you can just do:
> 
> ```
> <foo :slots="$slots"></foo>
> ```

So, there is no reserved `slot` prop in Vue 3. Any new ways to pass down scoped slots are available? Or planned? Or we still need old workarounds like `<template>` / `<slot>` pair?

PR #7819, opened 2018-03-13
Static type="checkbox" with v-bind (fix #7811)

### LINKED ISSUE #7811: v-bind=$attrs overrides the input type attribute in 2.5.14 and above
### Version
2.5.14

### Reproduction link
[https://jsfiddle.net/linnea/onw3Lfm4/2/](https://jsfiddle.net/linnea/onw3Lfm4/2/)

### Steps to reproduce
Create a custom component with an input wrapped in a div and:

1. inherit attributes 
`v-bind="$attrs"`
2. Set the type attribute on the input
`type="checkbox"` 
3. Have a v-model on that input.

### What is expected?
The inner input still receives type="checkbox"

No type attribute is inherited

---
We have a custom checkbox component that provides label styling etc, and used to be able to declare the input type as well as inherit attributes from the component. Other attributes that are set on the inner input are passed correctly (ie disabled, :true-value, etc), but not the type attribute, unless you bind it. This was working prior to v2.5.14.

A temporary fix we've implemented is changing `type="checkbox"` to `:type="'checkbox'"`, or setting v-bind="$attrs" to a computed property that returns an object with Object.assign({ type: 'checkbox'}, this.$attrs);

Removing the v-model from the input also fixes it, but we use it to handle multiple checkboxes, bound to the same array

Here's the same fiddle above, working in 2.5.13
https://jsfiddle.net/linnea/onw3Lfm4/4/

hirokiosame:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


src/platforms/web/compiler/modules/model.js
```diff
@@ -34,7 +34,7 @@ function preTransformNode (el: ASTElement, options: CompilerOptions) {
     if (map[':type'] || map['v-bind:type']) {
       typeBinding = getBindingAttr(el, 'type')
     }
-    if (!typeBinding && map['v-bind']) {
+    if (!map.type && !typeBinding && map['v-bind']) {
```
jkzing: This would make following case lose type binding: 
``` vue
<input type="checkbox" v-model="test" v-bind="{type: typeBindingValue}">
```

So maybe
```js
typeBinding = `(${map['v-bind']}).type || '${map.type}'`
```

posva: @hirokiosame If you could add the test as well, that would be great 👌

yyx990803: I don't think we should allow `v-bind` to override a static `type`, it's gonna be confusing if done accidentally and I don't think there's a valid use case.

PR #7826, opened 2018-03-13
Clean shell scripts

zapashcanon:
Hi,

I fixed various things spotted by shellcheck.

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/release-weex.sh
+++ b/scripts/release-weex.sh
@@ -1,6 +1,7 @@
+#!/bin/bash
 set -e
-CUR_VERSION=`node build/get-weex-version.js -c`
-NEXT_VERSION=`node build/get-weex-version.js`
+CUR_VERSION=$(node build/get-weex-version.js -c)
+NEXT_VERSION=$(node build/get-weex-version.js)
 
 echo "Current: $CUR_VERSION"
 read -p "Enter new version ($NEXT_VERSION): " -n 1 -r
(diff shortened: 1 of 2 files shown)


scripts/release-weex.sh
```diff
@@ -20,15 +21,15 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
-  cd -
+  )
 
-  cd packages/weex-template-compiler
-  npm version $NEXT_VERSION
+  ( cd packages/weex-template-compiler
+  npm version "$NEXT_VERSION"
   npm publish
-  cd -
+  )
```
elamperti: This is a nice trick! Though I'm not sure if it'd be clear for everyone what is it doing. I'd stick with the `cd -` just for clarity.

scripts/release.sh
```diff
@@ -32,23 +33,23 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
-  npm version $VERSION
+  ( cd packages/vue-template-compiler
+  npm version "$VERSION"
   if [[ -z $RELEASE_TAG ]]; then
     npm publish
   else
-    npm publish --tag $RELEASE_TAG
+    npm publish --tag "$RELEASE_TAG"
   fi
   cd -
```
elamperti: In any case, for consistency, shouldn't this `cd -` be replaced as well?

zapashcanon: Indeed. Will do. I'm not sure if shellcheck would have spotted it after running it a second time. Will try. If not, we may found a bug....

scripts/release-weex.sh
zapashcanon: I can add a comment like: `using a subshell to avoid having to cd back`.

It's not just to avoid writing a line, it also can prevent some troubles, see [SC2013](https://github.com/koalaman/shellcheck/wiki/SC2103).

elamperti: Oh, I didn't know about that one! It should be modified then. The comment would be helpful :)

scripts/release.sh
zapashcanon: Opened [an issue](https://github.com/koalaman/shellcheck/issues/1146). :)

scripts/release-weex.sh
zapashcanon: Done.

PR #7843, opened 2018-03-16
feat(weex): invoke handlers on virtual component instance

Hanks10100:

- [x] Feature

- [x] No

**Other information:**

> **These modifications only take effects on Weex platform.**

This is a partial feature of recycle-list.

Because of there is no virtual dom in virtual components, the event handler can only be bind on the virtual component template, but the event handler should be executed in the virtual component context (refer to this [brief explanation](https://segmentfault.com/a/1190000013697211#articleHeader13)). So at least two processes should be changed to support this feature.

**1. Using dynamic `this` for the event handler**

When compiling the `v-on` directive, the expression will be compiled into a handler function. Along with the `with(this){}` magic, the handler function's context (`this`) always be same as the render function. But it should be dynamic for virtual components, the real context of the event handler could be re-assigned at runtime.

To achieve this, I added an explicit `this` to the event handlers which are used in recyclable components.  ([platforms/weex/compiler/modules/event.js#L45-L60](https://github.com/Hanks10100/vue/blob/6be2fc83130c1aa1cf696800a2b7fb23eec9f6e1/src/platforms/weex/compiler/modules/event.js#L45-L60)) 

**2. Finding the expected context when invoking the event handler.**

At runtime, if an event is dispatched to a virtual component template, then find the real context by `componentId` and invoke the handler in the expected context. ([platforms/weex/runtime/modules/events.js#L48-L65](https://github.com/Hanks10100/vue/blob/6be2fc83130c1aa1cf696800a2b7fb23eec9f6e1/src/platforms/weex/runtime/modules/events.js#L48-L65))

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,8 +136,9 @@
     "shelljs": "^0.8.1",
     "typescript": "^2.7.1",
     "uglify-js": "^3.0.15",
+    "vue-template-es2015-compiler": "^1.6.0",
     "webpack": "^3.11.0",
-    "weex-js-runtime": "^0.23.6",
+    "weex-js-runtime": "^0.23.7",
     "weex-styler": "^0.3.0",
     "yorkie": "^1.0.1"
   },
--- a/src/compiler/codegen/events.js
+++ b/src/compiler/codegen/events.js
@@ -1,7 +1,9 @@
 /* @flow */
 
-const fnExpRE = /^([\w$_]+|\([^)]*?\))\s*=>|^function\s*\(/
-const simplePathRE = /^[A-Za-z_$][\w$]*(?:\.[A-Za-z_$][\w$]*|\['[^']*?']|\["[^"]*?"]|\[\d+]|\[[A-Za-z_$][\w$]*])*$/
+import { genWeexHandlerWithParams, genWeexHandler } from 'weex/compiler/modules/event'
+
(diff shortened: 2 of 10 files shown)


Hanks10100:
Why this PR only run the ci script of `test:weex`? Actually it'll lead a failure to `test:cover` and `test:e2e`. I have tried on my local machine, but I am not sure about how to solve it.

I guess the reason is the `require` which are used in the weex compiler.

If I [comment out those "requires"](https://github.com/Hanks10100/vue/compare/weex-event-handler...Hanks10100:07bb0545a64fc5291f7797e09f0623babd6b8409), of course the `test:weex` and type check will be failed, but [the `test:e2e` and `test:cover` will be passed](https://github.com/Hanks10100/vue/pull/6). So I think these modified codes won't cause a failure to the test cases, but the test driver (PhantomJS or Karma) can't work well with the CommonJS module syntax (`require`). I thought those codes would be *shaking off* while compiling, because of those packages only be used in the weex compiler.

@yyx990803 Do you have any idea to solve the ci failures?

linghengqian (commented):
You need to resolve merge conflicts.

Hanks10100:
This PR is outdated for a long time. I'll close it.

PR #7857, opened 2018-03-19
feat(compiler): add warning when v-show used on template tag

superMDguy:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
According to the docs, the `v-show` directive [doesn't support](https://vuejs.org/v2/guide/conditional.html#v-show)  the `<template>` element. So, if a `v-show` is used on a template, it is essentially ignored. However, it currently does that without any warning. This pull request adds a warning when a `v-show` directive is used on a template tag.

### DIFF
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -59,6 +59,13 @@ export function genElement (el: ASTElement, state: CodegenState): string {
   } else if (el.if && !el.ifProcessed) {
     return genIf(el, state)
   } else if (el.tag === 'template' && !el.slotTarget) {
+    if (process.env.NODE_ENV !== 'production' &&
+        el.directives && el.directives.length > 0 &&
+        el.directives.find(directive => directive.name === 'show')) {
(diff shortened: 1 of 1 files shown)


yyx990803:
Hey, thanks for the PR! A few things to note:

1. Actually, the only directives that make sense on `<template>` are `v-if` and `v-for`. So we can check and warn not only `v-show`, but any directives that are not `v-if` or `v-for`.

2. This may not be the best place to have a big chunk of logic since it kinda breaks the readability of the function which is mostly branches. We should extract it into a separate function, something like:

``` js
if (process.env.NODE_ENV !== 'production') {
  checkInvalidTemplateDirs(el)
}
```

superMDguy:
I've made the changes. I was thinking other directives also wouldn't be allowed, but I wasn't sure if that was the case 100% of the time.

felixbuenemann:
Shouldn't `v-slot` also be allowed?

PR #7861, opened 2018-03-20
Adds missing `asyncMeta` during VNode cloning

luisatmaniak:
- [x] Bugfix

- [x] No

**Other information:**
This fixes a crash during SSR, where a parent functional component is using an async component in the render function. If the prop is not copied, `renderAsyncComponent` crashes during rendering. Looks like this bug was introduced in 62a922e865f5e578f67b386cb614abfc173d7851


yyx990803:
Thanks for the fix, is it possible to provide a reproduction for the case so we can better determine if this fix is sufficient?

yyx990803:
I merged it because `asyncMeta` indeed should be copied, but if possible, I'd still love to see a reproduction of the case so that we can be sure this fix covers the bug properly. Thanks.

PR #7926, opened 2018-03-29
fix(provide): Merges symbol provides

### LINKED ISSUE #7923: Mixin provide with Symbols don't get merged properly
### Version
2.5.16

### Reproduction link
[https://jsfiddle.net/hirokiosame/38nzxq6c/26/](https://jsfiddle.net/hirokiosame/38nzxq6c/26/)

### Steps to reproduce
Run the code. The injected result will show `[undefined, "bar"]` in the console.

### What is expected?
For the merging to work properly. The expected result is `['foo', 'bar']`, which is what you'll see when you convert `keyA` and `keyB` into strings.

Merging fails:  `[undefined, "bar"]`

hirokiosame:
Fixes merging multiple provides using Symbols

fix #7923

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The test I added only passes in an environment with symbols (since it's handling a case that fails with symbols). This fails when the test is executed with Phantom, since it doesn't have Symbol support natively or with a polyfill, but passes in Chrome.

I can add babel-polyfill but seemed excessive for one test. Wanted to know how to move forward with that one before making any drastic changes to making it work.

### DIFF
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -51,14 +51,13 @@ function mergeData (to: Object, from: ?Object): Object {
   let key, toVal, fromVal
 
   const keys = hasSymbol
-    ? Reflect.ownKeys(from).filter(key => {
-      /* istanbul ignore next */
-      return Object.getOwnPropertyDescriptor(from, key).enumerable
-    })
+    ? Reflect.ownKeys(from)
     : Object.keys(from)
 
   for (let i = 0; i < keys.length; i++) {
     key = keys[i]
+    // in case the object is already observed...
+    if (key === '__ob__') continue
     toVal = to[key]
     fromVal = from[key]
     if (!hasOwn(to, key)) {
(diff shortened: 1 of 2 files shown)


src/core/util/options.js
```diff
@@ -48,7 +49,14 @@ if (process.env.NODE_ENV !== 'production') {
 function mergeData (to: Object, from: ?Object): Object {
   if (!from) return to
   let key, toVal, fromVal
-  const keys = Object.keys(from)
+
+  const keys = hasSymbol
+    ? Reflect.ownKeys(from).filter(key => {
+      /* istanbul ignore next */
```
hirokiosame: Wasn't entirely sure if this was needed but took it from `inject.js` to be safe.

posva:
Thanks for the contribution and sorry for the delay! I moved the test so it isn't run if `Reflect` isn't supported

yyx990803:
Note: `Reflect.ownKeys` + filter with descriptor check is about 10x slower than `Object.keys()`, I think we need to avoid this cost for normal data merging, and only do this for `provide()` merges.

PR #7929, opened 2018-03-30
fix(utils/looseEqual): update looseEqual to handle dates (close #7928)

### LINKED ISSUE #7928: <select> is broken when options values are bound to Dates
### Version
2.5.16

### Reproduction link
[https://jsfiddle.net/azpxdt3a/1/](https://jsfiddle.net/azpxdt3a/1/)

### Steps to reproduce
Try select any item except first one.

### What is expected?
Item selected.

<select> always shows first item. Once any item except first is selected, it's no more possible to select first one (actually, it's possible, as <select> shows correct selected item after the same item was selected twice in a row)

---
It seems the bug was introduced in ver 2.4.2. Does not reproduced with objects instead of dates: https://jsfiddle.net/pr847ek7/

w3cj:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -282,15 +282,21 @@ export function looseEqual (a: any, b: any): boolean {
         return a.length === b.length && a.every((e, i) => {
           return looseEqual(e, b[i])
         })
-      } else if (!isArrayA && !isArrayB) {
-        const keysA = Object.keys(a)
-        const keysB = Object.keys(b)
-        return keysA.length === keysB.length && keysA.every(key => {
-          return looseEqual(a[key], b[key])
-        })
       } else {
-        /* istanbul ignore next */
-        return false
+        const isDateA = a instanceof Date
+        const isDateB = b instanceof Date
+        if (isDateA && isDateB) {
+          return a.toUTCString() === b.toUTCString()
+        } else if (!isArrayA && !isArrayB && !isDateA && !isDateB) {
+          const keysA = Object.keys(a)
+          const keysB = Object.keys(b)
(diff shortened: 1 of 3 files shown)


src/shared/util.js
```diff
@@ -282,15 +282,21 @@ export function looseEqual (a: any, b: any): boolean {
...
-      } else if (!isArrayA && !isArrayB) {
-        const keysA = Object.keys(a)
-        const keysB = Object.keys(b)
-        return keysA.length === keysB.length && keysA.every(key => {
-          return looseEqual(a[key], b[key])
-        })
       } else {
-        /* istanbul ignore next */
-        return false
+        const isDateA = a instanceof Date
```
posva: hmm, why did you added it this way instead of another `else if` condition?

w3cj: I originally defined the variables at the top below the isArrayA, and isArrayB variables, but that was calling `instanceof` for everything that got passed in and added about 500 milliseconds to the test run. It's my first time perusing the code base. Not sure how much this method is used and if it would have a significant impact on anything else.

Also, I originally called instanceof directly inside the if, but tried to follow the code style of putting them in variables first.

posva:
I created #7940, this is what I meant by adding an else if. Tell me if I missed a scenario you thought about, I kept the test exactly as you added it
Couldn't add your commit, sorry 🙁

w3cj:
Looks good to me, and no worries about the commit. Glad I could help locate the bug!

PR #7948, opened 2018-04-02
refactor(compiler): Document and refactor filter parser

8uurg:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Adding documentation

- [ ] Yes
- [x] No

**Other information:**
Hey, I am a member of a group of students (@LarsStegman, @PhilipeLouchtch, @TimSpeelman) that has been analysing Vue for the course Software Architecture. As part of this course it was suggested to create a contribution to reduce technical debt.
During this analysis we found that the filter parser was hard to understand,
this PR is supposed to remedy this by adding additional documentation.
String (single, double, template quoted) and regex matching have been overhauled to use a special method to seek the ending quote instead, this effectively behaves the same as the code did before.

We ran a benchmark with a weird filter string, the result of which indicated there is no degradation in performance.

### DIFF
--- a/src/compiler/parser/filter-parser.js
+++ b/src/compiler/parser/filter-parser.js
@@ -7,8 +7,7 @@ export function parseFilters (exp: string): string {
   let square = 0
   let paren = 0
   let lastFilterIndex = 0
-  let c, i, filters
-  let expression = ''
+  let c, i, expression, filters
 
   for (i = 0; i < exp.length; i++) {
     c = exp.charCodeAt(i)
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks a lot for this PR and sorry for letting it sit. While I think this helps with readability to some extent, I don't really think it's worth it. The original parsing state is simple enough so the gain from the `seek` function is marginal. Declaring variables for all the constants also bloats the final gzipped size by close to half a kb, which should be avoided (unfortunate, but in future codebases we will definitely leverage TypeScript's const enum for this)

PR #7953, opened 2018-04-03
fix: stop render when failed setAttr

derushio:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Description
* fix: stop render when failed setAttr
    * ex) when render compiled editable jade/pug texts, there are cases that have incorrect attrs.
    * change: stop render -> render & logging Vue warn

### DIFF
--- a/src/platforms/web/runtime/modules/attrs.js
+++ b/src/platforms/web/runtime/modules/attrs.js
@@ -88,27 +88,33 @@ function setAttr (el: Element, key: string, value: any) {
 }
 
 function baseSetAttr (el, key, value) {
-  if (isFalsyAttrValue(value)) {
-    el.removeAttribute(key)
-  } else {
-    // #7138: IE10 & 11 fires input event when setting placeholder on
-    // <textarea>... block the first input event and remove the blocker
-    // immediately.
-    /* istanbul ignore if */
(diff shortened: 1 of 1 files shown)


posva:
Thanks for the PR but adding a try/catch would slow down the code when the problem comes from the code written in a different templating language.
Can you show us what cases is this helping out?
If it is ever added, it must be added in development mode only

narumi18wa:
Thanks for reply.

I wanted to make Pug editor preview.

I didn't know that try-catch makes code slow down.
I think this try-catch doesn't have to be added in dev mode.
Because my probrem occurs in production mode too.

I want to resolve the problem by other approach.
But I can't catch this Error anywhere.
I tried Vue.config.errorHandler and renderError, but Vue.js stops before catching the error.
I want to catch the error in renderError.
Do you have any good ideas?

I'm not sure if this case is an issue for you.
Does the problem need to be reported to  `GitHub issues`?

yyx990803:
Thanks for the PR, but this kind of error should be handled using provided error handlers (e.g. `errorCaptured`)

PR #7974, opened 2018-04-06
Allow for custom rule of blocks parsing

yassipad:
I am currently working on a Nativescript+Vue template, which includes a specific webpack config that allows for differentiated templates based on the target (web or native). Single file components are extremely useful and being able to add a custom rule to the parsing of their blocks would make it possible to write a single file containing a `native` template alongside a `web` template, both of which sharing the same <script> logic or <style> rules. So far, the only workaround is to change the 'vue-loader' in webpack but by adding this it would be possible to use the actual official 'vue-loader' and change only the necessary part.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/vue-template-compiler/build.js
+++ b/packages/vue-template-compiler/build.js
@@ -623,15 +623,20 @@ function parseComponent (
           return cumulated
         }, {})
       };
-      if (isSpecialTag(tag)) {
-        checkAttrs(currentBlock, attrs);
-        if (tag === 'style') {
-          sfc.styles.push(currentBlock);
-        } else {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR, but I think it's a bad idea to expose a method on `module.exports` and expects it to be overwritten. The changes should also be made to `src/sfc/parser` instead of the built file.

I know this is probably outdated effort, but if you still have a similar need, it should probably be done differently.

PR #8019, opened 2018-04-13
fix(slots): preserve reactivity inside of $slots, fix #7856

### LINKED ISSUE #7856: Evaluation of slot content's context is inconsistent with "slot-scope" behavior.
### Version
2.5.16

### Reproduction link
[https://codepen.io/dasdeck/pen/LdbqLL?editors=0001](https://codepen.io/dasdeck/pen/LdbqLL?editors=0001)

### Steps to reproduce
When you open the pen, a message box will get opened because the full context of Consumer1 gets evaluated.
Since the slot is not shown, this evaluation is unnecessary at this point and negates optimizations.
When one declares a slot-scope, like in the Consumer2 components, Vue shows the "correct" behavior.

### What is expected?
The slotted context should evaluate in a lazy manner

The slotted context gets evaluate right away.

fenduru:
Previously $slots was not reactive which caused use cases such as
portal-vue to not update when the contents of the slots updated. This
was because there were no reactive properties accessed when using $slots
and renders were only being triggered by calls to $forceUpdate when slot
contents changed.

This change makes $slots reactive, so updates are detected appropriately
regardless of where the slot is used.

fixes #7856

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

Related to https://github.com/LinusBorg/portal-vue/issues/105

### DIFF
--- a/src/core/instance/render.js
+++ b/src/core/instance/render.js
@@ -21,7 +21,7 @@ export function initRender (vm: Component) {
   const options = vm.$options
   const parentVnode = vm.$vnode = options._parentVnode // the placeholder node in parent tree
   const renderContext = parentVnode && parentVnode.context
-  vm.$slots = resolveSlots(options._renderChildren, renderContext)
+  defineReactive(vm, '$slots', resolveSlots(options._renderChildren, renderContext), null, true)
   vm.$scopedSlots = emptyObject
   // bind the createElement fn to this instance
   // so that we get proper render context inside it.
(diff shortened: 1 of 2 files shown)


fenduru:
@yyx990803 can this be merged?

PxyUp:
For me create reactive $slots property, it is not solution issue(in that case of course it is one of the solution), but slots it is not like reactive variable, with watches and observer, it is another type.

yyx990803:
This seems to be unrelated to the original issue discussed in #7856. Also, the added test case is not really a use case I think we should be encouraging. I'd rather implement proper portal support in core instead of relying on such usage.

Issue #8017, opened 2018-04-13
2.5.14 introduces TypeError bind is not a function


rjoo:
### Version
2.5.14

### Reproduction link
[https://jsfiddle.net/bad4a8mb/1/](https://jsfiddle.net/bad4a8mb/1/)

### Steps to reproduce
Provide a non function to the `bind` utility method.

### What is expected?
In <= 2.5.13, it rendered but failed silently: https://jsfiddle.net/ujxu9mvd/. Not sure if we'd want to keep that expectation, but existing projects will fail to render completely. A dev-mode warning at the very least will be helpful in identifying the issue.

### What is actually happening?
Uncaught TypeError

---
The refactoring of the `bind` method to use the native bind, exposes a coding error when upgrading from 2.5.13. We have a fairly large codebase and one of the developers mistakenly added an object as a component method (eg. a global Vue events instance). It made its way into production and worked fine until upgrading to 2.5.14. It took a while to find it as there was no clear indication as to where the error was happening.

Some other devs experienced this as well:
https://github.com/vuejs/vue/pull/7491
https://github.com/vuejs/vue/commit/324672a4621a3123cac90b18d4b1818bb33dddc6

posva:
A warning in dev mode would be fine

thedamon:
This was a weird one to track down! We had a js file being used as a mixin, but one of the exports was an array.

PR #8022, opened 2018-04-13
feat: add warning in dev mode when method is not function (#…

yaodingyd:
#8017
add warning in dev mode when method is not function 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


src/core/instance/state.js
```diff
@@ -273,6 +273,11 @@ function initMethods (vm: Component, methods: Object) {
           `Avoid defining component methods that start with _ or $.`
         )
       }
+      if (typeof methods[key] !== 'function') {
```
posva: Could you make this test in development mode only and add a test (or/and adapt existing one), please?

src/core/instance/state.js
```diff
@@ -273,6 +273,12 @@ function initMethods (vm: Component, methods: Object) {
           `Avoid defining component methods that start with _ or $.`
         )
       }
+      if (methods[key] != null && typeof methods[key] !== 'function') {
+        warn(
+          `Method "${key}" should be a function.`
+        )
+        break
```
posva: you could instead set the method[key] to null, so the code works but does nothing

src/core/instance/state.js
```diff
@@ -273,6 +273,12 @@ function initMethods (vm: Component, methods: Object) {
           `Avoid defining component methods that start with _ or $.`
         )
       }
+      if (methods[key] != null && typeof methods[key] !== 'function') {
```
posva: instead of checking if it's null, move it up with an `else if`

sirlancelot:
A check and a warning already exists for this in the block directly above the one added by this PR.

yaodingyd:
that's sad.

Issue #8015, opened 2018-04-13
Merge common class names of component and root element of that component


jimut:
### What problem does this feature solve?
It looks a little awkward if there are duplicate class names in a DOM element.

### What does the proposed API look like?
If you look into [this fiddle](https://jsfiddle.net/zLnzuh1f/). 

The `div` element of `foo` component will have two `bar` classes in it. So it looks like.

```html
<div class="bar bar"></div>
```

If the two classes would have been merged into one.

```html
<div class="bar"></div>
```

It would have worked and behaved exactly in the same way but the code will look a lot cleaner.

astronomersiva:
I am working on this!

Mushr0000m:
I have the same issue. I need to keep the class in the component because it make sense for the component to know it's own class name for style etc..

But also need to keep the class attribute on the component call because it's a wrapper component around some server side generated content (header and footer), so without this class I have a glitch while Vuejs mount the component with the class name and so the browser doesn't apply the css.

I tried to look in `beforeMount` or `mounted`to remove duplicates but can't find a way. I know it works anyway, but the double class name looks weird, so if you know a solution to remove this, it would be nice.

Vansh0204:
@jimut can u please assign me this issue.

PR #8035, opened 2018-04-16
refactor: merge common class names of a component(re #8015)

astronomersiva:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/platforms/web/util/class.js
+++ b/src/platforms/web/util/class.js
@@ -37,7 +37,11 @@ export function renderClass (
   dynamicClass: any
 ): string {
   if (isDef(staticClass) || isDef(dynamicClass)) {
-    return concat(staticClass, stringifyClass(dynamicClass))
+    const concatenatedClassNames = concat(staticClass, stringifyClass(dynamicClass))
+    return concatenatedClassNames
+      .split(' ')
+      .filter((item, index, array) => index === array.indexOf(item))
+      .join(' ')
   }
   /* istanbul ignore next */
   return ''
--- a/test/unit/modules/vdom/modules/class.spec.js
+++ b/test/unit/modules/vdom/modules/class.spec.js
@@ -51,6 +51,15 @@ describe('vdom class module', () => {
     expect(elm).toHaveClass('class2')
   })
 
+  it('should remove duplicate class names', () => {
+    const vnode = new VNode('p', {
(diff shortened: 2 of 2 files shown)


src/platforms/web/util/class.js
```diff
@@ -37,7 +37,8 @@ export function renderClass (
   dynamicClass: any
 ): string {
   if (isDef(staticClass) || isDef(dynamicClass)) {
-    return concat(staticClass, stringifyClass(dynamicClass))
+    const concatenatedClassNames = concat(staticClass, stringifyClass(dynamicClass))
```
lusarz: What do you think about:

    const staticClasses = (staticClass || '').split(' ')
    const dynamicClasses = (dynamicClass || '').split(' ')
    return Array.from(new Set([...staticClasses, ...dynamicClasses])).join(' ')

instead of L39 - L85 ?

astronomersiva: This is what I wrote initially but then I couldn't get tests to pass in PhantomJS.

lusarz: Ok, I understand. Sorry I didn't check it before writing my comment.

 If you want you can probably use solution:

    return concat(staticClass, stringifyClass(dynamicClass))
      .split(' ')
      .filter((item, index, array) => index === array.indexOf(item))
      .join(' ')

astronomersiva: I have updated my changes. Thanks for the suggestion :)

astronomersiva:
A gentle reminder @yyx990803 and @posva .

posva:
I wonder if as this may get executed a lot of times in an application, it may be worth writing a `concatUnique` function that behaves as a Set. But I also wonder if there is any technical benefit from not having class names repeated

fnlctrl:
I don't think it's worth the overhead to remove the duplicates (this is a serious one as it will be called on every class name render), and having duplicate class names doesn't really harm anything.

yyx990803:
I don't think the perf overhead would be worth it. Class merging is a pretty compute intensive part of the rendering process and it's not clear what we are getting in return aside from minor aesthetic improvements.

PR #8038, opened 2018-04-16
refactor: Add isString

243083df:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -2,7 +2,7 @@
 
 import { genHandlers } from './events'
 import baseDirectives from '../directives/index'
-import { camelize, no, extend } from 'shared/util'
+import { camelize, no, extend, isString } from 'shared/util'
 import { baseWarn, pluckModuleFunction } from '../helpers'
 
 type TransformFunction = (el: ASTElement, code: string) => string;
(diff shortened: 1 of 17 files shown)


src/compiler/codegen/index.js
```diff
@@ -499,7 +499,7 @@ function genProps (props: Array<{ name: string, value: any }>): string {
 
 /* istanbul ignore next */
 function generateValue (value) {
-  if (typeof value === 'string') {
+  if (isString(value)) {
```
lusarz: What do you think about `return isString(value) ? transformSpecialNewlines(value) : JSON.stringify(value)` notation ?

dsonet:
Is this really needed or something just more like over-engineering?

lusarz:
For me it's good approach to create `isString` like methods. It increases readability of code.

There are methods like `isTrue`, `isFalse`, `isObject` in `shared/util`, so why don't create similar for `isString` ?

Please also take a look here:
https://github.com/vuejs/vue/blob/f7b5c86ef2620c1d7ec275f86ae0f380ccaea4ef/src/platforms/web/util/class.js#L37-L39
Mixing different levels of abstraction (`isObject` near `typeof value === 'string'`) is not a good practice.

243083df:
As said lusarz, there is methods like isTrue|isFalse|isObject|isPrimitive, so i dont see over-engineering here

linghengqian (commented):
You need to resolve merge conflicts.

PR #8066, opened 2018-04-22
Refactor: get rid of toArray method

lusarz:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/global-api/use.js
+++ b/src/core/global-api/use.js
@@ -1,16 +1,13 @@
 /* @flow */
 
-import { toArray } from '../util/index'
-
 export function initUse (Vue: GlobalAPI) {
-  Vue.use = function (plugin: Function | Object) {
+  Vue.use = function (plugin: Function | Object, ...args) {
     const installedPlugins = (this._installedPlugins || (this._installedPlugins = []))
     if (installedPlugins.indexOf(plugin) > -1) {
       return this
     }
 
     // additional parameters
-    const args = toArray(arguments, 1)
     args.unshift(this)
     if (typeof plugin.install === 'function') {
       plugin.install.apply(plugin, args)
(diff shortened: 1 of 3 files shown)


nickmessing:
Since at the moment we transpile vue to ES5.1 for IE support - that does not bring any benefit and actually decreases performance:
```js
  for (var _len = arguments.length, b = Array(_len > 1 ? _len - 1 : 0), _key = 1; _key < _len; _key++) {
    b[_key - 1] = arguments[_key];
  }
```

lusarz:
I think main benefit of these changes is related with code quality. If it is possible to write something with less amount of code, why don't do it ?

I see also a few places in `vue` code where spread operator like `...args` is used, so it would be nice to be consistent:
https://github.com/vuejs/vue/blob/f2b476d4f4f685d84b4957e6c805740597945cde/src/core/observer/array.js#L27

https://github.com/vuejs/vue/blob/472a2896bd4f156be168edfecb6ac432b853beb4/src/platforms/weex/entry-framework.js#L163

nickmessing:
You are right there, it should be consistent, but I am not so sure about code quality. Sometimes I'd rather think more about performance when developing a framework. And using `Array.prototype.slice.call(list, start)` may bring performance benefit depending on array usage (more info [here](https://v8project.blogspot.ro/2017/09/elements-kinds-in-v8.html)). Probably needs more feedback from the team.

yyx990803:
The original usage of `toArray` instead of spread arguments throughout codebase is intentional. Using `...args` results in duplicated inline code chunks being generated, causing larger bundle sizes, slower parsing and less optimized performance.

PR #8128, opened 2018-05-03
Checked for undefined/null on component during destroy

carsongee:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When we run our test suite with relatively >beta13 on vue-test-utils we end up with this issue failing a portion of our test suite as it appears somehow the componentInstance is undefined by the time this hook gets called.

I spend about an hour trying to find where to put a test where the destroy hook gets called, but couldn't figure out how to test it properly.  I even attempted to find a good spot by having that function `throw`, but even that wasn't very helpful, so I'd love some help with that.

### DIFF
--- a/src/core/vdom/create-component.js
+++ b/src/core/vdom/create-component.js
@@ -86,7 +86,7 @@ const componentVNodeHooks = {
 
   destroy (vnode: MountedComponentVNode) {
     const { componentInstance } = vnode
-    if (!componentInstance._isDestroyed) {
+    if (!!componentInstance && !componentInstance._isDestroyed) {
       if (!vnode.data.keepAlive) {
         componentInstance.$destroy()
       } else {


posva:
Did you make sure it also fails without vue test utils?

carsongee:
Good point, no it doesn't if I just do the `$mount` myself. So it sounds like something is going on over there with the VNode lifecycle too, but the fix here seems relevant as well, since it seems harmless to null check before operating on the object. Of course unless it was left out explicitly so that this sort of thing would happen to help track down other bugs lol

posva:
I think you should share the test that caused the error (a boiled down version is better 😄 ). pinging @eddyerburgh just in case but I will also look at it

carsongee:
I thought you might want that, here is the most boiled down version I have that still fails:
```
<template>
  <div class="tasks__task-container col-sm-6">
    <transition name="fade" mode="out-in">
      <div v-if="!isEditing" key="editing">
      </div>
      <task-form
        v-else
        key="editing"
        ></task-form>
    </transition>
  </div>
</template>
<script>
import TaskForm from './task-form.vue'

export default {
  components: { TaskForm, },
  props: {
    task: {
      type: Object,
      default: () => {},
    },
    isCreator: Boolean,
    isParticipant: Boolean,
  },
  data() {
    return {
      isCreating: false,
      isEditing: false,
    }
  },
}
</script>
```

and the spec (using jest if that helps):
```
import { shallow } from '@vue/test-utils'
import Vue from 'vue'

import TaskItem from './task-item.vue'

describe('Test suite for Task Item', () => {
  let wrapper
  let vm

  beforeEach(() => {
    wrapper = shallow(TaskItem, {
      propsData: {
        task: {
          id: 55,
        },
        isCreator: false,
        isParticipant: false,
      },
    });
    ({ vm } = wrapper)
  })
  it('shows the tasks form when isEditing, otherwise shows the task', () => {
    vm.isEditing = true
  })
})
```

PR #8130, opened 2018-05-04
fix(SSR): Properly handle newlines and whitespace in SSR classnames (fix #7859)

### LINKED ISSUE #7859: Vue SRR renders newline characters into class attributes
### Version
2.5.16

### Reproduction link
[Gist](https://gist.github.com/brophdawg11/07aee0e6e20b911993e6dc1eaa6b2d04)

### Steps to reproduce
* Download the above gist to an `index.js` file
* `npm install --save vue vue-server-renderer`
* `node index.js`

### What is expected?
Once rendered in the browser, all 6 `long-classname-*` class styles should be applied to the second element.

Because the class attribute is rendered with `long-classname-3\n`, the styles for `long-classname-3` are not applied

---
This can be worked around by adding a space before the mid-attribute line break, however this doesn't play nicely with the ESLint `no-trailing-spaces` rule and the settings for many common editors.

brophdawg11:
When encountering newlines in SSR classnames, do not render "\n" and
instead replace the new line with a space.

Once newlines are removed, indentation from subsequent lines can leave
classnames abnormally spaced out, so replace multiple spaces with a
single space

fix #7859

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/platforms/web/compiler/modules/class.js
+++ b/src/platforms/web/compiler/modules/class.js
@@ -22,7 +22,7 @@ function transformNode (el: ASTElement, options: CompilerOptions) {
     }
   }
   if (staticClass) {
-    el.staticClass = JSON.stringify(staticClass.replace(/\n/g, ' ').replace(/ +/g, ' '))
+    el.staticClass = JSON.stringify(staticClass.replace(/\s+/g, ' '))
   }
   const classBinding = getBindingAttr(el, 'class', false /* getStatic */)
   if (classBinding) {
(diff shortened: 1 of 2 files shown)


src/platforms/web/compiler/modules/class.js
```diff
@@ -22,7 +22,7 @@ function transformNode (el: ASTElement, options: CompilerOptions) {
     }
   }
   if (staticClass) {
-    el.staticClass = JSON.stringify(staticClass)
+    el.staticClass = JSON.stringify(staticClass.replace(/\n/g, ' ').replace(/ +/g, ' '))
```
williamchong007: would a `/\s+/g` be better? I was looking into this issue as well

brophdawg11: Good catch - updated

PR #8138, opened 2018-05-06
fix: In development mode, warns if user tries to Vue.set a property t…

### LINKED ISSUE #8129: when using $set, why not make the prop reactive when is already exists?
### Version
2.5.16

### Reproduction link
[https://codepen.io/JJPandari/pen/gzLVBq?editors=1010](https://codepen.io/JJPandari/pen/gzLVBq?editors=1010)

### Steps to reproduce
See the codepen snippet. Follow the comment there to change the vm's data and see what happens.

### What is expected?
Even if the prop already exists, using `set` still makes it reactive, thus trigger view update.

Using `set` later doesn't update the view.

---
Related source code: https://github.com/vuejs/vue/blob/3eb37acf98e2d9737de897ebe7bdb7e9576bcc21/src/core/observer/index.js#L192 I think most users would expect `set` to make the prop reactive whenever it's used. I initially opened [an issue for the api doc](https://github.com/vuejs/vuejs.org/issues/1601) because it wasn't clear (for me) about this. But the comment in the source code is.

bigtunacan:
…hat already exists.

In development mode, warns if user tries to Vue.set a property that already exists. Issue reported
in #8129. Codepen demonstrating the issue available at
https://codepen.io/chrisvfritz/pen/rvzgBR?editors=1010

fix #8129

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -196,10 +196,15 @@ export function defineReactive (
  * already exist.
  */
 export function set (target: Array<any> | Object, key: any, val: any): any {
-  if (process.env.NODE_ENV !== 'production' &&
-    (isUndef(target) || isPrimitive(target))
-  ) {
-    warn(`Cannot set reactive property on undefined, null, or primitive value: ${(target: any)}`)
+  if (process.env.NODE_ENV !== 'production') {
+    if (isUndef(target) || isPrimitive(target)) {
+      warn(`Cannot set reactive property on undefined, null, or primitive value: ${(target: any)}`)
+    }
+    if (Object.getOwnPropertyDescriptor(target, key) &&
+      (typeof (Object.getOwnPropertyDescriptor(target, key).get) === 'undefined') &&
+      !Array.isArray(target)) {
+      warn(`Cannot enable reactivity on a property that is already defined: ${(key: any)}`)
+    }
   }
   if (Array.isArray(target) && isValidArrayIndex(key)) {
     target.length = Math.max(target.length, key)


bigtunacan:
@posva I'm trying to add a test for this, but struggling a bit.  Here is a codepen of a test that I expect should pass for this.  https://codepen.io/bigtunacan/pen/rvGgoB?editors=0010#

chrisvfritz:
> I'm trying to add a test for this, but struggling a bit.

Could you provide more information on where you struggled? For example, are you not sure where the test should go, or did running the tests not work, etc?

bigtunacan:
@chrisvfritz I placed the test in `test/unit/modules/observer/observer.spec.js`.

The test fails with the generic message `FAILED Observer Cannot enable reactivity on a property that is already defined`

So then I ran the tests against Chrome with inspector to see what is happening; the warning is getting hit in test the same as in dev, but yet it fails in my test (which I also linked in the codepen above).

I could also push my changes onto this pull request, but I didn't necessarily want to do that when I know the test isn't working yet.

chrisvfritz:
@bigtunacan You may have to wrap your expectation in `waitForUpdate` (like in [this test](https://github.com/vuejs/vue/blob/dev/test/unit/modules/observer/observer.spec.js#L307-L311)).

chrisvfritz:
Can you upload the exact code you're trying? It's OK if the test is failing - we'll squash when we merge anyway.

bigtunacan:
@chrisvfritz Ok; I've updated the PR

test/unit/modules/observer/observer.spec.js
```diff
@@ -316,6 +316,21 @@ describe('Observer', () => {
...
+        person: {
+          age: 32
+        }
+      }
+    }).$mount()
+    vm.person.job = 'Programmer'
+    Vue.set(vm.person, 'job', 'Educator')
+    waitForUpdate(() => {
+      expect(`Cannot enable reactivity on a property that is already defined`).toHaveBeenWarned()
+    }).end(done)
```
chrisvfritz: I think this should be `then` rather than `end`.

bigtunacan: I've tried that as well, but the result is the same.

test/unit/modules/observer/observer.spec.js
```diff
@@ -316,6 +316,21 @@ describe('Observer', () => {
     }).then(done)
   })
 
+  it('Cannot enable reactivity on a property that is already defined', done => {
+    const vm = new Vue({
```
dhensche: A Vue instance must have a template prop or render fn, your test is throwing an error before getting to the checks

bigtunacan: @dhensche Thanks; that was it.

bigtunacan:
@posva I believe this should be ready to merge now.

bigtunacan:
@posva Just curious since it's been sitting for awhile and shows as "Changes requested".  Are you still waiting on anything from me to merge this in?  I thought everything was good now?

yyx990803:
I think it makes more sense to convert a property to a reactive one if it's not, which should be the intention in most cases anyways.

bigtunacan:
> I think it makes more sense to convert a property to a reactive one if it's not, which should be the intention in most cases anyways.

I think the issue is that someone may add a property with standard assignment after the fact as shown in this codepen from the original issue https://codepen.io/chrisvfritz/pen/rvzgBR?editors=1010

Most likely this is not what the person meant to do so it would be good to provide a warning in dev environments.

It could be changed to set the property to reactive, but that would seem inconsistent as you would being with a property that is not reactive and then at some point in time it becomes reactive.

chrisvfritz:
I agree with you @bigtunacan. We just had an internal discussion about this in the Vue team and never came up with any valid use cases for making an unreactive property reactive - and the ability just opens a subtle hole for reactivity bugs. So a warning does seem ideal here.

PR #8143, opened 2018-05-07
Support `components` obj in functional components (fix #7492)

### LINKED ISSUE #7492: Functional single file component with components option.
### Version
2.5.13

### Reproduction link

NG pattern (functional)
https://codesandbox.io/s/004vv2onw0

OK pattern (no functional)
https://codesandbox.io/s/q9k5q8qq56

### Steps to reproduce
I found can't use `components` option when  `functional` single file component.

```html
<template functional>
  <div>
    <some-children />
  </div>
</template>

<script>
import SomeChildren from "./SomeChildren"

export default {
  components: {
    SomeChildren
  }
}
</script>
```

It's occure `Unknown custom element`.

### What is expected?
Not occure `Unknown custom element` and use child component

It's occure `Unknown custom element`

---

In workaround, it not occure when use `Vue.component`.

```js
import Vue from "vue"
import SomeChildren from "./SomeChildren"
Vue.component("some-children", SomeChildren);

export default {}

// can use  <some-children />
```

hirokiosame:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/vdom/create-functional-component.js
+++ b/src/core/vdom/create-functional-component.js
@@ -6,6 +6,7 @@ import { resolveInject } from '../instance/inject'
 import { normalizeChildren } from '../vdom/helpers/normalize-children'
 import { resolveSlots } from '../instance/render-helpers/resolve-slots'
 import { installRenderHelpers } from '../instance/render-helpers/index'
+import { extend } from 'shared/util'
 
 import {
   isDef,
@@ -39,6 +40,10 @@ export function FunctionalRenderContext (
     // $flow-disable-line
     parent = parent._original
   }
+
+  contextVm.$options = Object.create(contextVm.$options)
+  contextVm.$options.components = extend(Object.create(contextVm.$options.components), options.components)
+
   const isCompiled = isTrue(options._compiled)
   const needNormalization = !isCompiled
 
--- a/src/core/vdom/create-functional-component.js
+++ b/src/core/vdom/create-functional-component.js
@@ -27,20 +27,20 @@ export function FunctionalRenderContext (
   const options = Ctor.options
   // ensure the createElement function in functional components
   // gets a unique context - this is necessary for correct named slot check
-  let contextVm
+  const contextVm = Object.create(parent)
   if (hasOwn(parent, '_uid')) {
-    contextVm = Object.create(parent)
     // $flow-disable-line
     contextVm._original = parent
   } else {
     // the context vm passed in is a functional context as well.
     // in this case we want to make sure we are able to get a hold to the
     // real context instance.
-    contextVm = parent
+
     // $flow-disable-line
     parent = parent._original
   }
 
+  // support functional components hash
   contextVm.$options = Object.create(contextVm.$options)
   contextVm.$options.components = extend(Object.create(contextVm.$options.components), options.components)
 
--- a/test/unit/features/options/functional.spec.js
+++ b/test/unit/features/options/functional.spec.js
@@ -242,6 +242,34 @@ describe('Options functional', () => {
     expect(vm.$el.childNodes[1].namespaceURI).toContain('svg')
   })
 
+  it('should accept components', () => {
+    const HelloComp = {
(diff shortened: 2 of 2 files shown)


y-nk:
👍

kushalmahajan:
In which circumstances will you need this functionality? Just curious!

y-nk:
@kushalmahajan we (the projects i'm involved with) tends to write all component in single vue files with template blocks and avoid render function as much as possible since without JSX it could become quite messy very fast. Here's the typical use-case :

```
<template functional>
  <div class="you-can-imagine-complex-domtree-here">
    <my-custom-component />
  </div>
</template>

<script>
import MyCustomComponent from 'somehwere-else'

export default {
  components: { MyCustomComponent }
}
</script>
```

If i'm not mistaken, this would be the only reason for such a thing.

kushalmahajan:
@y-nk. I see! Same like the normal components property on non-functional Vue components.  

If it helps, then I would like to mention that if you write functional components with render functions, the components property is not required for it to work. I don't know the underlying reason but it's definitely not mandatory there. Plus, I am writing just **.js file** only so the advantage one gets is **full fledged dev tooling**,  which .vue is still not offering ,  apart from just doing function without template as a component.

Below is an  example of a simple dropdown component I built
```
import { DropdownWrapper } from "./styles";
import VList from "../list/list";
import VItem from "../item";

const Dropdown = {
  functional: true,
  props: {
    list: Array,
    wrapperStyles: Object
  },
  // eslint-disable-next-line
  render(h, { props, listeners }) {
    const handleClick = listeners.click ? listeners.click : () => {};
    return (
      <DropdownWrapper styles={props.wrapperStyles || {}}>
        <VList
          list={props.list || []}
          renderItem={item => <VItem size="sm">{item}</VItem>}
          onClick={handleClick}
        />
      </DropdownWrapper>
    );
  }
};
export default Dropdown;

```

hitautodestruct:
Can this go into 2.6.1? Please?

rodrigopedra:
This would be very handy, sometimes I separate some components that contains just some svg icons using vue-svg-loader and has no additional logic within them. I wish I could write them as functional.

Writing them with render function is unpractical in larger teams with varied skill levels.

I read in a [comment in another PR](https://github.com/vuejs/vue/pull/8287#issuecomment-432327101) that  in v3 stateful components will be much faster to initialize, and I agree that we should keep functional components minimal and don't add all the options that a stateful component has.

But using other components is great for composition and, in my opinion, would be a great addition and allow for cleaner components. 

Also we might not be able to migrate all precious codebases to v3 when it is released, so the faster initialization times and memory gains we could have now with this is much beneficial, in my opinion.

hirokiosame:
Some new info on Evan's vision of functional components: https://github.com/vuejs/rfcs/blob/functional-async-api-change/active-rfcs/0000-functional-async-api-change.md

Based on this, it seems more unlikely this will ever get merged.

hitautodestruct:
> Some new info on Evan's vision of functional components: https://github.com/vuejs/rfcs/blob/functional-async-api-change/active-rfcs/0000-functional-async-api-change.md
> 
> Still a proposal but based on this, it seems more unlikely this will ever get merged.

Based on the information in the above link, it's best to avoid future use of the `<template functional>` option for functional components and focus on purely render functional components if needed.

Even though he states that there will be a graceful degradation of this feature.

posva:
The v3 the difference between functional and regular components won't be as significant performance wise. So, you can keep writing templates, it may be easier to refractor to functions now and maybe again in v3

y-nk:
@posva regarding the RFC, i was wondering why `<template functional>` was dropped since it's sugar syntax prior template compilation. wouldn't it be possible to keep the keyword/functionality but wire it to a plain function calling the current template render signature properly ?

PR #8179, opened 2018-05-12
Add iterable supports for v-for 

mymyoux:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Adds support for iterable object implementing [Symbol.iterator]()
It shouldn't have an impact for existing code because non iterable objects are not impacted.
Vue will have the same behaviour than a for...of expression for iterables. (actual interest for using iterables)
Allows us to pass an iterable directly to a template without giving internal properties or converting it to an array

I wrote 4 new tests. No existing tests needed changes.
All tests passed. I ran tests with PhantomJS (lack of Symbol support), Firefox and Chrome.

### DIFF
--- a/src/core/instance/render-helpers/render-list.js
+++ b/src/core/instance/render-helpers/render-list.js
@@ -26,19 +26,13 @@ export function renderList (
     }
   } else if (isObject(val)) {
     if (hasSymbol && val[Symbol.iterator]) {
-      const iterator:Iterator<any> = val[Symbol.iterator]()
       ret = []
-      i = 0
-      let next
-      do {
-        next = iterator.next()
-        if (!next.done) {
-          key = i
-          ret.push(render(next.value, key, i))
-          i++
-        }
+      const iterator: Iterator<any> = val[Symbol.iterator]()
+      let result = iterator.next()
+      while (!result.done) {
+        ret.push(render(result.value, ret.length))
+        result = iterator.next()
       }
-      while (!next.done)
     } else {
       keys = Object.keys(val)
       ret = new Array(keys.length)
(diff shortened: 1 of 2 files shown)


mitar:
Related: #5893

mitar:
So I have been using the following for a while now: https://github.com/vuejs/vue/blob/cb062c50cdd21628dcf758fe54c84d17d75cb2ca/src/core/instance/render-helpers/render-list.js

But I thought that in #5893 it is said that this will not go into core for now?

src/core/instance/render-helpers/render-list.js
```diff
@@ -25,11 +25,27 @@ export function renderList (
...
+    if (hasSymbol && val[Symbol.iterator]) {
+      const iterator:Iterator<any> = val[Symbol.iterator]()
+      ret = []
+      i = 0
+      let next
+      do {
+        next = iterator.next()
+        if (!next.done) {
+          key = i
+          ret.push(render(next.value, key, i))
```
mitar: Why not just `ret.push(render(next.value, ret.length))`?

mymyoux: good point I can update this

mymyoux:
> But I thought that in #5893 it is said that this will not go into core for now?

I'm aware of this but I think if you choose to implement [Symbol.iterator] you choose to break old browsers compatibility. This way your iterable object has the same behaviour in Vue than in other parts of your code.
What do you think ?

mitar:
I personally like it, I was all for #5893.

PR #8214, opened 2018-05-21
feat(warn): warn when an existing property starting with $ is not pro…

### LINKED ISSUE #8213: Vue props throw incomprehensible error when passing variables starting with symbols
### Version
2.5.17-beta.0

### Reproduction link
[https://codesandbox.io/s/r0lr534k94](https://codesandbox.io/s/r0lr534k94)

### Steps to reproduce
Open the console and see the error:
```Javascript
proxyConsole.js:72 [Vue warn]: Property or method "$_testMixin_testData" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <TestMixin> at /src/App.vue
       <Root>
```

### What is expected?
It's expected for the prop to be defined

The prop is undefined if the variable contains symbols in the beginning

---
I am not completely sure if this is a bug or intended, since I think it is a bug I opened this issue, yet if this is intended the error shown should be giving more details, it took me some time to understand what was the error because it isn't clear just from the error message.

posva:
…perly accessed

Closes #8213

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/proxy.js
+++ b/src/core/instance/proxy.js
@@ -24,7 +24,7 @@ if (process.env.NODE_ENV !== 'production') {
     )
   }
 
-  const warnStartingWithDollar = (target, key) => {
+  const warnReservedPrefix = (target, key) => {
     warn(
       `Property "${key}" must be accessed with "$data.${key}" because ` +
       'properties starting with "$" or "_" are not proxied in the Vue instance to ' +
@@ -58,7 +58,7 @@ if (process.env.NODE_ENV !== 'production') {
       const isAllowed = allowedGlobals(key) ||
         (typeof key === 'string' && key.charAt(0) === '_' && !(key in target.$data))
       if (!has && !isAllowed) {
-        if (key in target.$data) warnStartingWithDollar(target, key)
+        if (key in target.$data) warnReservedPrefix(target, key)
         else warnNonPresent(target, key)
       }
       return has || !isAllowed
@@ -68,7 +68,7 @@ if (process.env.NODE_ENV !== 'production') {
   const getHandler = {
     get (target, key) {
       if (typeof key === 'string' && !(key in target)) {
-        if (key in target.$data) warnStartingWithDollar(target, key)
+        if (key in target.$data) warnReservedPrefix(target, key)
         else warnNonPresent(target, key)
       }
       return target[key]
--- a/src/core/instance/proxy.js
+++ b/src/core/instance/proxy.js
@@ -55,7 +55,8 @@ if (process.env.NODE_ENV !== 'production') {
   const hasHandler = {
     has (target, key) {
       const has = key in target
-      const isAllowed = allowedGlobals(key) || (typeof key === 'string' && key.charAt(0) === '_')
+      const isAllowed = allowedGlobals(key) ||
+        (typeof key === 'string' && key.charAt(0) === '_' && !(key in target.$data))
       if (!has && !isAllowed) {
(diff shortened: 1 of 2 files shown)


src/core/instance/proxy.js
```diff
@@ -24,6 +24,16 @@ if (process.env.NODE_ENV !== 'production') {
...
+  const warnStartingWithDollar = (target, key) => {
+    warn(
+      `Property "${key}" must be accessed with "$data.${key}" because ` +
+      'properties starting with "$" or "_" are not proxied in the Vue instance to ' +
+      'prevent conflicts with Vue internals' +
+      'See: https://vuejs.org/v2/api/#data',
+      target
+    )
+  }
+
```
Justineo: Should we do the same thing when we come across properties prefixed with `_`?

posva: that's true, I thought there is an eslint warning for it. Let me check again

posva: I added that as well but I wonder if it is correct. I'm not sure what _babel helpers_ are we talking about in 8e1478e3

carneiroskeeled: Why not proxying both? I changed all mixins in a big project to respect this rule on Style Guide (https://vuejs.org/v2/style-guide/#Private-property-names-essential) and now when I am trying to pass it to a component prop it gets there undefined because it is not being proxied

posva: to prevent conflicts. We cannot check all existing properties (which would be ideal), for conflicts, so we test that instead.
Nowadays, if you want to create plugins that inject custom properties, using Symbols is probably the safest way

carneiroskeeled: Yet the style guide recommends it, I think this issue should be given some taught, either change the rule for private property names or proxy everything, (maybe giving a warning in development process). Not proxying this is removing functionality and forcing developers that follow the recommended style guide find workarounds because of limited functionality to prevent others who may not read the documentation and override vue properties.

src/core/instance/proxy.js
```diff
@@ -45,9 +55,11 @@ if (process.env.NODE_ENV !== 'production') {
   const hasHandler = {
     has (target, key) {
       const has = key in target
-      const isAllowed = allowedGlobals(key) || (typeof key === 'string' && key.charAt(0) === '_')
+      const isAllowed = allowedGlobals(key) ||
+        (typeof key === 'string' && key.charAt(0) === '_' && !(key in target.$data))
       if (!has && !isAllowed) {
-        warnNonPresent(target, key)
+        if (key in target.$data) warnStartingWithDollar(target, key)
```
Justineo: Does it have to be some property starting with preserved prefixes here? It feels a little weird to me that after these checks we concluded that the property starts with `_` or `$`.

src/core/instance/proxy.js
```diff
@@ -24,6 +24,16 @@ if (process.env.NODE_ENV !== 'production') {
     )
   }
 
+  const warnStartingWithDollar = (target, key) => {
```
Justineo: Something like `warnPreservedPrefix`?

src/core/instance/proxy.js
carneiroskeeled: Proxying properties starting with $_ would be an option

PR #8217, opened 2018-05-22
fix #8093: Use custom toString method when interpolating objects if implemented

### LINKED ISSUE #8093: Use toString() on object interpolation before toJSON()
### Version
2.5.16

### Reproduction link
[https://jsfiddle.net/Lyq90ukx/1/](https://jsfiddle.net/Lyq90ukx/1/)

### Steps to reproduce
When interpoling a object, it uses the `JSON.stringify` method instead of `toString`, which is normally the proper method to set a way to convert an object to a string. We could try to see if the `toString` method is present before calling the `toJSON` one ?

### What is expected?
In the fiddle: 
>string: test

In the fiddle:
>string: { "json": "jsonVal" }

---
Thanks.

mathieutu:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Description:**

Hi, 
When interpolating an object, it uses the JSON.stringify method instead of toString, which is normally the proper method to set a way to convert an object to a string. 
This PR allow you to use your custom `toString` method to render your object if it exists. It's tested (I've also added a test for toJson), and it should not be a breaking change.

Thanks you for your amazing work folks, and see you in Paris next month!

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -77,7 +77,7 @@ export function isValidArrayIndex (val: any): boolean {
 export function toString (val: any): string {
   return val == null
     ? ''
-    : isPlainObject(val) && val.toString === _toString
+    : Array.isArray(val) || (isPlainObject(val) && val.toString === _toString)
       ? JSON.stringify(val, null, 2)
       : String(val)
 }
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -77,7 +77,7 @@ export function isValidArrayIndex (val: any): boolean {
 export function toString (val: any): string {
   return val == null
     ? ''
-    : typeof val === 'object'
+    : typeof val === 'object' && (!isPlainObject(val) || val.toString() === _toString.call(val))
       ? JSON.stringify(val, null, 2)
       : String(val)
 }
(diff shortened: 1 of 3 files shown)


mathieutu:
I'm adding some example of the problem to make me understood better:

[Edit Vue Template](https://codesandbox.io/s/0plonqx6jp)

@posva:

I can confirm that your (so used!) `Object.create(null)` example works exactly the same than before (like arrays), as it's not a "plain object" (`_toString(Object.create(null))` gives `[object Undefined]`).

And to write here what I told you yesterday, it _could_ be a BC for those who had a object with a `toString` method implemented and still want to use the `toJson`  in templates. I think we can easily consider that this is not a standard case. 😅

See you! 👋

src/shared/util.js
```diff
@@ -77,7 +77,7 @@ export function isValidArrayIndex (val: any): boolean {
 export function toString (val: any): string {
   return val == null
     ? ''
-    : typeof val === 'object'
+    : typeof val === 'object' && (!isPlainObject(val) || val.toString === _toString)
```
phanan: I think this function can really use early returns now.

mathieutu: As you want, I kept the original style to make as few changes as possible and let the diff easier to read, but it'll be easy to refactor.

PR #8240, opened 2018-05-23
Add async option

### LINKED ISSUE #8200: Re-implement async config option
Vue Test Utils has a synchronous API by walking all watchers and setting them to `sync`, but this implementation is buggy.

Reimplementing an async private config option would solve this problem for Vue Test Utils.

Add an `async` config option that would be used when watchers are created.

The same option that was removed in [this commit](https://github.com/vuejs/vue/commit/ad11aa122cd09a4510fd9d8829e25f4a90b178c2#diff-4d10b522e21ed424211799a4e11c6103L191).

I'm happy to make a PR to add this.

eddyerburgh:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Closes #8200

We can use this in Vue Test Utils to run tests synchronously without editing Vue internals

### DIFF
--- a/src/core/observer/dep.js
+++ b/src/core/observer/dep.js
@@ -2,6 +2,7 @@
 
 import type Watcher from './watcher'
 import { remove } from '../util/index'
+import config from '../config'
 
 let uid = 0
 
@@ -36,6 +37,12 @@ export default class Dep {
   notify () {
     // stabilize the subscriber list first
     const subs = this.subs.slice()
+    if (process.env.NODE_ENV !== 'production' && !config.async) {
+      // subs aren't sorted in scheduler if not running async
+      // we need to sort them now to make sure they fire in correct
+      // order
+      subs.sort((a, b) => a.id - b.id)
+    }
     for (let i = 0, l = subs.length; i < l; i++) {
       subs[i].update()
     }
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -118,8 +118,8 @@ export default ({
   mustUseProp: no,
 
   /**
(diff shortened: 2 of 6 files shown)


posva:
Do you think it would make sense to use an ENV variable so it is completely stripped out in normal builds?

trollepierre:
what's up with this amazing feature?

src/core/observer/dep.js
```diff
@@ -36,6 +37,12 @@ export default class Dep {
   notify () {
     // stabilize the subscriber list first
     const subs = this.subs.slice()
+    if (process.env.NODE_ENV !== 'production' && !config.async) {
```
bost-h: Is it necessary to check for `NODE_ENV` ? I usually have different tests configurations, and one of them sets `NODE_ENV` to production to be as close as possible to the production behaviour. By reading this code, I guess my tests won't work.

sobrinho:
Waiting for that too, Vuetify is near impossible to test on vue-test-utils.

I'm seeing errors related to _transitionClasses not defined everywhere.

benirose:
Is this still going into 2.6?

eddyerburgh:
It should be out in 2.5.18

aldarund:
@eddyerburgh is there any ETA on 2.5.18 ?

agualis:
> @eddyerburgh is there any ETA on 2.5.18 ?
🎉 🎉 
https://github.com/vuejs/vue/releases/tag/v2.5.18-beta.0

PR #8250, opened 2018-05-26
fix(#7913): Prevent erroneous warning when using <slot> inside slot-scope

### LINKED ISSUE #7913: Unexpected "duplicate presence of slot"
### Version
2.5.2

### Reproduction link
[https://codesandbox.io/s/mzvkppmvo8](https://codesandbox.io/s/mzvkppmvo8)

### Steps to reproduce
1. I created component with scoped slot (AppSwitcher.vue)
2. Then I use it in another component with their own slot (HelloWorld.vue with slot "subtext")
3. Add some element to slot (div in App.vue) 

### What is expected?
It should work without errors

Changes in AppSwitcher.vue caus "Duplicate presence of slot "subtext" found in the same render tree" error but there are no duplicates.

---
Also, adding `slot-scope` to div in App.vue solves problem and no error there, but why it happens without `slot-scope`?

mpbarlow:
Because slotNodes inside a slot-scope context are already set to \_rendered = true after initial
render, the warning for duplicate slot presence always fires when a slot-scope prop change triggers
a re-render. With this change, the compiler tracks whether any slot-scoped elements have been
encountered yet at the point the slot is compiled. If so, the direct ancestors of the slot are checked
for slot-scope presence, and if one is found, the warning is supressed.

This is admittedly not a perfect solution, as within a slot-scope context the warning now does not fire even when there _are_ duplicate slots, but I couldn't find a good way to get around that.

fix #7913

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -18,6 +18,7 @@ export class CodegenState {
   maybeComponent: (el: ASTElement) => boolean;
   onceId: number;
   staticRenderFns: Array<string>;
+  slotScopeinAst: boolean;
 
   constructor (options: CompilerOptions) {
     this.options = options
@@ -29,6 +30,8 @@ export class CodegenState {
     this.maybeComponent = (el: ASTElement) => !isReservedTag(el.tag)
     this.onceId = 0
     this.staticRenderFns = []
+    // we can skip checks for slot-scope parents if we haven't seen any
+    this.slotScopeinAst = false
   }
 }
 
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks for the PR! I managed to completely remove the "no duplicated slots" restrictions in 530ca1b2, so the warning is no longer necessary.

PR #8287, opened 2018-06-01
Methods in functional components

hirokiosame:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Currently it's difficult to use the functional template since you lose access to closures where as the render method can reference closed-over functions.

### DIFF
--- a/src/core/vdom/create-functional-component.js
+++ b/src/core/vdom/create-functional-component.js
@@ -6,6 +6,7 @@ import { resolveInject } from '../instance/inject'
 import { normalizeChildren } from '../vdom/helpers/normalize-children'
 import { resolveSlots } from '../instance/render-helpers/resolve-slots'
 import { installRenderHelpers } from '../instance/render-helpers/index'
+import { extend } from 'shared/util'
 
 import {
   isDef,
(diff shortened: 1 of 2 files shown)


Kingwl:
why `functional` component need method?

hirokiosame:
@Kingwl I already explained in the PR body but, when using a render method instead of using a functional template, I can write functions outside of the component and reference them via closure. However, when using a template, I no longer have access to the closure.

I can reference them via `$options.methods` but seems hacky and would prefer official support in offering a similar API to stateful components.

Kingwl:
I respect your work, but I think there will be a better solution
/ cc: @posva

hirokiosame:
Here's an issue in vue-loader (it was originally in vue but closed) that might explain the use-case more clearly: https://github.com/vuejs/vue-loader/issues/1291

What are you thinking @Kingwl ?

yyx990803:
I don't think we should be going further along the route of "making functional components work like normal components in a Vue file", for the following reasons:

- In v3, stateful components are much faster to initialize, and this should make it less necessary to use functional components just for performance
- v3 stateful components can also return multiple root nodes

If we want to make functional components fully equivalent to stateful ones in a Vue file, we go down the rabbit hole of supporting components, directives, methods, provide/inject and all other options... which eventually defeats the simplicity of functional components. I think we should go back and make functional components more about "functions".

PR #8297, opened 2018-06-05
Adds handler for kebab-case event for v-bind:sync; fix #6428

### LINKED ISSUE #6428: `v-bind.sync` listens for camelCase event
This markup:

```html
<component :max-items.sync="showItems"></component>
```

Currently compiles to:

```javascript
function render() {
    with(this) {
        return _c('component', {
            attrs: {
                "max-items": showItems
            },
            on: {
                "update:maxItems": function($event) {
                    showItems = $event
                }
            }
        })
    }
}
```

I've always written event handlers in kebab-case. It surprised me that the compiler created an event handler which is in camelCase.

The compiler should output `.sync` event handlers in camelCase:

```javascript
            on: {
                "update:max-items": function($event) {
                    showItems = $event
                }
            }
```

shortdiv:

- [x] Bugfix
- [x] Feature

- [x] No

**Other information:**

Addresses #6428 and https://github.com/vuejs/vuejs.org/issues/1648

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -5,7 +5,7 @@ import { parseHTML } from './html-parser'
 import { parseText } from './text-parser'
 import { parseFilters } from './filter-parser'
 import { genAssignmentCode } from '../directives/model'
-import { extend, cached, no, camelize, kebabize } from 'shared/util'
+import { extend, cached, no, camelize, hyphenate } from 'shared/util'
 import { isIE, isEdge, isServerRendering } from 'core/util/env'
 
 import {
@@ -541,7 +541,7 @@ function processAttrs (el) {
             syncGen = genAssignmentCode(value, `$event`)
             addHandler(
               el,
-              `update:${kebabize(name)}`,
+              `update:${hyphenate(name)}`,
               syncGen
             )
             addHandler(
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -5,7 +5,7 @@ import { parseHTML } from './html-parser'
 import { parseText } from './text-parser'
 import { parseFilters } from './filter-parser'
 import { genAssignmentCode } from '../directives/model'
-import { extend, cached, no, camelize } from 'shared/util'
+import { extend, cached, no, camelize, kebabize } from 'shared/util'
 import { isIE, isEdge, isServerRendering } from 'core/util/env'
 
 import {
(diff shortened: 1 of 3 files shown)


chrisvfritz:
@yyx990803 Is there anything more that should be one on this? Being able to universally recommend kebab-case event names will simplify a couple sections of the docs and allow us to add [this rule](https://github.com/vuejs/eslint-plugin-vue/issues/356) to the ESLint plugin. 🙂

src/compiler/parser/index.js
```diff
@@ -538,10 +538,16 @@ function processAttrs (el) {
             name = camelize(name)
           }
           if (modifiers.sync) {
+            syncGen = genAssignmentCode(value, `$event`)
+            addHandler(
+              el,
+              `update:${kebabize(name)}`,
```
sodatea: The `hyphenate` function might be a better fit here.

chrisvfritz: Good point! 🙂 I forgot that one existed. @shortdiv, can you update to use `hyphenate` instead?

src/shared/util.js
```diff
@@ -158,6 +158,14 @@ export const camelize = cached((str: string): string => {
...
 })
 
+/**
+ * Kebabize a camelCased string
+ */
+const kebabizeRE = /([a-z])([A-Z])/g
+export const kebabize = cached((str: string): string => {
+  return str.replace(kebabizeRE, (m, p1, p2) => `${p1}-${p2}`).toLowerCase()
+})
+
```
chrisvfritz: Since we're no longer using this, can you delete it please?

shortdiv: You got it!

yyx990803:
// TODO merge #8842 with this one together

src/shared/util.js
```diff
@@ -158,6 +158,14 @@ export const camelize = cached((str: string): string => {
   return str.replace(camelizeRE, (_, c) => c ? c.toUpperCase() : '')
 })
 
+/**
+ * Kebabize a camelCased string
+ */
+const kebabizeRE = /([a-z])([A-Z])/g
+export const kebabize = cached((str: string): string => {
```
Fl0pZz: Please explain why the cached function is used

Justineo:
> // TODO revert changes in #8845 that will become unnecessary after merging this

Only [this line](https://github.com/vuejs/vue/pull/8845#discussion-diff-220505690L51) can be reverted. But I think it's okay to keep it as is.

chrisvfritz:
Thinking about this a little more, I wonder if `addHandler` itself should actually create both camelCase and kebab-case listeners for every event. This would fix cases where a library is emitting a camelCased event that can't be listened to in HTML (even though we warn against it in the docs) and also solve a [JSX issue](https://github.com/vuejs/babel-plugin-transform-vue-jsx/issues/164).

yyx990803:
@chrisvfritz that would significantly bloat the compiled output, not sure if it's worth it, especially when this is not an issue when using string templates.

PR #8371, opened 2018-06-18
Provide amd id for umd build

Globegitter:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**
We are running a dev-setup where all our code is converted to amd module and then concatenated to one file. This is the approach that [Bazel](https://bazel.build/) takes for js dev. For more details see: https://github.com/alexeagle/angular-bazel-example/wiki/Running-a-devserver-under-Bazel When depending on the vue build the problem is it does not have an id so when the umd file is being loaded requirejs complains about an anonymous module. Just by providing the id this is fixed and we can depend on vue from our application.

### DIFF
--- a/scripts/config.js
+++ b/scripts/config.js
@@ -71,6 +71,9 @@ const builds = {
     entry: resolve('web/entry-runtime.js'),
     dest: resolve('dist/vue.runtime.js'),
     format: 'umd',
+    amd: {
+      id: 'vue',
+    },
     env: 'development',
     banner
   },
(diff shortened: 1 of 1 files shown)


Globegitter:
Angular for example provides amd ids and it seems to work fine there. I see that in theory it makes it less portable, but how else would one want to call vue? I would be surprised if anyone would not have `import Vue from 'vue'` or the equivalent depending on their module format. But if there is sentiment against it that is also fair enough and this can be closed.

PR #8374, opened 2018-06-19
fix(transition): transition-group should only listen for first-level children's end events

verstoff:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

If in your list you have transitions on the nested elements, they will trigger the transitionend events that will cause finishing of transition-group's transition.

### DIFF
--- a/src/platforms/web/runtime/components/transition-group.js
+++ b/src/platforms/web/runtime/components/transition-group.js
@@ -116,6 +116,9 @@ export default {
         addTransitionClass(el, moveClass)
         s.transform = s.WebkitTransform = s.transitionDuration = ''
         el.addEventListener(transitionEndEvent, el._moveCb = function cb (e) {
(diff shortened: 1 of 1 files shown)


posva:
Could you provide a minimal repro of the bug please?

verstoff:
@posva 
Sure: [codesandbox repro](https://codesandbox.io/s/9yq45q0wk4)

posva:
Thanks, now it is much clearer! This seems reasonable to me but could break some people work, so it would be interesting if there is any reasonable usage of nested elements transitions to trigger that transition-group. This is because, as you said, it's very easy to workaround the problem by adding ` @transitionend.stop` to the nested element that also has a transition

PR #8422, opened 2018-06-27
fix(transition): check existence of `el.parentNode`, fix #8199

### LINKED ISSUE #8199: "Cannot read property '_pending' of null" when using transition component.
### Version
2.5.16

### Reproduction link
[https://jsfiddle.net/6sny3yq9/](https://jsfiddle.net/6sny3yq9/)

### Steps to reproduce
1. Open fiddle and type anything into the first field marked "broken". 
2. See in the console the error message "Error in nextTick: "TypeError: Cannot read property '_pending' of null"
3. Notice that Vue has stopped working
4. Refresh the fiddle and type anything into the second field marked "working"
5. Notice that everything works as expected

### What is expected?
When the value of the input field is "foo" the computed property `bar` will evaluate to true and the "a" span will display. 

When the user changes the value in the input field from "foo" to anything else, the computed property `bar` will evaluate to false and the "b" span (rendered using v-html) will display.

Initially, the value of the input field is "foo" and the computed property `bar` is evaluated to true, so the "a" span is displaying. 

But, when the user changes the value in the input field from "foo" to anything else, Vue crashes and this error message is displayed: Error in nextTick: "TypeError: Cannot read property '_pending' of null"

---
It takes this very specific scenario for the error to appear. In the fiddle, I've shown that by removing the `@click="show = !show"` event listener, Vue works as expected with no errors.

Here are two other changes that, when made individually, result in Vue working as expected:
  - removing the `<transition>` component surrounding the "a" span
  - using `<span v-else>{{ b }}</span>` instead of `<span v-else v-html="b"></span>`

haoqunjiang:
Fix #8199 
If the new parentNode gets a `textContent` or `innerHTML` property during patching, the `transition` node would have been detached early, which means `el.parentNode` no longer exists.
So this existential check is necessary.

This is an imperfect fix because while errors are suppressed, transition animations still get lost.
It is due to the different mechanism of `v-html` and VDOM element patching. But fixing that would be much more complex and I don't think we'd bother fixing such an edge case.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -32,6 +32,12 @@ export const emptyNode = new VNode('', {}, [])
 
 const hooks = ['create', 'activate', 'update', 'remove', 'destroy']
 
+function childrenIgnored (vnode) {
+  return vnode && vnode.data && vnode.data.domProps && (
+    vnode.data.domProps.innerHTML || vnode.data.domProps.textContent
+  )
+}
+
 function sameVnode (a, b) {
   return (
     a.key === b.key && (
@@ -39,6 +45,7 @@ function sameVnode (a, b) {
         a.tag === b.tag &&
         a.isComment === b.isComment &&
         isDef(a.data) === isDef(b.data) &&
+        !childrenIgnored(a) && !childrenIgnored(b) &&
         sameInputType(a, b)
       ) || (
         isTrue(a.isAsyncPlaceholder) &&


sodatea:
Oops, e2e test fails due to a bug of unpkg.com
https://github.com/unpkg/unpkg.com/issues/118

PR #8537, opened 2018-07-21
fix(types): relax the return type of props default option

ktsn:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Currently, the return type of `default` option is annotated with `object` since it is used for generating fresh object for each component instance. But there is another use case for `default` option of function in [portal-vue](https://github.com/LinusBorg/portal-vue) which is for dynamically generating unique value as default.
https://github.com/LinusBorg/portal-vue/blob/develop/src/components/portal.js#L16

The current typing breaks `this` type inference when the `default` function returns non-object type. I just relaxed the return type to `any` to avoid breaking the inference.

### DIFF
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -133,7 +133,7 @@ export type PropValidator<T> = PropOptions<T> | Prop<T> | Prop<T>[];
 export interface PropOptions<T=any> {
   type?: Prop<T> | Prop<T>[];
   required?: boolean;
-  default?: T | null | undefined | (() => object);
+  default?: T | null | undefined | (() => T | null | undefined);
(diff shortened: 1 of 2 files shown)


ktsn:
Maybe `() => T | null | undefined` is better instead of `() => any`?

fjc0k:
thx!
```ts
let zIndex = 5000

export default Vue.extend({
  name: 'Popup',
  props: {
    zIndex: {
      type: Number,
      default: (): number => zIndex++
    }
  }
})
```

PR #8581, opened 2018-07-31
fix(#8567): handle errors on immediate watchers

### LINKED ISSUE #8567: Immediate watcher skips parent(s) errorCaptured and jumps straight to the top error handler
### Version
2.5.16

### Reproduction link
[https://codesandbox.io/s/wkpy6x5r25](https://codesandbox.io/s/wkpy6x5r25)

### Steps to reproduce
Simply opening and running the sandbox should show the error. Instead of getting an alert that App.vue caught the error in HelloWorld's watcher, the error escapes all the way to the top level. To see the expected behavior, open Hello World and set the watcher to immediate: false. Then hit the button to trigger the watcher. In this case the error is caught by the parent, App.vue. In the immediate case, this does not happen.

### What is expected?
Errors in immediate watchers should be caught by the parent component(s)'s errorCaptured().

The error escapes to the nextTick hander. This handler has no ctx or vm to pass to the errorHandler. When errorHandler does not receive a vm, it cannot call the parent errorCaptured handlers.

---
This appears to be an oversight in src/core/instance/state.js. In the case options.immediate is true, cb.call(vm, watcher.value) should be wrapped in a try catch that explicitly calls handleError with the vm.

A proposed fix:
```
    if (options.immediate) {
      try {
        cb.call(vm, watcher.value)
      } catch (e) {
        handleError(e, vm, `immediate call for watcher ${watcher.expression}`)
      }
    }
```

afontcu:
As stated in the [related issue](https://github.com/vuejs/vue/issues/8567), the handle callback call should be wrapped in a try/catch that explicitly calls handleError.

fix #8567

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This is my first PR so feel free to provide feedback or request modifications both in the code and the PR itself! 😄

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -346,7 +346,11 @@ export function stateMixin (Vue: Class<Component>) {
     options.user = true
     const watcher = new Watcher(vm, expOrFn, cb, options)
     if (options.immediate) {
-      cb.call(vm, watcher.value)
+      try {
+        cb.call(vm, watcher.value)
+      } catch (error) {
+        handleError(error, vm, `callback for immediate watcher "${watcher.expression}"`)
+      }
     }
     return function unwatchFn () {
       watcher.teardown()
--- a/test/unit/features/error-handling.spec.js
+++ b/test/unit/features/error-handling.spec.js
@@ -92,6 +92,16 @@ describe('Error handling', () => {
     }).then(done)
   })
 
+  it('should recover from errors in user immediate watcher callback', done => {
+    const vm = createTestInstance(components.userImmediateWatcherCallback)
+    waitForUpdate(() => {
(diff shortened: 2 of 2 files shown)


src/core/instance/state.js
```diff
@@ -346,7 +346,11 @@ export function stateMixin (Vue: Class<Component>) {
     options.user = true
     const watcher = new Watcher(vm, expOrFn, cb, options)
     if (options.immediate) {
-      cb.call(vm, watcher.value)
+      try {
```
posva: Should we call `watcher.getAndInvoke` @yyx990803 ?

PR #8586, opened 2018-08-01
feat: Store functional render context on vnode in development

znck:
Used in https://github.com/vuejs/vue-devtools/pull/719 to display prop values in component inspector.

### DIFF
--- a/src/core/vdom/create-functional-component.js
+++ b/src/core/vdom/create-functional-component.js
@@ -124,7 +124,7 @@ function cloneAndMarkFunctionalResult (vnode, data, contextVm, options, renderCo
   clone.fnContext = contextVm
   clone.fnOptions = options
   if (process.env.NODE_ENV !== 'production') {
-    clone.devtoolsMeta = renderContext
+    ;(clone.devtoolsMeta = clone.devtoolsMeta || {}).renderContext = renderContext
   }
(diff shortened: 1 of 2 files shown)


src/core/vdom/create-functional-component.js
```diff
@@ -105,24 +105,27 @@ export function createFunctionalComponent (
...
-function cloneAndMarkFunctionalResult (vnode, data, contextVm, options) {
+function cloneAndMarkFunctionalResult (vnode, data, contextVm, options, renderContext) {
   // #7817 clone node before setting fnContext, otherwise if the node is reused
   // (e.g. it was from a cached normal slot) the fnContext causes named slots
   // that should not be matched to match.
   const clone = cloneVNode(vnode)
   clone.fnContext = contextVm
   clone.fnOptions = options
+  if (process.env.NODE_ENV !== 'production') {
+    clone.devtoolsMeta = renderContext
```
chrisvfritz: Should we make `devtoolsMeta` an object with a `renderContext` property instead? For example:

```js
clone.devtoolsMeta = { renderContext }
```

I'm thinking it might be easier to add other information to `devtoolsMeta` later this way.

chrisvfritz:
Should we also add a test for this?

PR #8613, opened 2018-08-07
fix(chrome): workaround background-clip style attribute

### LINKED ISSUE #8612: When processing the style attribute，Vue.js miss the -webkit profix
### Version
2.5.17

### Reproduction link
[https://jsfiddle.net/k1uptbpw/7/](https://jsfiddle.net/k1uptbpw/7/)

### Steps to reproduce
open the jsfiddle link，and use devtool to inspect the element

### What is expected?
the style on the div element should be "-webkit-background-clip: text;"

the style on the div element is "background-clip: text;"

---
This is a Chrome's bug（my version is 67）, and it will be ok when using "style.cssText" to set the style. And according to Can I Use, when use this property, it must with the prefix "-webkit", so I think there is two way to resolve this:
1.(ugly)change vue's "setProp" method
2.add some case in the official document

posva:
Closes #8612

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/platforms/web/runtime/modules/style.js
+++ b/src/platforms/web/runtime/modules/style.js
@@ -9,7 +9,7 @@ const setProp = (el, name, val) => {
   /* istanbul ignore if */
   if (cssVarRE.test(name)) {
     el.style.setProperty(name, val)
-  } else if (name === '-webkit-background-clip') {
+  } else if (name === '-webkit-background-clip' && val === 'text') {
     // #8612 probably a webkit bug in Chrome 67
     el.setAttribute('style', (el.getAttribute('style') || '') + `${name}: ${val};`)
   } else if (importantRE.test(val)) {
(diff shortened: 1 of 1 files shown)


src/platforms/web/runtime/modules/style.js
```diff
@@ -9,6 +9,9 @@ const setProp = (el, name, val) => {
   /* istanbul ignore if */
   if (cssVarRE.test(name)) {
     el.style.setProperty(name, val)
+  } else if (name === '-webkit-background-clip') {
```
Justineo: Should we add `&& val === 'text'` too? It only fails on this experimental value.

yyx990803:
https://github.com/vuejs/vue/issues/8612#issuecomment-432408604

PR #8630, opened 2018-08-09
(fix #8337) test(component-slot): new child vnode gets destroyed instead of old one

### LINKED ISSUE #8337: Nested component's v-model doesn't trigger its updated lifecycle method when a parent dynamic component is updated by provide/inject
### Version
2.5.16

### Reproduction link
[https://jsfiddle.net/linnea/zmkt7ajy/15/](https://jsfiddle.net/linnea/zmkt7ajy/15/)

### Steps to reproduce
1. Click on the working checkbox. The v-model value updates and the "updated" lifecycle event fires a console log.
2. Click on the "busted" checkbox, which is the same markup as the working checkbox, but wrapped within a div inside the DynamicWrap component. 

Notice that the checkbox appears checked, but the v-model doesn't update and the "updated" lifecycle event of the checkbox never fires.

### What is expected?
The second checkbox's v-model updates correctly

No updated lifecycle event fires, the v-model is not updated.

---
What's happening:
The DynamicWrapper is rerendered after mount when the UnwrappedCheckbox is created inside of its default slot. UnwrappedCheckbox uses provide/inject to update the DynamicWrap component to wrap it with a 'div' instead of a 'label' element. This clones the content inside, but once the Checkbox is nested a layer deeper, it stops behaving as expected.

Sorry for the monster title and complex replication, let me know if there's anything more I can do to clarify. For context, the purpose of this DynamicWrapper is to act as a type of "FormItem" component. It wraps any element placed inside of the component (In this case a Checkbox) with a label element unless it's something like a Checkbox or Radio. We have our own Checkbox and Radio, that we'd like to tell the FormItem to unwrap the label if a Checkbox is placed inside it.

I've also tested in the beta prerelease and the issue remains.

Possible related previous issues:
https://github.com/vuejs/vue/issues/6372
https://github.com/vuejs/vue/issues/6915

hirokiosame:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
I'm not confident in the quality of the fix (whether it's addressing the bug directly or indirectly) but deep-cloning the vnode in `createElm` when a vnode used in a previous render is encountered seems to fix the bug.

In any case, a test has been added to repro the bug.

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -138,6 +138,9 @@ export function createPatchFunction (backend) {
       // reference node. Instead, we clone the node on-demand before creating
       // associated DOM element for it.
       vnode = ownerArray[index] = cloneVNode(vnode)
+      if (vnode.children) {
+        vnode.children = vnode.children.map(cloneVNode)
+      }
     }
 
     vnode.isRootInsert = !nested // for transition enter check
--- a/test/unit/features/component/component-slot.spec.js
+++ b/test/unit/features/component/component-slot.spec.js
@@ -886,4 +886,47 @@ describe('Component slot', () => {
       expect(vm.$el.textContent).toBe('foo')
     }).then(done)
   })
+
+  // #8337
+  it('should not destroy the new child vm', done => {
+    const Parent = {
+      render (h) {
+        return h(this.API.tag, this.$slots.default)
+      },
+      data () {
+        return { API: { tag: 'p' }}
+      },
+      provide () {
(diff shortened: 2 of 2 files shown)


hirokiosame:
Not sure what's up with the failed lint-flow-type test since I didn't touch the file it's throwing an error at

PR #8644, opened 2018-08-12
fix(types): scoped slots can return a single VNode

KaelWD:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No
- [x] Maybe, but it was wrong to start with


yyx990803:
I think it actually makes more sense to update the implementation to match the types so that it always return arrays. That should also make code more consistent.

Related: https://github.com/vuejs/vue/issues/8056

KaelWD:
I forgot I'd already made this, ended up doing it again in #9131. I agree the implementation could be changed, but until then this change is needed to ensure the types reflect the *current* implementation.

Issue #273, opened 2014-05-16
Why not using HTML5 data-* attributes


SeriousM:
Hi, is there a good reason why you don't go with the [`data-*`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#data-*) attributes and invent your own standard with `v-*`?

From https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes:

> In addition, search crawlers do not index data attributes' values.

MinonHeart:
It casue an error when I use use ```Vue.config({ prefix: 'data'})``` to overwrite ```v-*``` (Vue 1.0.16 csp), and I found noting related to this in document.

yyx990803:
@MinonHeart you are looking at an old issue and that config option has been deprecated long time ago. The short answer is you can no longer configure the prefix, because `v-` attributes are removed when rendered anyway.

xMort:
I would welcome this feature. Any attribute that is not in HTML specification or starts with _data-*_ prefix is not valid. Your IDE will also complain about this and I don't like warnings in my code :(

Here is simple demo in the latest IntelliJ IDEA
![image](https://cloud.githubusercontent.com/assets/1880148/14915831/1de26a84-0e16-11e6-8271-99e8ae069066.png)

posva:
@xMort Actually it's valid HTML but there're a lot of IDEs and editors that does not respect the spec. I couldn't find the spec so I couldn't share it 😢 , sorry

alanondra:
>Actually it's valid HTML but there're a lot of IDEs and editors that does not respect the spec.

No, it's not. All custom attributes not belonging to a declared XML namespace must be prefixed with `data-` and must be `[a-z]+(-[a-z]+)*`.

FibreFoX:
Am i right, that these specs are the ones talked about?
XML-Spec: https://www.w3.org/TR/2011/WD-html5-20110525/infrastructure.html#xml-compatible
HTML-Spec: https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#embedding-custom-non-visible-data-with-the-data-attributes

yyx990803:
Note the difference between **valid HTML syntax** vs **fully-valid HTML**. The former means it adheres to the syntax constraints laid out in the spec, and only the syntax; the latter entails following the spec to every extreme detail, e.g. if an element doesn't have a certain attribute then you cannot use it on that element.

[...]

PR #8663, opened 2018-08-16
fix(parser): add data-v- prefix to make directive comply with WCAG2 A

bowen31337:
Add data-v- prefix for the directive to make the site comply with WCAG2 level A standard

Attribute with v- prefix is regarded as invalid attribute in https://validator.w3.org/, that results
in the accessibility error for not complying with WCAG2 A. In the western countries, the public
sites built with Vuejs in such way have the risks of being sued for not making the sites accessible.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Reference: https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-parses.html

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -19,14 +19,14 @@ import {
   pluckModuleFunction
 } from '../helpers'
 
-export const onRE = /^@|^v-on:/
-export const dirRE = /^v-|^@|^:/
+export const onRE = /^@|^v-on:|^data-v-on:/
+export const dirRE = /^v-|^data-v-|^@|^:/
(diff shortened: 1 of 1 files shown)


posva:
Thanks for your PR 🙂 
Your resulting HTML is wc3 compliant, and that's what matters. See #273

bowen31337:
It does not hurt to add “data-“ prefix, it gives devs the option to choose the directive. If the framework does not provide this option, then it is the framework to blame.

It is undeniable that the directive will be removed when the rendered tree is complete, BUT BUT BUT It is really unaccessible when the JavaScript of the page is disabled.

PR #8666, opened 2018-08-17
feat(compiler): allow unicode characters for component name as described in #8564

### LINKED ISSUE #8564: Invalid component name error for valid name that uses unicode characters
### Version
2.5.16

### Reproduction link
[https://jsfiddle.net/50wL7mdz/536596/](https://jsfiddle.net/50wL7mdz/536596/)

### Steps to reproduce
1. register component with unicode component name. ex: `my-컴포넌트`
2. use it in template. ex: `<my-컴포넌트></my-컴포넌트>`
3. Vue warn is logged and component is not rendered.

### What is expected?
No error, no warning. The component should be rendered properly.

The component is not rendered with this warning.

> [Vue warn]: Invalid component name: "my-컴포넌트". Component names can only contain alphanumeric characters and the hyphen, and must start with a letter.

---
[Component-Names in Vue Guide](https://vuejs.org/v2/guide/components-registration.html#Component-Names) says,

> we strongly recommend following the W3C rules for custom tag names (all-lowercase, must contain a hyphen)

However, [W3C rules](https://www.w3.org/TR/html53/semantics-scripting.html#potentialcustomelementname) allow lots of unicode characters for component name. I think these two rules are important:

- They start with a lowercase ASCII letter, ensuring that the HTML parser will treat them as tags instead of as text.
- They do not contain any uppercase ASCII letters, so user agents can always treat HTML elements ASCII-case-insensitively.

The rules prevent using uppercase ASCII letters, and does not force using only lowercase ASCII letters. They allow unicode characters, but Vue.js doesn't accept unicode component names.

Furthermore, the error level is not `error` but `warn`, but it doesn't work in some cases like using it as custom element in template. It works when routed by vue router.

youngrok:
resolve #8564

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


youngrok:
In addition to #8564 , I added support for unicode property path in watcher.

src/core/util/options.js
```diff
@@ -253,11 +254,10 @@ function checkComponents (options: Object) {
 }
 
 export function validateComponentName (name: string) {
-  if (!/^[a-zA-Z][\w-]*$/.test(name)) {
+  if (!new RegExp(`^[a-zA-Z][\\-\\.0-9_a-zA-Z${unicodeLetters}]*$`).test(name)) {
```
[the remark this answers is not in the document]
Justineo: And shall we also allow component name start with characters apart from `a-zA-Z`? Thoughts? /ping @yyx990803

src/core/util/lang.js
```diff
@@ -1,5 +1,12 @@
 /* @flow */
 
+/**
+ * unicode letters used for parsing html tags, component names and property paths.
+ * use https://www.w3.org/TR/html53/semantics-scripting.html#potentialcustomelementname
+ * except \u10000-\uEFFFF because of performance problem
```
Justineo: I made a benchmark for this and it turns out that in most modern browsers except Safari, it runs fastest if we include `\u10000-\uEFFFF`:

https://jsperf.com/unicode-regex-test

src/core/util/options.js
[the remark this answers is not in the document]
Justineo: Another problem is case conversion. We use to convert component/prop names between kebab-case with PascalCase/camelCase. If we expand the character set to a broaden extent, something like `'按钮-组件'` will become `'按钮组件'` and lose a part of the original semantics.

src/core/util/lang.js
youngrok: Well, in my computer, `npm test` failed because of timeout when including `\u10000-\uEFFFF`. It passed without the extra characters. I don't know why the performance difference exists, yet.

src/compiler/parser/html-parser.js
```diff
@@ -11,13 +11,11 @@
...
 import { isNonPhrasingTag } from 'web/compiler/util'
+import { unicodeLetters } from '../../core/util/lang'
 
 // Regular Expressions for parsing tags and attributes
 const attribute = /^\s*([^\s"'<>\/=]+)(?:\s*(=)\s*(?:"([^"]*)"+|'([^']*)'+|([^\s"'=<>`]+)))?/
-// use https://www.w3.org/TR/html53/semantics-scripting.html#potentialcustomelementname
-// except \u10000-\uEFFFF because of performance problem
-export const pcenchars = '[\\-\\.0-9_a-zA-Z\\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]'
-const ncname = `[a-zA-Z_]${pcenchars}*`
+const ncname = `[a-zA-Z_][\\-\\.0-9_a-zA-Z${unicodeLetters}]*`
```
cutPicturesMan, line 18: 'a-zA-Z' is repeat in unicodeLetters

Issue #8728, opened 2018-08-29
An anti-pattern in computed property may cause performance issue


Yuyz0112:
### Version
2.5.17

### Reproduction link
[https://codesandbox.io/s/140202wlq](https://codesandbox.io/s/140202wlq)

### Steps to reproduce
Open the code sandbox link and click the button to use the computed property 'filteredEntities', and you will find it cost a lot of time.

After check the call tree, I found this is because the code calls the getter of 'entities'  many times which will then call the `dependArray` function.

The code leading to this is:
```js
const len = this.entities.length;
for (let i = 0; i < len; i++) {
  const e = this.entities[i];
  // do something with e
}
```
If this.entities has a length of n, its getter will be called n + 1 times. And every time when the getter was called, it will call dependArray on the value of this.entities, which is the array with length n.
Since dependArray will iterate the value, the depend function will be called **(n + 1)^2** times totally.

I found this code when I'm reviewing someone's PR, and the original code use `for (let i = 0; i < this.entities.length; i++) ` and make another n times call to the getter.

It was easy to avoid this problem, such as using some array methods like `filter` to only call getter once.
Even cache the value of entities by `const cache = this.entities` can solve it.

Although it was easy to avoid of this pattern, I still think it is dangerous because the original code was not an obviously anti-pattern.
And this will not show any performance issue when the array is small, but may cause serious performance issue when the array is large in some production environment.

### What is expected?
I've read the related source code and I think this is the expected result of the observer system.

Then I read some chapter of the guide again to ensure there is no mention about this, so I'm not sure whether we need to add this as a NOT TODO in the computed property chapter.

### What is actually happening?
Described.

plainnany:
Is this issue going to be fixed in vue version 2.+ ？

PR #8734, opened 2018-08-31
fix(#8728): only execute the dependArray function once

HcySunYang:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: This PR is not a bug fix, but a performance improvement for issue #8728.

- [ ] Yes
- [x] No

**Other information:**
The `newDepIds` property of each Watcher instance object holds all the dependencies collected by this evaluation. We can prevent the `dependArray` function from being called multiple times by checking if the dependency has been collected.

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -158,10 +158,11 @@ export function defineReactive (
     configurable: true,
     get: function reactiveGetter () {
       const value = getter ? getter.call(obj) : val
-      if (Dep.target) {
+      const targetWatcher = Dep.target
+      if (targetWatcher) {
         dep.depend()
         if (childOb) {
-          const needDepend = Array.isArray(value) && !Dep.target.checkRelated(childOb.dep)
+          const needDepend = Array.isArray(value) && !targetWatcher.checkRelated(childOb.dep)
           childOb.dep.depend()
           if (needDepend) {
             dependArray(value)
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -161,8 +161,9 @@ export function defineReactive (
       if (Dep.target) {
         dep.depend()
         if (childOb) {
+          const needDepend = Array.isArray(value) && !Dep.target.checkRelated(childOb.dep)
           childOb.dep.depend()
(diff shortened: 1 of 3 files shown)


Yuyz0112:
em.. I think the changes in this PR cause a regression bug.

It looks like the `dependArray` method was designed to solve the dependency problem in a nested array.

```vue
<template>
  <div>
    <p>{{ nest[0] }}</p>
    <button @click="nest[0].push('next')">add one</button>
  </div>
</template>

<script>
export default {
  data() {
    return {
      nest: [
        ['init'],
      ],
    };
  },
}
</script>
```

Have not try this example with your branch yet, but I think the `checkRelated` will cause the nest array not reactive.

HcySunYang:
The `dependArray` function will still execute, except that it will only execute `once`.
`Watcher` collects dependencies by executing getters, and each time the getter is executed, the `watcher.cleanupDeps` function is executed to reset `newDepIds`. That is, everything is a new start every time the watcher's getter is re-executed, but in the process, there is no need to execute the `dependArray` function multiple times. It does not cause nested arrays to be not reactive.

Yuyz0112:
my bad, the check was executed before depend so it is ok:)

wangyi7099 (commented):
How about
```javascript
const needDepend = !Dep.target.checkRelated(dep)
   if (Dep.target && needDepend ) {
        dep.depend()
        if (childOb) {
          childOb.dep.depend()
          if (Array.isArray(value)) {
        
            dependArray(value)
        }
```
Because  plain-object will still cause useless redepend.

HcySunYang:
@wangyi7099 
When value is a plain-object, your code will cause it to fail to collect dependencies. And performance issues only occur in the case of large arrays, usually we don't need to care about objects

wangyi7099:
@HcySunYang 
I'm sorry. My sample code has some errors. I  have updated it.

>  And performance issues only occur in the case of large arrays, usually we don't need to care about objects

I think you are right. But I also think case about plain-object will also save some performance cost.

PR #8735, opened 2018-08-31
fix (#8731): drop enumerated attrs

### LINKED ISSUE #8731: Drop special logic for “enumeratedAttrs”
### Version
2.5.17

### Reproduction link
[http://jsfiddle.net/Justineo/upwrqm0z/](http://jsfiddle.net/Justineo/upwrqm0z/)

### Steps to reproduce
Bind `'plaintext-only'` or an unsupported value to `contenteditable` attribute.

### What is expected?
Rendering `contenteditable="plaintext-only"` for the element.

`contenteditable="true"` is rendered.

---
For reference:

| `:contenteditable` | `contentEditable` in Vue | `contentEditable` in HTML ||
| -- | -- | -- | -- |
| `""` | `'true'` | `'true'` ||
| `"false"` | `'false'` | `'false'` ||
| `"true"` | `'true'` | `'true'` ||
| `"plaintext-only"` | `'true'` | `'plaintext-only'` | ❗️ |
| `"foo"` | `'true'` | `'inherit'` | ❗️ |
| `null` | `'false'` |||
| `undefined` | `'inherit'` |||
| `false` | `'false'` |||
| `true` | `'true'` |||

Vue currently treat `enumeratedAttrs` as enumerated attributes with only `'true'`/`'false'` values (boolean-like enumerated attributes) and this made them behave like boolean attribtues. While actually these attributes may have more than two states and are extensible for other values. For example, in HTML, `contenteditable`'s corresponding IDL attribute `contentEditable` resolves to `'inherit'` instead of `'false'` when `contenteditable` is not  present. And the latest W3C draft has also defined more possible values for it (`'plaintext-only'` is already accepted in Chrome) so we have to remove this logic, otherwise developers can't leverage these new features.

And if we drop the `enumeratedAttrs` thing and treat them like other ordinary attributes, there will be some influence on the following cases:

| `v-bind` | before | after |
| -- | -- | -- |
| `null` | `'false'` | / |
| `false` | `'false'` | / |

I'm not sure if we should regard this as breaking because we never documented this usage that binding `null`/`false` to `contenteditable`/`draggable`/`spellckeck` will be resolved as `'false'`.

Justineo:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No
- [x] Not sure

As explained in #8731.

**Other information:**

### DIFF
--- a/src/platforms/web/runtime/modules/attrs.js
+++ b/src/platforms/web/runtime/modules/attrs.js
@@ -13,7 +13,6 @@ import {
   xlinkNS,
   getXlinkProp,
   isBooleanAttr,
-  isEnumeratedAttr,
   isFalsyAttrValue
 } from 'web/util/index'
 
@@ -51,7 +50,7 @@ function updateAttrs (oldVnode: VNodeWithData, vnode: VNodeWithData) {
     if (isUndef(attrs[key])) {
       if (isXlink(key)) {
         elm.removeAttributeNS(xlinkNS, getXlinkProp(key))
-      } else if (!isEnumeratedAttr(key)) {
+      } else {
         elm.removeAttribute(key)
       }
     }
@@ -74,8 +73,6 @@ function setAttr (el: Element, key: string, value: any) {
         : key
       el.setAttribute(key, value)
     }
-  } else if (isEnumeratedAttr(key)) {
-    el.setAttribute(key, isFalsyAttrValue(value) || value === 'false' ? 'false' : 'true')
   } else if (isXlink(key)) {
     if (isFalsyAttrValue(value)) {
       el.removeAttributeNS(xlinkNS, getXlinkProp(key))
--- a/src/platforms/web/server/modules/attrs.js
+++ b/src/platforms/web/server/modules/attrs.js
@@ -10,7 +10,6 @@ import {
 
 import {
   isBooleanAttr,
-  isEnumeratedAttr,
   isFalsyAttrValue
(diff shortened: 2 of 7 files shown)


yyx990803:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable

The logic is still needed to coerce `true` and `false` to `"true"` and `"false"` - we just need to allow additional values.

Justineo:
Then maybe we should coerce all boolean bindings for non-boolean attributes to string. There are other attributes like `aria-hidden` which accept `'true'`/`'false'` values. Now it seems that we are coercing `true` but not `false`. Is it better to apply this coercing to all normal attributes? There seems no need to have special logic for enumerated attributes. @yyx990803

yyx990803:
Hmm that's true... however, dropping the logic altogether would technically be a breaking change, as code that previous produced valid usage would now require manual coercing from the user.

So I think for 2.x it's better to just loosen the coercing to skip non-boolean values so that other values for `contenteditable` can be used.

yyx990803:
I think for v2 we need to keep as much backwards compat as possible... in e632e9a0 I've added special treatment for `contenteditable` for now. Most aria-x attributes have explicit treatment of non-listed values mandated in the spec so it's less of a problem... In v3 we will not do any coercing.

PR #8755, opened 2018-09-04
fix(component): clean up memory leak after loading async component completes (fix #8740)

### LINKED ISSUE #8740: Memory Leak When Using Async Components
### Version
2.5.17

### Reproduction link
[https://codesandbox.io/s/4k43rylw0](https://codesandbox.io/s/4k43rylw0)

### Steps to reproduce
1. Load the reproducer and take a memory heap snapshot. Searching the heap you will find 4 VueComponents (App, Container, Child, Async Child)
image

2. Click the Destroy button. This toggles a v-if which will remove Child component and thus AsyncChild component as well.
3. At this point, the page is updated as expected
4. Take another memory heap snapshot. Search the heap again for "VueComponent" and you will see that 4 VueComponents still exist
image
image

### What is expected?
After destroying Child component, the component and it's asynchronously loaded child component should be destroyed and the memory freed up.

Vue is holding on to a context reference which is preventing these components from being garbage collected

---
This is happening due to the resolve-async-component logic located here: https://github.com/vuejs/vue/blob/52719ccab8fccffbdf497b96d3731dc86f04c1ce/src/core/vdom/helpers/resolve-async-component.js#L57-L68

In the forceRender function on line 64, after calling $forceUpdate, the contexts array should be cleared to free up those references so that the components will be able to be garbage collected at a later point in time

mattlavallee:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/vdom/helpers/resolve-async-component.js
+++ b/src/core/vdom/helpers/resolve-async-component.js
@@ -64,6 +64,14 @@ export function resolveAsyncComponent (
     const forceRender = () => {
       for (let i = 0, l = contexts.length; i < l; i++) {
         contexts[i].$forceUpdate()
+
+        const currContext = contexts[i]
+        const contextIdx = contexts.indexOf(currContext)
+        if (contextIdx >= 0) {
+          currContext.$nextTick(() => {
+            contexts.splice(contextIdx, 1)
+          })
+        }
       }
     }
 
--- a/src/core/vdom/helpers/resolve-async-component.js
+++ b/src/core/vdom/helpers/resolve-async-component.js
@@ -61,17 +61,18 @@ export function resolveAsyncComponent (
     const contexts = factory.contexts = [context]
     let sync = true
 
-    const forceRender = () => {
+    const forceRender = (renderCompleted: boolean) => {
       for (let i = 0, l = contexts.length; i < l; i++) {
         contexts[i].$forceUpdate()
 
-        const currContext = contexts[i]
-        const contextIdx = contexts.indexOf(currContext)
-        if (contextIdx >= 0) {
-          currContext.$nextTick(() => {
-            contexts.splice(contextIdx, 1)
-          })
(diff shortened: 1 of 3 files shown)


test/unit/features/transition/transition-mode.spec.js
```diff
@@ -457,7 +457,7 @@ if (!isIE9) {
             setTimeout(() => {
               resolve({ template: '<div><h1>component B</h1></div>' })
               Vue.nextTick(next)
-            }, (duration + buffer) * 1.5)
+            }, (duration + buffer) * 1.7)
```
mattlavallee: I was noticing timing issues while running the vue tests on my laptop locally. This test would pass if part of the entire suite, but if it was marked as a focused test, it would fail due to timing issues. I encountered this with the code prior to my change as well

src/core/vdom/helpers/resolve-async-component.js
```diff
@@ -61,9 +61,21 @@ export function resolveAsyncComponent (
...
+    const forceRender = (renderCompleted: boolean) => {
       for (let i = 0, l = contexts.length; i < l; i++) {
         contexts[i].$forceUpdate()
+
+        if (!renderCompleted) {
+          continue
+        }
+
+        const currContext = contexts[i]
+        const contextIdx = contexts.indexOf(currContext)
```
wa3l: isn't `contextInx === i`?

mattlavallee: @wa3l - good catch. The correct value of `i` is preserved when used in `$nextTick` if I just assign it to `contextIdx`.
I have updated the diff to include this refactor

beyond8512:
是是是

PR #8901, opened 2018-10-07
refactor: remove some elses

omarkdev:
This pull request has a intention to remove some useless else's to simplify the code.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/sfc/parser.js
+++ b/src/sfc/parser.js
@@ -98,13 +98,13 @@ export function parseComponent (
   function padContent (block: SFCBlock, pad: true | "line" | "space") {
     if (pad === 'space') {
       return content.slice(0, block.start).replace(replaceRE, ' ')
-    } else {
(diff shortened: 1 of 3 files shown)


visualfanatic:
Why only those three files? There are a lot of files that have redundant `else` block.

src/sfc/parser.js
```diff
@@ -98,13 +98,13 @@ export function parseComponent (
   function padContent (block: SFCBlock, pad: true | "line" | "space") {
     if (pad === 'space') {
       return content.slice(0, block.start).replace(replaceRE, ' ')
-    } else {
```
dsonet: I don't really think it's a good idea to remove the else condition here, since it's much much clearer to read & understand.

PR #8925, opened 2018-10-10
fix #8862: cannot watch unicode properties like 'a.中文'

### LINKED ISSUE #8862: cannot watch unicode properties like  'a.中文'
### Version
2.5.17

### Reproduction link
https://jsfiddle.net/k72f3Lon/

### Steps to reproduce
1.webpack project with vue .
2.add a component .
3.create instance of Vue .
4.add a property `data:function(){return {a:{中文:"",b:""}}}` to the instance.
5.add watch properties to the instance as `watch:{'a.b':function(val){},'a.中文':function(val){}}`
6.console show error that Watcher only accepts simple dot-delimited paths. 

### What is expected?
no error the code run continue

the code seems stopped to run

derek-li:
Hi all at Vue!

This pull request resolves the issue regarding watchers not watching unicode properties like 'a.中文'.

My method for resolving this case is to throw a warning when non-alphanumeric, $, or _ characters are detected. My reasoning behind throwing a warning instead of supporting special characters is that I think  supporting special characters would promote poor naming habits that could only make code more difficult to read and understand.

I'm pretty new to open source so any feedback would be awesome and please let me know if you'd like to see any changes or improvements!

Link to issue: https://github.com/vuejs/vue/issues/8862

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/core/observer/watcher.js
+++ b/src/core/observer/watcher.js
@@ -80,13 +80,21 @@ export default class Watcher {
     } else {
       this.getter = parsePath(expOrFn)
       if (!this.getter) {
+        if (this.getter === false) {
+          process.env.NODE_ENV !== 'production' && warn(
+            `Failed watching path: "${expOrFn}" ` +
+            'Watcher only accepts simple dot-delimited paths. ' +
+            'For full control, use a function instead.',
+            vm
+          )
+        } else {
+          process.env.NODE_ENV !== 'production' && warn(
+            `Failed watching path: "${expOrFn}" ` +
+            'Vue watchers only accept alphanumeric, $, or underscore naming of properties.',
+            vm
+          )
+        }
         this.getter = function () {}
-        process.env.NODE_ENV !== 'production' && warn(
-          `Failed watching path: "${expOrFn}" ` +
-          'Watcher only accepts simple dot-delimited paths. ' +
-          'For full control, use a function instead.',
-          vm
-        )
       }
     }
     this.value = this.lazy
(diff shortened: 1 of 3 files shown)


posva:
This is not fixing the issue it says to be fixing 🤔

src/core/observer/watcher.js
```diff
@@ -80,13 +80,21 @@ export default class Watcher {
     } else {
       this.getter = parsePath(expOrFn)
       if (!this.getter) {
+        if (this.getter === false) {
```
yyx990803: I don't like that we are relying on magic return values (`false`) here, as the logic is separated in two places. Also the check is warning only so it's better to place it right here inside a `if (process.env.NODE_ENV !== 'production') { ... }` block, using an inline regexp.

yyx990803:
Note this may become invalid if we merge #8666, I'm still debating whether we should support it.

src/core/observer/watcher.js
derek-li: @yyx990803 Thanks for taking the time to review this PR. 

I've gone ahead and made the commits to stop relying on random return values. I also tried your suggestion with having the check in an `if (process.env.NODE_ENV !== 'production') { ... }` block, but felt that it was unclear to have two regex checks in different places (`watcher` and `parsePath`) and did not resolve your comment on the logic being separated, so I grouped them together.

Thanks again and please let me know what you think!

yyx990803:
Closed via #8666

PR #8944, opened 2018-10-13
perf(deps): improve the order to teardown watcher from O(n*m) to O(n) (fix vuex#1383)

ktsn:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: performance improvement

- [ ] Yes
- [x] No

**Other information:**
This PR fixes vuejs/vuex#1383.

Simplified reproduction / benchmark: https://github.com/ktsn/vue-deps-perf-demo
The demo will destroy 1000 components that has a computed property depends on an array with 1000 length. Then print its time.

On my machine, the result is like the following:
current: 3434ms
patched: 224ms

On the current reactive system implementation, when a computed property depends on an array, its children object will also be registered to computed property's watcher as dependency.

When the watcher is torndown, the watcher iterates its dependencies (size `n`). The dependencies remove the watcher from their watcher list (size `m`). So it will cost O(n*m).

So it would become slow if there are many computed properties and a large array.

This PR change the watcher list in `Dep` object to object map to make the order linear. The entire order of the watcher teardown will be O(n).

There may be more overhead on `delete` operator. I'm not sure how it affects actual application though.

### DIFF
--- a/src/core/observer/dep.js
+++ b/src/core/observer/dep.js
@@ -1,7 +1,6 @@
 /* @flow */
 
 import type Watcher from './watcher'
-import { remove } from '../util/index'
 import config from '../config'
 
 let uid = 0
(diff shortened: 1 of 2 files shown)


sqal:
There's aleady similar PR https://github.com/vuejs/vue/pull/8157

PR #8946, opened 2018-10-14
Typings: Improve $slots and $scopedSlots type to prevent unchecked access to undefined

thenickname:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

When working directly with render functions in TypeScript / tsx, currently the following code is unsafe and unchecked by the ts compiler:

```ts
this.$sopedSlots.default() // default can be undefined
```

The change in this PR makes TypeScript complain about the unsafe access accordingly and the code has to be expressed in a way that prevents calling an undefined function, e.g.:

```ts
( this.$scopedSlots.default || ( () => 'default value' ) )( props )
```

### DIFF
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -27,7 +27,7 @@ export interface Vue {
   readonly $root: Vue;
   readonly $children: Vue[];
   readonly $refs: { [key: string]: Vue | Element | Vue[] | Element[] };
-  readonly $slots: { [key: string]: VNode[] };
+  readonly $slots: { [key: string]: VNode[] | undefined };
   readonly $scopedSlots: { [key: string]: ScopedSlot | undefined };
   readonly $isServer: boolean;
   readonly $data: Record<string, any>;
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -28,7 +28,7 @@ export interface Vue {
   readonly $children: Vue[];
   readonly $refs: { [key: string]: Vue | Element | Vue[] | Element[] };
   readonly $slots: { [key: string]: VNode[] };
-  readonly $scopedSlots: { [key: string]: ScopedSlot };
+  readonly $scopedSlots: { [key: string]: ScopedSlot | undefined };
   readonly $isServer: boolean;
   readonly $data: Record<string, any>;
   readonly $props: Record<string, any>;
--- a/types/test/options-test.ts
+++ b/types/test/options-test.ts
@@ -296,8 +296,8 @@ Vue.component('component-with-scoped-slot', {
     child: {
       render (this: Vue, h: CreateElement) {
         return h('div', [
-          this.$scopedSlots['default']({ msg: 'hi' }),
-          this.$scopedSlots['item']({ msg: 'hello' })
+          this.$scopedSlots['default']!({ msg: 'hi' }),
+          this.$scopedSlots['item']!({ msg: 'hello' })
         ])
       }
     }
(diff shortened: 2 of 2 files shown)


ktsn:
Thank you for making this PR. However including `undefined` in the scoped slots value is actually inaccurate. 

In the TypeScript, index signature is not guarantees whether the referenced **key** is exists. This is similar that an array type can be return `undefined` if accessing index of out of length.

```ts
const obj: Record<string, number> = { foo: 1, bar: 2 }
obj.baz // -> undefined

const arr: number[] = [1, 2, 3]
arr[5] // -> undefined
```

Some case that Including `undefined` is appropriate would be the case that **key** is exists but the **value** is `undefined` such as:

```ts
const obj: Record<string, number | undefined> = { foo: 1, bar: undefined }
```

`Object.values()` type would help you to understand these differences:

```ts
const a: Record<string, number | undefined> = { a: 1 }

// will be `(number | undefined)[]` even though `undefined` will be never appeared
const b = Object.values(a) 
```

thenickname:
Hey, thank you for the quick reaction.

> In the TypeScript, index signature is not guarantees whether the referenced key is exists.

Actually, I believe that it effectively does guarantee the following: It tells the TypeScript compiler that every possible key is permitted and (thats the important part) that all the values of the given object are **guaranteed to have actual content of the type `ScopedSlot`**.

Let me give you an example to hopefully better illustrate what is happening with the current type definition:

Let's first look at the current type definition for `$scopedSlots`:

```ts
interface Vue {
}
```

Now, in a vue component we try to call the default slot function:

```ts
this.$scopedSlots.default( props );
```

TypeScript won't complain about this code. But at runtime, `default` slot could be empty. The code above in that case will lead to a runtime error, because `default` is not a function and hence not callable.

Now if we tweak the type definition like that:

```ts
interface Vue {
    readonly $scopedSlots: { [key: string]: ScopedSlot | undefined };
}
```

This is what happens:

bildschirmfoto 2018-10-16 um 00 41 20

It says `cannot invoke an object which is possibly undefined`. And that is a good thing. It forces you to do a safety check before calling the function. For example like that:

```tsx
<div class={styles.root}>
    {this.$scopedSlots.default ?
        this.$scopedSlots.default( props ) :
        'Fallback content'
    }
</div>
```

So, to sum it up:

Right now the TypeScript compiler won't complain about slot access because of the current type defintition for $scopedSlots. Because the content that is referenced by the key is defined to **always be a ScopedSlot**, TypeScript will happily allow a function call even if no slot function is present. It does not account for the fact that at runtime a slot is not necessarily defined.

I hope this explanation better illustrates why I believe this PR is an improvement of the current typing.

types/vue.d.ts
```diff
@@ -28,7 +28,7 @@ export interface Vue {
   readonly $children: Vue[];
   readonly $refs: { [key: string]: Vue | Element | Vue[] | Element[] };
   readonly $slots: { [key: string]: VNode[] };
```
KaelWD: Non-scoped slots can be undefined too.

ktsn:
I see your point. As scoped slot and slot optional, it may be better to have `undefined` value 🙂 

I was thinking scoped slot and slot always there if the object key is exists but it was incorrect as they can be blank. Not sure how actual implementation is but it sounds good to have `undefined` in that case.

ktsn:
I see your point. As scoped slot and slot optional, it may be better to have `undefined` value 🙂 

I was thinking scoped slot and slot always there if the object key is exists but it was incorrect as they can be blank. Not sure how actual implementation is but it sounds good to have `undefined` in that case.

ktsn:
I see your point. As scoped slot and slot optional, it may be better to have `undefined` value 🙂 

I was thinking scoped slot and slot always there if the object key is exists but it was incorrect as they can be blank. Not sure how actual implementation is but it sounds good to have `undefined` in that case.

KaelWD:
You missed the one in `VNodeData`, I've fixed it with #9131

thenickname:
@KaelWD thanks

Issue #6190, opened 2017-07-21
'too much recursion' hit when injecting Vue-based mixins


razvyc:
### Version
2.4.1

### Reproduction link
[https://jsfiddle.net/razvansbsb/2z359v28/](https://jsfiddle.net/razvansbsb/2z359v28/)

### Steps to reproduce
1. Define a component A that uses 2 mixins that each define their own provide properties wrapped in a Vue component, e.g. foo and bar.
2. Define a component B inside A that uses inject to add foo and bar properties

### What is expected?
Both foo and bar should be available in component B.

### What is actually happening?
Get 'too much recursion' exception

posva:
You have to return plain objects in provide: https://jsfiddle.net/posva/nvbyv9wc/
I recommend you taking a look to provide inject: https://vuejs.org/v2/api/#provide-inject

razvyc:
This used to work in 2.3.4. It stopped working in 2.4.0.
The problem occurs when trying to recursively merge the objects in `mergeData`. Can we add some smarts and if you detect a `Vue` object only clone the `data` properties?
Thanks!

jasonsanjose:
@posva what's the meaning behind this note here https://vuejs.org/v2/api/#provide-inject:

>Note: the provide and inject bindings are NOT reactive. This is intentional. However, if you pass down an observed object, properties on that object do remain reactive.

I took this to mean that reactive objects were not the default behavior but could be used if implemented.

posva:
@jasonsanjose It means that if you pass down objects that are being observed the injected properties can become reactive: https://jsfiddle.net/4yycfnfm/

jasonsanjose:
@posva thanks for the clarification. We'll give that a try.

aeharding:
I would also like to chime in that I received this issue trying to upgrade to 2.4.x. The regression from 2.3.x with an unhelpful error message made it really hard to track down.

I was also not doing anything weird, just copy-pasting this mixin from the vue-hackernews-2.0 project:
https://github.com/vuejs/vue-hackernews-2.0/blob/master/src/util/title.js

xiechao1211:
i use the element-ui's el-upload compoent with vee-validate, it happens
wrong info is:Maximum call stack size exceeded at mergeData,so now ,what that i can do.

jasonsanjose:
@posva we're still looking for a solution. If you have time to chime in we would really appreciate it. If not, hopefully someone else here might have some ideas or at least be in the same situation.

[...]

PR #8967, opened 2018-10-18
fix(mergeData): skip recursive call if values are identical

KaelWD:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Fixes https://github.com/vuetifyjs/vuetify/issues/5633

I've been unable to reproduce this in isolation, but have run into it several times in the past. This check should also reduce unnecessary calls to `mergeData`, I noticed thousands of instances of identical objects being merged that didn't cause errors.

### DIFF
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -55,7 +55,11 @@ function mergeData (to: Object, from: ?Object): Object {
     fromVal = from[key]
     if (!hasOwn(to, key)) {
       set(to, key, fromVal)
-    } else if (isPlainObject(toVal) && isPlainObject(fromVal)) {
+    } else if (
+      toVal !== fromVal &&
+      isPlainObject(toVal) &&
+      isPlainObject(fromVal)
+    ) {
       mergeData(toVal, fromVal)
     }
   }


KaelWD:
Yeah, but I don't know where to put `b` and `c` to make that happen. It only causes a stack overflow if `a` has recursive references, in our case it's a provided vue instance. In your case it would just be an unnecessary loop over `a` that doesn't do anything as all the properties obviously exist in both arguments.

posva:
It's worth checking what the actual source of the problem is. There could be another problem somewhere else. That being said, the change is a good idea IMO

KaelWD:
There's a similar issue #6190, but that's trying to merge two *different* vue objects so I don't think this patch will fix it unless it's getting hung up on `$root` or something.

jvbianchi:
Any plans on merging this?

jrvaja:
@KaelWD Hope this will be merged in the up coming sprint.

arpit9295:
Hi!
Wondering if the actual source of the problem was ever discovered?

PR #9016, opened 2018-11-02
 fix(v-model): Clear composing flag on blur. Fixes #8945

### LINKED ISSUE #8945: v-model doesn't work as expected with hebrew on mobile phone
### Version
2.5.17

### Reproduction link
[https://jsfiddle.net/yakovyarmo/kw64ynt8/3/](https://jsfiddle.net/yakovyarmo/kw64ynt8/3/)

### Steps to reproduce
Start typing in the input using hebrew language on mobile device.

### What is expected?
The data should be update on every input.

The data is updated some of the time and cause inconsistency.

---
I suspect that the issue occur becuse for some reason the browser treat hebrew with IME.
There has to be a way around it. maybe by a modifier that will trigger the update for native input.

tmorehouse:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

In some instances, the `.composing` flag can be left set as `true` if the user moves focus while they are in the middle of composition. See issue #8945

This fix adds a `blur` handler that sets the `e.target.composing` to `false` if it is still set to `true` when the blur happens.

### DIFF
--- a/src/platforms/web/runtime/directives/model.js
+++ b/src/platforms/web/runtime/directives/model.js
@@ -41,8 +41,8 @@ const directive = {
         // this also fixes the issue where some browsers e.g. iOS Chrome
         // fires "change" instead of "input" on autocomplete.
         el.addEventListener('change', onCompositionEnd)
-        // If user moves focus before composition ends, clear the composing flag
-        // See #
+        // If user moves focus before composition ends, clear the composing flag.
+        // See #8945
         el.addEventListener('blur', onBlur)
         /* istanbul ignore if */
         if (isIE9) {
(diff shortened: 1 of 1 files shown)


tmorehouse:
I'm wondering if a `change` event should also be triggered on `blur` if the composing flag was still set to `true`?

tmorehouse:
Actually from some testing, it appears that `focusout` fires after `blur` (which fires after `change`).

yyx990803:
thanks for the PR but I don't think this is relevant to #8945 or needed - the behavior described in #8945 is the intended behavior, and when the user blurs an input box during composition, it already triggers a `compositionend` event.

tmorehouse:
Ah, yeah, that is true.

Issue #9038, opened 2018-11-07
Throw error when using interpolation instead of JS expression for slot names


revolter:
### What problem does this feature solve?
For this demo:

[https://jsfiddle.net/2os9jken/](https://jsfiddle.net/2os9jken/)

opening the console and running it throws the `Interpolation inside attributes has been removed. Use v-bind or the colon shorthand instead.` error for the `span`, but it should also throw it for the `slot`, as it has the exact same issue for its `name` attribute.

### What does the proposed API look like?
N/A

revolter:
Well, old references still exist, and it's not that obvious for new users. It surely confused me in #9037.

IhsenBen:
can I work on this?

DaZuiZui:
It is not supported to use interpolation {{}} to assign values to attributes, so use the v-bind command (or abbreviated ":") to specify attributes.

tysoncung:
I noticed this issue. Have you tried checking the error logs or console output? That might help narrow down the root cause. I'd be happy to help investigate if you can share more details about your environment (OS, version, etc.).

zhanglinqian:
I'd like to work on this issue.

PR #9050, opened 2018-11-10
feat(parser): throw error when using interpolation instead of expresssion (#9038)

kamilchlebek:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Yes it's already an open issue with `good first issue` label. Resolves: #9038

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -454,6 +454,18 @@ function processOnce (el) {
 function processSlot (el) {
   if (el.tag === 'slot') {
     el.slotName = getBindingAttr(el, 'name')
+    // Fixes: #9038
+    if(process.env.NODE_ENV !== 'production'){
+      const slotName = el.attrsMap['name']
+      if(!dirRE.test(slotName) && parseText(slotName, delimiters)){
+        warn(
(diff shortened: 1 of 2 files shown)


src/compiler/parser/index.js
```diff
@@ -454,6 +454,18 @@ function processOnce (el) {
 function processSlot (el) {
   if (el.tag === 'slot') {
     el.slotName = getBindingAttr(el, 'name')
+    // Fixes: #9038
+    if(process.env.NODE_ENV !== 'production'){
+      const slotName = el.attrsMap['name']
+      if(!dirRE.test(slotName) && parseText(slotName, delimiters)){
```
yyx990803: Why `dirRE.test` here? Seems unnecessary.

kamilchlebek: I've added this to keep the same behavior as in place where previous warning exists (line 596). As you can see that previous warning is shown only when `dirRE.test` is false.

`processSlot` is executed outside of `processAttrs` thus I've added this additional check,  to prevent double warnings

PR #9084, opened 2018-11-18
Fixes some long sentences to short sentences, and make document more formal

cmpt376Kor:
- "an MIT-Open" -> "a MIT"
- breaks some long sentences into several short sentences
- changes some informal language into formal language

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Vue.js is a MIT-licensed open source project. Its ongoing development made possi
 
 #### What's the difference between Patreon and OpenCollective?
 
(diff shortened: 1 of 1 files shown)


awulkan:
I think you unfortunately messed up the grammar in a few places.

> Funds donated via OpenCollective manages with transparent expenses.

Makes no sense.

posva:
Thanks for your interest. Unfortunately, awulken is right. This pr introduces new mistakes. Let's keep the Readme the way it it 😀
PS: it's an MIT

PR #9095, opened 2018-11-20
fix #9046 disable dep collection when invoking lifecycle hooks will be useless when invoking computed in lifecycle

### LINKED ISSUE #9046: Unexpected update on app.vue caused by calling computed on mounted
### Version
2.5.17

### Reproduction link
[https://jsfiddle.net/je3odvk8/](https://jsfiddle.net/je3odvk8/)

### Steps to reproduce
Try click "parent button" which update the parent component's data, but I don't understand why this causes app to update. But if you remove the line #32 console.log(this.childName), this won't happened.

### What is expected?
Expected only updating the Parent component and Child component when data updated on the Parent.

App updated caused by a child component when having console logging the computed data

ysj16:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/observer/dep.js
+++ b/src/core/observer/dep.js
@@ -56,7 +56,7 @@ Dep.target = null
 const targetStack = []
 
 export function pushTarget (_target: ?Watcher) {
-  if (Dep.target) targetStack.push(Dep.target)
+  if (Dep.target !== null) targetStack.push(Dep.target)
   Dep.target = _target
 }
 


ysj16:
In lifecycle.js lines 316, the method 'callHook', disable dep collection when invoking lifecycle hooks by 'pushTarget()', which means the dep collection will be useless in the lifecycle method. The principle of this is that when pushTarget(), the params--watcher is 'undefined', so in the lifecycle method--mounted, Dep.target is 'undefined', which means will not run dep collection.
But I've found a BUG in this method， when invoking computed in mounted，it will pushTarget a lazyWatcher and then popTarget, during the process, it will lose pre Dep.target, which should be 'undefined' and make it to be the renderWatcher, so will run dep collection and collect the child component‘s property by renderWatcher.

ysj16:
In lifecycle.js [lines316](https://github.com/vuejs/vue/blob/dev/src/core/instance/lifecycle.js#L316)  , the method 'callHook', disable dep collection when invoking lifecycle hooks by 'pushTarget()', which means the dep collection will be useless in the lifecycle method. The principle of this is that when pushTarget(), the params--watcher is 'undefined', so in the lifecycle method--mounted, Dep.target is 'undefined', which means will not run dep collection.
But I've found a BUG in this method， when invoking computed in mounted，it will pushTarget a lazyWatcher and then popTarget, during the process, it will lose pre Dep.target, which should be 'undefined' and make it to be the renderWatcher, so will run dep collection and collect the child component‘s property by renderWatcher.

ysj16:
@jacekkarczmarczyk distinguish undefined from null

ysj16:
@posva I've pushed a test

Issue #8912, opened 2018-10-09
A deliberately empty slot-scope attribute will not be rendered as a scoped slot


decademoon:
### Version
2.5.17

### Reproduction link
[https://jsfiddle.net/decademoon/50wL7mdz/759740/](https://jsfiddle.net/decademoon/50wL7mdz/759740/)

### Steps to reproduce
```
<foo>
  <bar slot-scope/>
</foo>
```

### What is expected?
`<bar>` should be a scoped slot.

### What is actually happening?
`<bar>` is a slot.

---
As a workaround, I've been using a dummy variable to force it to be a scoped slot:

```<bar slot-scope="scope"/>```

posva:
I'm pretty sure we talked about this in another issue but couldn't find it 🤔

basselAhmed:
Hi @decademoon I'd like to pick this up, can you check this sandbox and tell me if it's the correct behavior for your use case?
https://codesandbox.io/s/4q15jlq12w

smolinari:
For my curiosity, what is the use case for doing this?

Scott

decademoon:
@basselAhmed Yes that appears to be correct.

@smolinari One example is the `Foo` component may want to render the slot multiple times, in which case the slot needs to be scoped in order for it to be a function which will return a different vnode instance each time it is called.

smolinari:
Um, can you expand that with a real world example, maybe even with code? I'm still not understanding. Sorry. And thanks for being patient with me.

Scott

Justineo:
@smolinari See #8496. Currently Vue doesn't allow a slot to be rendered more than once, the workaround for now is using a scoped slot without meaningful scope arguments.

basselAhmed:
Thanks, I'll create a PR for this shortly.

WORMSS:
> @smolinari See #8496. Currently Vue doesn't allow a slot to be rendered more than once, the workaround for now is using a scoped slot without meaningful scope arguments.

Is that true? Because if you add two default slots, it's rendered twice. (Which I am trying to actually not happen)

jacekkarczmarczyk:
The comment was added before the fix was released

Issue #8496, opened 2018-07-16
Allow a single slot to be rendered for multiple times


Justineo:
### What problem does this feature solve?
Currently if we render a slot for multiple times (eg. inside a `v-for` loop), we'll get

```
Duplicate presence of slot "${name}" found in the same render tree - this will likely cause render errors.
```

I know it's intentional to prevent render errors (like losing correct data binding) but consider the following use case: I have a `v-breadcrumbs` component which has a default item separator of a text node `/`, and I want to allows users to specify their own separators which can be anything, like an `<v-icon>` component.

Currently if I want to make it work, I have to define a scoped slot in the `v-breadcrumbs` component and bind nothing to it:

```html
<slot name="sep" v-bind="{}">/</slot>
```

And component users must define `slot-scope` on it and not use anything from it:

```html
<v-breadcrumbs>
  <v-icon slot="sep" slot-scope="_" name="angle-right"/>
</v-breadcrumbs>
```

And currently Vue is using slots as fallbacks for scoped slots with the same name. Whether the users can use a slot doesn't rely on if they want data from slot scope, but on whether the slot is gonna be rendered for multiple times inside the component self, this may raise more confusion for our users (like #8175).

So why don't we just get rid of such caveats and let a slot to be rendered more than once? We can clone them on duplication to prevent render errors as we already run the scoped slot function each time anyway. This makes the logic a lot simpler for users IMO: if we do not expect to use data from the slot scope, just use a slot, otherwise we'll always have to declare `slot-scope` to create the binding.

In addition, this would make documenting the (scoped) slots of a component more consistent, we just declare the data structure for a slot scope and users can decide if they want some data to bind to the `slot-scope` and don't need to care about how many times it will be rendered.

### What does the proposed API look like?
No additional API needed, just clone the slot nodes upon render and remove the waring about duplicated slots.

It may be implemented in userland today in a quite hacky way, see: https://codesandbox.io/s/lp11y2wovz

I just tweaked the `cloneVNode` function from the Vue core a little bit and it seemed to be working as expected using render function (without losing reactivity and event binding). There maybe some other edge situations I missed so correct me if it's not such trivial.

liximomo:
Why duplicate presence of slot would cause errors? I think it's common to resue slot. Maybe we should remove this warning.

Justineo:
@liximomo See the link I posted earlier. It's not just easy as “removing the warning”.

ky-is:
[...]

PR #9107, opened 2018-11-24
Allow empty slot-scope to render a scoped slot instead of a normal one

basselAhmed:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Please check issue #8912 and #8496 
This PR allows rendering a single slot multiple times by passing an empty slot-scope which makes the slot be rendered as a scoped slot instead of a normal one.

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -204,7 +204,7 @@ export function parse (
       if (currentParent && !element.forbidden) {
         if (element.elseif || element.else) {
           processIfConditions(element, currentParent)
-        } else if (element.slotScope) { // scoped slot
+        } else if (element.slotScope || element.slotScope === '') { // scoped slot
           currentParent.plain = false
           const name = element.slotTarget || '"default"'
           ;(currentParent.scopedSlots || (currentParent.scopedSlots = {}))[name] = element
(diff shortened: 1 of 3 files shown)


src/compiler/parser/index.js
```diff
@@ -476,7 +476,10 @@ function processSlot (el) {
         )
       }
       el.slotScope = slotScope || getAndRemoveAttr(el, 'slot-scope')
-    } else if ((slotScope = getAndRemoveAttr(el, 'slot-scope'))) {
+    } else if (
+        (slotScope = getAndRemoveAttr(el, 'slot-scope')) ||
+        (slotScope = getAndRemoveAttr(el, 'slot-scope')) === ''
```
posva: shouldn't we be able to do `slotScope === ''` as the second check?

posva: I wonder if we could only check against `!= null`

basselAhmed: Yes, we could do that as the second check, I don't think null is the only falsy value we should check against @posva

posva: it's `null` and `undefined`, but yeah, I didn't check further, just dropped the idea haha

KaelWD:
Vue 2.6 now has `v-slot` instead, this can be closed.

PR #9143, opened 2018-12-04
fix(parser#9038): Fix for slot "name" property using interpolation as well…

zleight1:
… as the "slot" property. (vuejs#9038)

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -465,14 +465,28 @@ function processOnce (el) {
 
 function processSlot (el) {
   if (el.tag === 'slot') {
-    el.slotName = getBindingAttr(el, 'name')
-    if (process.env.NODE_ENV !== 'production' && el.key) {
-      warn(
-        `\`key\` does not work on <slot> because slots are abstract outlets ` +
-        `and can possibly expand into multiple elements. ` +
-        `Use the key on a wrapping element instead.`
-      )
+    let slotName 
+    slotName = getBindingAttr(el, 'name')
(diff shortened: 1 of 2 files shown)


posva:
Did you see there is already https://github.com/vuejs/vue/pull/9050 ?

zleight1:
Yes, I missed that originally but submitted the PR because the other one misses a spot a little bit further down (and because I didn’t see any other easy bugs to get into vue source code).

The added test that checks for interpolation on slot=“” checks for this scenario.

src/compiler/parser/index.js
```diff
@@ -465,14 +465,28 @@ function processOnce (el) {
...
 function processSlot (el) {
   if (el.tag === 'slot') {
-    el.slotName = getBindingAttr(el, 'name')
-    if (process.env.NODE_ENV !== 'production' && el.key) {
-      warn(
-        `\`key\` does not work on <slot> because slots are abstract outlets ` +
-        `and can possibly expand into multiple elements. ` +
-        `Use the key on a wrapping element instead.`
-      )
+    let slotName
```
satishrdd: Why use let if we are not editing slotName more than once here,why not use const ?

zleight1: I was trying to be consistent with the other code but you're right.

Either way there was another PR for this bug so I haven't followed up on this..

Issue #9181, opened 2018-12-11
genChildren function with error normalizationType


skyline0705:
### Version
2.5.19

### Reproduction link
[https://codepen.io/anon/pen/roaKMY](https://codepen.io/anon/pen/roaKMY)

### Steps to reproduce
open the link and open the console

### What is expected?
console should not has error, and the result in dom should be empty

### What is actually happening?
console shows "Cannot read property 'length' of undefined
​    at simpleNormalizeChildren"

---
before the version 2.5.16, it works fine, and I saw the source, when compile the template, function genChildren choose an error normalizationType, it should be `ALWAYS_NORMALIZE`.

and I think there is something wrong when compiling v-for and component with an "undefined" list because I find this words in `'src/core/vdom/helpers/normalize-children.js'`:

https://github.com/vuejs/vue/blob/d780dd2e2adcf71f40c086055a659a9a2b4a8282/src/core/vdom/helpers/normalize-children.js#L27-L30

niutech:
I can confirm this issue in vue@2.5.20 when `obj` is undefined in `v-for="o in obj"`.

PR #9182, opened 2018-12-11
fix(types): genChildren function with error normalizationType

skyline0705:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/render-helpers/render-list.js
+++ b/src/core/instance/render-helpers/render-list.js
@@ -32,8 +32,9 @@ export function renderList (
       ret[i] = render(val[key], key, i)
     }
   }
-  if (isDef(ret)) {
-    (ret: any)._isVList = true
+  if (!isDef(ret)) {
+    ret = []
   }
+  (ret: any)._isVList = true
   return ret
 }
--- a/test/unit/features/component/component.spec.js
+++ b/test/unit/features/component/component.spec.js
@@ -186,6 +186,50 @@ describe('Component', () => {
     }).then(done)
   })
 
+  it('components with v-for and empty list', done => {
+    const vm = new Vue({
+      template:
+        '<div attr>' +
+          '<foo v-for="item in list">{{ item }}</foo>' +
+        '</div>',
+      data: {
+        list: undefined
+      },
+      components: {
+        foo: {
+          template: '<div><slot></slot></div>'
+        },
+      }
+    }).$mount()
+    expect(vm.$el.innerHTML).toBe('')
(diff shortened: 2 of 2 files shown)


skyline0705:
bug fix for https://github.com/vuejs/vue/issues/9181

first I try to change function `genChildren`, and change `normalizationType` to `ALWAYS_NORMALIZE`, it works in my case

but when I try this code:
```javascript
var vm = new Vue({
  el: '#demo',
  template: `
      <div>
        <foo v-for="item in list">
          {{ item }}
        </foo>
      </div>
  `,
  data: {
    list: undefined
  },
  components: {
    foo: {
      template: '<div><slot></slot></div>'
    },
  }
})

setTimeout(() => {
  vm.list = [1,2,3,4,5];
}, 1000);
```

I found it now works because the root element `div` has no attrs, and found the function `createElement` has an error params order;

so second I try to change file `src/compiler/codegen/index`
and change the function call `_c`, change the code
```javascript
code = `_c('${el.tag}'${
    data ? `,${data}` : '' // data
  }${
    children ? `,${children}` : '' // children
  })`
```
to
```javascript
code = `_c('${el.tag}'${
    data ? `,${data}` : ',undefined' // data
  }${
    children ? `,${children}` : '' // children
  })`
```
I think it will works, but…… the unit test shows that I'm wrong, there are many error 😭

finally I fix the `renderList` function, let it always return an array, and everything goes well……


I know use an `undefined` value in v-for is invalid, but before v2.5.16, it works well, I think it should be warn in the next version like v2.6.x, and should keep work in v2.5.x

yyx990803:
I think it's better to handle it directly in `simpleNormalizeChildren` - it avoids allocating empty array on each render and allows the diff algorithm to take a faster path.

See 47487607fbb99339038cf84990ba341c25b5e20d

yyx990803:
Ended up using your fix in 847e4937 when trying to apply additional optimization for the case... sorry!

skyline0705:
Ok~ No problem~~And another thing, the fundamental cause of this is because the params order at function `createElement`, that is why I finally decided to change function `renderList`: the `createElement` calls are too flexible😂.

maybe there need a little more test case of compile template, like the test case `components with v-for and empty list, root element has no attrs` I added before. the steps is on the last comment https://github.com/vuejs/vue/pull/9182#issuecomment-446085762

PR #9217, opened 2018-12-18
feat(compiler): switch to new `removeWhitespace` option (#9208)

### LINKED ISSUE #9208: Handle whitespaces in templates more wisely
Currently we have an boolean option for `vue-template-compiler`: `preserveWhitespace`. It either removes all whitespace-only text nodes, or leave them untouched. There were already earlier questions about either behavior: #6200, #7701, #9021, #9127. I think the current behavior is kind of oversimplified to cover actual usage.

When we write a template we tend to leverage line breaks and indents to make it more readable, like:

```html
<div class="item">
  <div class="aside">Aside</div>
  <div class="main">Main</div>
</div>
```

And if we choose to layout this component with inline formatting context, sometimes we may not want to precisely control the margin between the inner parts with CSS, instead of using the size of a whitespace (which is related to those `font-*` styles). For similar cases we don't want these whitespace-only text nodes. This leads to `preserveWhitespace: false` (and it even became the default behavior for Vue CLI 3: https://github.com/vuejs/vue-cli/commit/1864cef09e186cdf094fca876f8638a8ba1b4adb).

But when we craft some document/article-like content, this behavior becomes annoying. With `preserveWhitespace: false`, the following template:

```html
<p>
  Welcome to <b>Vue.js</b> <i>world</i>.
  Have fun!
</p>
```

Will generate:

```html
<p>
  Welcome to <b>Vue.js</b><i>world</i>.
  Have fun!
</p>
```

Which looks like:

> Welcome to <b>Vue.js</b><i>world</i>. Have fun!

And this is clearly not desired.

-In short, I suggest we offer a new compiler option, to apply the strategy React uses to handles whitespaces for JSX ([source](https://reactjs.org/docs/jsx-in-depth.html#string-literals-1)):

> JSX removes whitespace at the beginning and ending of a line. It also removes blank lines. New lines adjacent to tags are removed; new lines that occur in the middle of string literals are condensed into a single space.

For examples:

```html
<p>
  Welcome to <b>Vue.js</b> <i>world</i>.
  Have fun!
</p>
```

The whitespaces between `<p>` and `Welcome` are removed but the one between `</b>` and `<i>` and the one between `.` and `Have` are preserved thus giving us:

```html
<p>Welcome to <b>Vue.js</b> <i>world</i>. Have fun!</p>
```

This seem to be much more reasonable IMO. And in this mode users will have more flexibility to better serve their different purposes.

In general, the proposal is:

1. Keep `preserveWhitespace` but mark it as deprecated.
2. Offer a new option to specify whether/how to remove whitespace-only text nodes: `removeWhitespace: 'with-linebreak' | 'any' | 'none'`.
3. Ignore `preserveWhitespace` if `removeWhitespace` is specified.

Justineo:
Provide `removeWhitespace` option for smarter whitespace trimming. Resolves #9208.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

See ##9208.

**Other information:**

Also added the missing test cases for the old `preserveWhitespace` option.

### DIFF
--- a/test/unit/modules/compiler/parser.spec.js
+++ b/test/unit/modules/compiler/parser.spec.js
@@ -757,4 +757,81 @@ describe('parser', () => {
     const ast = parse(`<p>{{\r\nmsg\r\n}}</p>`, baseOptions)
     expect(ast.children[0].expression).toBe('_s(msg)')
   })
+
+  it('should remove whitespaces correctly with `preserveWhitespace: false`', () => {
+    const options = extend({
+      preserveWhitespace: false
+    }, baseOptions)
+
+    const ast = parse('<p>\n  Welcome to <b>Vue.js</b> <i>world</i>.\n  Have fun!\n</p>', options)
+    expect(ast.tag).toBe('p')
+    expect(ast.children.length).toBe(4)
+    expect(ast.children[0].type).toBe(3)
+    expect(ast.children[0].text).toBe('\n  Welcome to ')
+    expect(ast.children[1].tag).toBe('b')
+    expect(ast.children[1].children[0].text).toBe('Vue.js')
+    expect(ast.children[2].tag).toBe('i')
+    expect(ast.children[2].children[0].text).toBe('world')
+    expect(ast.children[3].type).toBe(3)
+    expect(ast.children[3].text).toBe('.\n  Have fun!\n')
+  })
+
+  it('should remove whitespaces correctly with `removeWhitespace: \'with-line-break\'`', () => {
+    const options = extend({
+      removeWhitespace: 'with-line-break'
+    }, baseOptions)
+
+    const ast = parse('<p>\n  Welcome to <b>Vue.js</b> <i>world</i>.\n  Have fun!\n</p>', options)
+    expect(ast.tag).toBe('p')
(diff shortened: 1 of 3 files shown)


test/unit/modules/compiler/parser.spec.js
```diff
@@ -757,4 +757,81 @@ describe('parser', () => {
...
+    expect(ast.children[0].type).toBe(3)
+    expect(ast.children[0].text).toBe('Welcome to ')
+    expect(ast.children[1].tag).toBe('b')
+    expect(ast.children[1].children[0].text).toBe('Vue.js')
+    expect(ast.children[2].type).toBe(3)
+    expect(ast.children[2].text).toBe(' ')
+    expect(ast.children[3].tag).toBe('i')
+    expect(ast.children[3].children[0].text).toBe('world')
+    expect(ast.children[4].type).toBe(3)
+    expect(ast.children[4].text).toBe('.\n  Have fun!')
```
yyx990803: Shouldn't this newline be condensed to a single space?

Justineo:
Nice 👍

Issue #9205, opened 2018-12-16
[ssr] Add Vue function to render VNode to html string


DominusVilicus:
### What problem does this feature solve?
**Use case:**

I'm developing a Server-Side Renderer for Vue (which works with Express, Koa & etc. Will increase migration to Vue). For the SSR's head management to work, it needs a stable API to render `VNode`s to text.

**The way my Vue SSR package will function:**
`master.vue`
```html
<template>
    <div id="app">
        <slot name="content"></slot>
    </div>
</template>
<script>
export default {
    created: function(){
        if(this.$isServer){
            this.$ssrContext.head = "HEAD HERE" // Something needed like:  renderVNodesToString(this.$slots.head)
        }
    },
}
</script>
```

`home.vue`
```html
<template>
    <master>
        <template slot="content">
            Hello World
        </template>
        <template slot="head">
            <script src="https://unpkg.com/vue/dist/vue.js"></script>
            <title>Hello</title>
        </template>
    </master>
</template>
<script>
import master from "layouts/master.vue"

export default {
    components: {
        master
    }
}
</script>
```

My goal is getting `home.vue`'s `head` slot rendered into a string and injecting it into the `this.$ssrContext` so it can be read and injected on the server-side

in `master.vue`, I can access `this.$slots.head` with no issue, and it contains the correct `VNode`s

my question is, how can I render them into a string? a way to basically do:
```js
this.$ssrContext.head = renderVNodesToString(this.$slots.head)
```

From my research, I have been unable to find an easy way to do this.

---- 

**To understand how the renderer works**
```js
const renderer = createBundleRenderer(bundle.server, {
    runInNewContext: false,
    inject: false,
    template: `<!DOCTYPE html>
        <html>
            <head>
                {{{ head }}}
                {{{ renderResourceHints() }}}
                {{{ renderStyles() }}}
            </head>
            <body>
                <!--vue-ssr-outlet-->
                <script>${ bundle.client }</script>
            </body>
       </html>`
})
```
This is the code for the serverbundlerenderer

### What does the proposed API look like?
```js
/**
* @param {VNode}
* 
* @returns {string} - VNode rendered to a html string
*/
Vue.renderVNode = function(VNode){
    //...
}
```

posva:
Technically, you should be able to do this by using a function that returns the vnode and render using what Vue already exports (https://ssr.vuejs.org/guide/#rendering-a-vue-instance).
Not sure of the utility of a feature like this.
[...]

PR #9324, opened 2019-01-15
[feature] allow template to be function in vue-server-renderer

DominusVilicus:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

@yyx990803 I've found a solution to #9205 

It's a very simple addition, to the `vue-server-renderer` package that has tonnes of use-cases and actually makes the `vue-server-renderer` much easier to use, as it provides a template function to the bundle renderer template compiler.

This isn't a breaking change and doesn't really need any tests, as it is a user-provided function

It would be amazing if this could be added in **2.6**, so I could use it in my package

### Use cases
1. Allows head management within `single file components` (see #9205)
2. Allows async functions within the `template` option
3. Makes the `vue-server-renderer` templater much easier to use, and allows users to use javascript interpolations. For example: `<!--vue-ssr-outlet-->` becomes `${result}`

#### Bonus
I'm creating a serverside rendering templating engine for **Vue** that will use this feature. It includes:
- A super simple head management system that exists entirely within the `<template>`, like native html
- full server-side rendering with Webpack
- CommonJS based, so users can easily implement it into koa/express projects
- templating using slots
- runtime compilation of Vue files (which are cached)

The package is aimed for users who want to start using Vue but want to keep express/koa routing which will make building serverside applications much easier for users, and will increase Vue adoption in the process

### DIFF
--- a/src/server/create-renderer.js
+++ b/src/server/create-renderer.js
@@ -81,12 +81,12 @@ export function createRenderer ({
         return false
       }, cb)
       try {
-        render(component, write, context, err => {
+        render(component, write, context, async (err) => {
           if (context && context.rendered) {
             context.rendered(context)
           }
           if (template) {
-            result = templateRenderer.render(result, context)
+            result = await templateRenderer.render(result, context)
           }
           if (err) {
             cb(err)
--- a/src/server/create-renderer.js
+++ b/src/server/create-renderer.js
@@ -23,7 +23,7 @@ export type RenderOptions = {
   directives?: Object;
   isUnaryTag?: Function;
   cache?: RenderCache;
-  template?: string;
+  template?: string | Function;
   inject?: boolean;
   basedir?: string;
   shouldPreload?: Function;
@@ -81,12 +81,13 @@ export function createRenderer ({
         return false
       }, cb)
       try {
-        render(component, write, context, async (err) => {
+        render(component, write, context, err => {
           if (context && context.rendered) {
             context.rendered(context)
           }
           if (template) {
-            result = await templateRenderer.render(result, context)
+            new Promise(resolve => resolve(templateRenderer.render(result, context)))
+              .then(res => cb(res))
           }
           if (err) {
             cb(err)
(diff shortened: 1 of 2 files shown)


DominusVilicus:
Please see #9205 for use case

It's great for head management, and actually makes the `vue-server-renderer` templating system much easier to use for users.

```js
const renderer = createBundleRenderer(bundle.server, {
            template: async function(result, context){
                return `
                <!DOCTYPE html>
                    <html>
                    <head>
                        ${ context.renderStyles() }
                    </head>
                    <body>
                        ${result}
                    </body>
                </html>`
            }
        })
```

DominusVilicus:
### Use cases
1. Allows head management within `single file components` (see #9205)
2. Allows async functions within the `template` option
3. Makes the `vue-server-renderer` templater much easier to use, and allows users to use javascript interpolitations. For example: `<!--vue-ssr-outlet-->` becomes `${result}`

### Bonus
I will be creating one of **Vue**'s simplest and easiest to use head management system that exists entirely within the `<template>`, like native html, and have created a vue server-side-renderer for koa & express.

Both will increase adoption of vue, and will make building serverside applications much easier for users

DominusVilicus:
Just realised vue's codebase doesn't accept async/await. Let me transform it into ES5 - Friendly code

DominusVilicus:
@yyx990803 will this be making 2.6? If so, I can assist in writing the documentation for this feature

src/server/create-renderer.js
```diff
@@ -23,7 +23,7 @@ export type RenderOptions = {
   directives?: Object;
   isUnaryTag?: Function;
   cache?: RenderCache;
-  template?: string;
+  template?: string | (content: string, context: any) => string;
```
lbogdan: Wouldn't `(context, content)` be a better API, consistent with, for example, vuex actions, which have `context` as their first argument?

PR #9347, opened 2019-01-20
types: changed Vue.set and Vue.delete typings to use keyof

zlumer:
Changed `Vue.set()` and `Vue.delete()` typings to prevent accidental index mismatch

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:
TypeScript typings

- [x] Yes
- [ ] No

This PR can break apps that use Vue.set/Vue.delete with incorrect types, but I wouldn't consider it to be a breaking change since the builds were already (unknowingly) broken before.
However, keys of type "string" may stop working without casting to a proper key type.

**Other information:**

While using `Vue.set()`, it was possible to update a non-existing field on an object, or pass an object of incorrect type as value and break everything.
I've changed `Vue.set()` and `Vue.delete()` declarations to better match their intent: now it can only be used with correct types.

Before:
```typescript
let obj = { height: 100 }
Vue.set(obj, "heigth", 50) // typo in the field name is ignored, obj becomes { height: 100, heigth: 50 }
console.log(obj.height) // still prints "100"
```
After:
```typescript
let obj = { height: 100 }
Vue.set(obj, "heigth", 50) // tsc error
```

Before:
```typescript
let obj: { width?: number } = { }
Vue.set(obj, "width", "0px") // field type is ignored, obj becomes { width: "0px" } while still typed as { width?: number }
if (typeof obj.width !== "undefined")
  obj.width.toFixed(10) // throws
```

After:
```typescript
let obj: { width?: number } = { }
Vue.set(obj, "width", "0px") // tsc error
```

I think that we can also remove the `Vue.set()`/`Vue.delete()` declarations with arrays, since `keyof` works on arrays as well.
This will become:
```typescript
export interface VueConstructor<V extends Vue = Vue> {
  // ...
  set<U, K extends keyof U>(objectOrArray: U, key: K, value: U[K]): U[K];
  delete<U>(objectOrArray: U, key: keyof U): void;
  // ...
}
```

### DIFF
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -91,9 +91,9 @@ export interface VueConstructor<V extends Vue = Vue> {
 
   nextTick(callback: () => void, context?: any[]): void;
   nextTick(): Promise<void>
-  set<T>(object: object, key: string | number, value: T): T;
+  set<U extends {}, K extends keyof U>(object: U, key: K, value: U[K]): U[K];
   set<T>(array: T[], key: number, value: T): T;
-  delete(object: object, key: string | number): void;
+  delete<U extends {}>(object: U, key: keyof U): void;
   delete<T>(array: T[], key: number): void;
 
   directive(


posva:
Thanks but this would break existing usage of `set` and `delete` as the key may not exist in the target object

zlumer:
I would argue that if the target object _type_ does not contain necessary keys, `Vue.set()` should not be allowed to add them and `Vue.delete()` definitely can't delete them (they don't exist).
At least `Vue.delete()` behaviour has to be changed.

Also, can you please elaborate what are the use cases for adding a field to an object if this object's type doesn't have this new field?

Consider this example:
```typescript
let x = {}
Vue.set(x, 'foo', 'bar')
x.foo // error
```
Doesn't look right to me.

zlumer:
Definitely read that before 😁
What I mean is that TypeScript itself does not allow creation of new properties on the object if the type doesn't define them. We can't do
```typescript
let obj = {}
obj.foo = 'bar'
```
I believe it should work identically with `Vue.set()`, one of the reasons being that we can't use that property immediately afterwards:
```typescript
let obj = {}
Vue.set(obj, 'foo', 'bar') // no error (but there should be one)
obj.foo // error
```

jonataswalker:
> You should read the documentation: https://vuejs.org/v2/api/#Vue-set

Isn't this too superficial?

zlumer:
@zhiquan-yu properties added to a reactive object with `Vue.set()` are not accessible immediately afterwards, as you can see in the example I provided earlier.
Deleting non-existing properties also doesn't make sense at all.
Please pay attention to the discussion.

Justineo:
Vue supports defining reactive data on the fly. So in your first example, Vue cannot tell whether it is a typo of you are intentionally adding a `heigth` property.

zhiquan-yu:
@zlumer You can try your demo at https://vuejs.org/v2/guide/ like this:

screen shot 2019-01-22 at 17 41 15

zlumer:
@zhiquan-yu it's JavaScript console in your browser, not TypeScript. Please stick to the subject here.

@Justineo that's what TypeScript types are for: we define what properties the object can have in it's type, and the compiler triggers an error if we try to access a non-existing property.
Consider this example:
```typescript
let obj: { foo?: string } = {}
obj.foo = 'bar' // OK
Vue.set(obj, 'foo', 'bar') // OK
obj.nonexisting = 'property' // not OK
Vue.set(obj, 'nonexisting', 'property') // OK, but should give an error instead
```
If we want to have a truly extendable object, we can define it's type as follows:
```typescript
let obj: { [key: string]: string | boolean } = {}

obj.foo = 'bar' // OK
Vue.set(obj, 'foo', 'bar') // OK

obj.bool = true // OK
Vue.set(obj, 'bool', true) // OK

obj.number = 0 // not OK
Vue.set(obj, 'number', 0) // OK, but should give an error instead
```

Issue #1961, opened 2015-12-04
Null is not treated as Object in component prop check


fnlctrl:
I have a component that let's user select an object from an array.
```
<comp :selected.sync="selectedObject">
// parent data
data: {
    selectedObject: null
}
// comp prop
props: {
    selected: {twoWay: true, type: Object}
}
```

and in parent template I have something like
```html
<div v-if="selectedObject"> blah blah blah </div>
<div v-else> please select an object </div>
```
which displays different content based on whether an object is selected.

and I get
```
[Vue warn]: Invalid prop: type check failed for selected="selectedObject". Expected Object, got Null.
```

I know i can just remove the "type: Object" check and suppress the warning,
but wouldn't it be better if component props checks treat null as an object?

Because javascript does that (typeof null === 'object'),
and setting null as an initial value (to my knowledge) doesn't affect mutation observation.

yyx990803:
It fails because `null` is not an Object... the `typeof` behavior is commonly considered a bad design in the JavaScript language itself.

Just initialize it to an empty object.

fnlctrl:
Initializing with an {} will cause it to evaluate to true.. which means I'll have to create another variable to store whether an object is really selected.
I'll just remove the type check then..

lsbyerley:
@fnlctrl you just need to change it to `v-if="Object.keys(selectedObj).length > 0"`

dulowski-marek:
@fnlctrl Which is a workaround, where either using such an exaggerated expression in template or writing a getter for each object to check if it is initialized seems not right.

Using Typescript with Vue complicates this even more, as one has to denote all objects as `Partial<T>` or `T | {}` for that approach to work. 

What is the status of this issue in upcoming 3.0 release?

vannitotaro:
Another workaround is to define a custom validator:
https://github.com/vuejs/Discussion/issues/374#issuecomment-137981959

Anyway something Vue-native like `type: [Object, null]` would be really useful to handle `value` prop of `v-model` in input-like components with direct mapping with server-side model field when the field accept null value (for example a date picker with date not required).

Christhofernatalius:
So, the documentation means by **any type** is not actually any type?
![image](https://user-images.githubusercontent.com/13524958/51589558-f8c7d500-1f19-11e9-9442-9dff3b16b9e7.png)

Flamenco:
When using Typescript, I have no way to specify `string | null`.

My use-case is a filter that sets the value to null when cleared.
[...]

PR #9358, opened 2019-01-23
feat(props): allow defining a required prop as null

posva:
Based on #1961. It wasn't possible to specify a required prop of type object and pass null as the
value

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

The tests cases reflect what I'm trying to fix. I remember this was pointed out multiple times in the past. We cannot explicitely pass a `null` prop when the type is `Object` and the prop is required. This would keep current behaviour but allow having the type `null` for explicit null values. I want to add more tests and maybe refactor but I remember this was problematic as it could break some validation, so I would like so feedback about types you may use that would break with this


Justineo:
Some potential inconsistency here:

```js
props: {
  foo: null, // skip
  bar: {
    type: null // skip
  },
  baz: {
    type: [String, null] // String or null
  }
}
```

While using constructors, all three usage share the same semantics.

posva:
Added some tests to reflect that as well

dilraj-vidyard (approved):

klondikemarlen:
What is the solution to this issue now that Vue 2 is deprecated? Just to not use null? (Obviously upgrading to Vue 3 would be a better solution)

Apparently you can use
```ts
withDefaults(
  defineProps<{
    foo?: number | null
  }>(),
  {
    foo: null,
  }
)
```
And Vue 2 will be happy.

PR #9371, opened 2019-01-25
perf: improve scoped slots change detection accuracy

yyx990803:
Ensure that state mutations that only affect parent scope only trigger parent update and does not affect child components with only scoped slots.

Implications:

- For optimal performance, scoped slots should always be preferred. (In v3 all slots are unified to work like scoped slots so that will be the default)

- In 2.6, any slots using the new `v-slot` syntax are compiled as scopedSlots internally. This doesn't affect the child if it's using template, since `<slot/>` can render both scoped and normal slots, but if the child is using render functions, it is recommended to always use `this.$scopedSlots`. It is safe to do so in 2.6 since all normal slots are proxied on `this.$scopedSlots` as well.

- Unfortunately, any child components with static slot content still need to be forced updated. This means the common use case of `<parent><child></child></parent>` doesn't benefit from this change, unless the default slot is explicitly forced into a scoped slot by using `<parent v-slot:default><child></child></parent>`. (We cannot directly force all slots into scoped slots as that would break existing render function code that expects the slot to be present on `this.$slots` instead of `this.$scopedSlots)

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -119,6 +119,7 @@ declare type ASTElement = {
   transitionMode?: string | null;
   slotName?: ?string;
   slotTarget?: ?string;
+  slotTargetDynamic?: boolean;
   slotScope?: ?string;
   scopedSlots?: { [name: string]: ASTElement };
 
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -354,11 +354,12 @@ function genScopedSlots (
   slots: { [key: string]: ASTElement },
   state: CodegenState
 ): string {
+  const hasDynamicKeys = Object.keys(slots).some(key => slots[key].slotTargetDynamic)
   return `scopedSlots:_u([${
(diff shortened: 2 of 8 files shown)


johnsoncheg:
Did the last sentence of the `implications` said that it's not recommended to use the new syntax to force transfer the `normal slot` to `scoped slot` according to the phase `We cannot directly force all slots into scoped slots as that would break existing render function code that expects the slot to be present on `this.$slots` instead of `this.$scopedSlots` @yyx990803

Yuliang-Lee:
Why static slot need to be forceUpdate?I can't think of any scenes that need forceUpdate

marsprince:
> Why static slot need to be forceUpdate?I can't think of any scenes that need forceUpdate

I think the reason that child components need to be forceUpdate is vue doesn't "know" static slot changed. Slot vnode is rendered in parent but used in child. We can't compare two vnode to see if it changed or unchanged.

PR #9379, opened 2019-01-27
feat(types): expose ElementAttributesProperty for TSX

andoshin11:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

# Descriptions
With current type definitions, it's extremely difficult for developers to pass props to children components when writing TSX since there's no exposed property that JSX can handle as element attributes properties.

 
This PullRequest achieves 2 things.
- Creates new property `_attrs` to the `ExtendedVue` type so it can be used inside `JSX.ElementAttributesProperty`.(Examples are below)
- Transform non-required props into optional properties of `props` interface.

## Example

Given these 2 children components...

```js
// PrimaryButton.tsx
import Vue from 'vue'

export default Vue.extend({
  name: 'Primary Button',
  props: {
    label: {
      type: String,
      required: true as true
    },
    color: String
  },
  computed: {
    style(): string {
      return `color: ${this.color || 'blue'};`
    }
  },
  render() {
    return <button style={this.style}>{this.label}</button>
  }
})
```

```js
// SecondaryButton.tsx
import Vue from 'vue'

export default Vue.extend({
  name: 'Secondary Button',
  props: ['color', 'label'],
  computed: {
    style(): string {
      return `color: ${this.color || 'blue'};`
    }
  },
  render() {
    return <button style={this.style}>{this.label}</button>
  }
})
```

and prepare `shims-tsx.d.ts` like this

```js
// shims-tsx.d.ts
import Vue, { VNode } from 'vue'

declare global {
  namespace JSX {
    interface Element extends VNode {}
    interface ElementClass extends Vue {}
    interface ElementAttributesProperty {
      _attrs: any
    }
    interface IntrinsicElements {
      [elem: string]: any
    }
  }
}
```

Property completion works as expected.

Type error warned as expected.

and no more errors with correct props declaration 👍

### DIFF
--- a/types/jsx.d.ts
+++ b/types/jsx.d.ts
@@ -1,0 +1,25 @@
+import { VNodeData } from "./vnode";
+
+export type KnownAttrs = Pick<
+  VNodeData,
+  "class" | "staticClass" | "key" | "ref" | "slot"
+> & {
+  style?: VNodeData["style"] | string
+  id?: string
+  refInFor?: boolean
+  domPropsInnerHTML?: string
+};
+
+type OuterProps<Props, RequiredPropNames extends keyof Props> = {
+  [K in RequiredPropNames]: Props[K]
+} &
+  { [K in Exclude<keyof Props, RequiredPropNames>]?: Props[K] };
+
+/**
+ * `_attrs` is a special property for using Vue components as JSX elements.
+ * By passing `_attrs` to `JSX.ElementAttributesProperty`, your component's props are
+ * exposed as JSX elemnt properties.
+ */
+export type JSXElementAttributesProperty<Props, RequiredPropsNames extends keyof Props> = {
+  _attrs: KnownAttrs & OuterProps<Props, RequiredPropsNames>
+};
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -2,7 +2,7 @@ import { Vue, CreateElement, CombinedVueInstance } from "./vue";
 import { VNode, VNodeData, VNodeDirective, ScopedSlot } from "./vnode";
 
 type Constructor = {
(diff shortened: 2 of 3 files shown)


ktsn:
Thanks for your amazing work! I briefly looked into it and it would be so useful for TSX users!
Only thing I'm concerning is that it will make type parameters of component more complicated. I'm not sure if we can directly declare the required/optional info into `Props` type parameter but I think it is ideal. I'll take a look if we can simplify the type declaration more.

As a side note, in Vue v3, we can declare `Props` type by hand. So we may include required/optional info in `Props` type directly.

skyrpex:
Any chance of this working along with https://github.com/kaorun343/vue-property-decorator? It's the only way I found to enable typehinting and autocompletion in my editor.

skyrpex:
Maybe we can't use the `@Prop` decorator but use some sort of `Vue.extend` to define the props and then make the class component extend it (something similar to what React users need to do).

```jsx
import { Vue, Component, Prop } from 'vue-property-decorator'

const YourComponentProps = Vue.props({
    propA: Number,
    propB: { type: String, default: 'default value' },
    propC: [String, Boolean],
});

@Component
export default class YourComponent extends Vue<YourComponentProps> {
    get text() {
        return `${this.propA} ${this.propB} ${this.propC}`;
    }
}
```

mbostwick:
@baocancode did you see the conflicting file blocking the automated checks?

andoshin11:
I left this PR wondering for too much time. I'll close this until Vue 3 comes along.
Thx for all of your feedbacks! much appreciated :)

PR #9386, opened 2019-01-29
style does not work with zIndex (fix #9382)

### LINKED ISSUE #9382: vbind:style does not work with zIndex
### Version
2.5.22

### Reproduction link
[https://codesandbox.io/s/rm36pyr37q](https://codesandbox.io/s/rm36pyr37q)

### Steps to reproduce
Given a very simple component:

```
<template>
  <div v-bind:style="styleObject">This div should have a large z-index</div>
</template>

<script>
export default {
  name: "HelloWorld",
  data() {
    return {
      styleObject: {
        maxWidth: "400px",
        minWidth: "auto",
        opacity: 1,
        padding: "5px",
        position: "fixed",
        right: "10px",
        top: "10px",
        zIndex: "100 !important" // this doesn't work
        // "z-index": "100 !important" // this works
      }
    };
  }
};
</script>
```

`zIndex` is not bound to style

### What is expected?
Binding `zIndex` does not get kebab-cased like `maxWidth` and `minWidth`. I expected it to work the same way. Passing `z-index` works as expected.

The `zIndex` variable is ignored entirely.

---
I was creating an alert-style box that needs a high z-index to appear "on top" of other elements on the page.

Itisfilipe:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The [documentation](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty#Parameters) states that the first parameter for `setProperty` must be hyphenated.

closes #9382

### DIFF
--- a/src/platforms/web/runtime/modules/style.js
+++ b/src/platforms/web/runtime/modules/style.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import { getStyle, normalizeStyleBinding } from 'web/util/style'
-import { cached, camelize, extend, isDef, isUndef } from 'shared/util'
+import { cached, camelize, extend, isDef, isUndef, hyphenate } from 'shared/util'
 
 const cssVarRE = /^--/
 const importantRE = /\s*!important$/
@@ -10,7 +10,7 @@ const setProp = (el, name, val) => {
   if (cssVarRE.test(name)) {
     el.style.setProperty(name, val)
   } else if (importantRE.test(val)) {
-    el.style.setProperty(name, val.replace(importantRE, ''), 'important')
+    el.style.setProperty(hyphenate(name), val.replace(importantRE, ''), 'important')
   } else {
     const normalizedName = normalize(name)
     if (Array.isArray(val)) {
--- a/test/unit/features/directives/style.spec.js
+++ b/test/unit/features/directives/style.spec.js
@@ -101,6 +101,13 @@ describe('Directive v-bind:style', () => {
     }).then(done)
   })
 
+  it('camelCase with !important', done => {
+    vm.styles = { zIndex: '100 !important' }
+    waitForUpdate(() => {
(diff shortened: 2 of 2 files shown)


src/platforms/web/runtime/modules/style.js
```diff
@@ -1,18 +1,19 @@
...
 const setProp = (el, name, val) => {
+  const hyphenatedName = hyphenate(name)
   /* istanbul ignore if */
-  if (cssVarRE.test(name)) {
-    el.style.setProperty(name, val)
+  if (cssVarRE.test(hyphenatedName)) {
+    el.style.setProperty(hyphenatedName, val)
   } else if (importantRE.test(val)) {
-    el.style.setProperty(name, val.replace(importantRE, ''), 'important')
+    el.style.setProperty(hyphenatedName, val.replace(importantRE, ''), 'important')
```
Justineo: We only need hyphenated name here. Other modification can be discarded.

```suggestion
    el.style.setProperty(hyphenate(name), val.replace(importantRE, ''), 'important')
```

Itisfilipe: Thanks for the review, this was also my first thought but considering that setProperty requires a hyphenated parameter I don't see a problem of letting this change happen in all invocations. Also, in line 16 the normalize function will convert it to camelCase every time so I don't see a problem there too. If we ultimately want to keep the scope of the change smaller as possible I agree with your proposal but if not I think is better for consistency to not have different versions of the same information spread over the function.

yyx990803: It's mostly performance. If we can avoid the hyphenate call in some branches then we should.

PR #9430, opened 2019-02-05
feat: add origin prop for <transition-group>, fix #8424

### LINKED ISSUE #8424: <transition-group> not working as expected on page scroll
### Version
2.5.16

### Reproduction link
[https://codesandbox.io/s/kxkmp9mov3](https://codesandbox.io/s/kxkmp9mov3)

### Steps to reproduce
1. Create a list using `<transition-group>`.
2. Put the list at the end of a long page.
3. Scroll to the bottom of the page.
4. Trigger some change so that the page height is reduced and force the viewport to scroll upwards a little bit.

### What is expected?
The existing items should stick to the container, instantly appear in the final position, without transitions.

The existing items jump out of the container because of the sudden change on vertical position and slowly move back to the expected position.

---
Internally, <transition-group> is using `getBoundingClinetRect()` to track the positions of transition items. This works fine when no page scroll is introduced. But when browsers force page scroll on certain situations, the container (and the rest of the page) flashed into the final position while transition items are stuck in the old position based on the viewport, which makes it look like they suddenly jump out of the document and start performing unexpected transitions.

Maybe we can provide some new prop on `<transition-group>`, say, `origin: 'viewport' | 'document'`, to optionally calculate positions based on the canvas origin instead of the viewport.

Justineo:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

As described in #8424, `<transition-group>` now records the items' position relative to the viewport (via `getBoundingClientRect()`). While page scroll happens just before the items start to perform transition, it will look like that they are flashed out of the container and slowly move to their original position.

This PR added a new prop: `origin` for `<transition-group>`, which is a `string` value that indicates how we are gonna calculate the position of transition items. Available values are `viewport` (the current behavior) and `document`. When set to `document`, the position of the items are always calculated relative to the top left corner of the document element, which will retain their position in the document when page scroll occurs.

**Other information:**

Here is a show case of this PR: https://codepen.io/Justineo/project/editor/XxEKgb

Scroll to the bottom and click `-` to see the transition.


Justineo:
On second thought: we are not always performing page scroll on the root element. If the page scroll is triggered by an element other than the root element, this fix would fail. Perhaps we should provide a way for users to specify the “origin” element themselves?

tmorehouse:
What about detecting the `scrollParent` of the element? walking up the DOM tree to find out which element is scrollable.

Justineo:
@tmorehouse The problem is that scroll doesn't have to be triggered on the closest `scrollParent`. We may have enum values like `viewport`/`document`/`scrollParent` or find a way to let users specify the origin manually.

PR #9479, opened 2019-02-12
feat(ssr): add ability to cleanup after request

### LINKED ISSUE #9463: Provide way to destroy app in SSR
This issue is related to: https://github.com/vuejs/vue-router/issues/2606

To recap:
A memory leak happens when the `router-view` is programmed to appear conditionally, and the component matching the view has a `beforeRouteEnter` guard and a callback is passed to it's `next(...)` method (e.g. `next(vm => {})`).

This will cause `vue-router` to poll every 16ms until the `router-view` materializes.

In a typical SSR application an instance of the app is created per request, which means the `router-view` will never appear, causing infinitely recursing poll methods.

A potential fix to this would be to detect when the app is destroyed in `vue-router`'s `poll` method, and allow the user to destroy the app that they created in `entry-server.js`.

A simplified example:
```javascript
export default context => {
  return new Promise((resolve, reject) => {
    const { app, router } = createApp(context)

    const { url } = context

    router.push(url)

    router.onReady(() => {
      resolve(app).then(destroyApp)
    }, reject)
  })
}
```

ronald-d-rogers:
close #9463

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The point of this "feature" is to be able to do something to the `app` to mark it as "invalid" so that `vue-router` has a way to detect and cleanup objects it might have left hanging around in memory. This feature might be able to be used for other reasons, but it is probably truly a fix. Also, a change would need to be approved in `vue-router` as well for the memory leak there to even go away (see https://github.com/vuejs/vue-router/issues/2606), and that might get fixed by other means, so this might not be worth doing anything with if that doesn't also get approved.

Because we're is in control of when the app is created in SSR it didn't seem like there was any other sensible option but to give us control of when to clean it up, though maybe something can be done based on `runInNewContext`. Also, maybe a more explicit API would be better, e.g. some way to mark the app as "served", `$isServer && $isServed`.

Note I'm submitting this knowing that there is a good chance that this will be tossed, so feel free to do so if this is not what is wanted.

### DIFF
--- a/src/server/bundle-renderer/create-bundle-renderer.js
+++ b/src/server/bundle-renderer/create-bundle-renderer.js
@@ -28,6 +28,10 @@ type RenderBundle = {
   modules?: { [filename: string]: Array<string> };
 };
 
+function getRenderArguments(app) {
+  return app && app._isVue ? { app, onComplete: null } : app
+}
+
 export function createBundleRendererCreator (
   createRenderer: (options?: RenderOptions) => Renderer
 ) {
@@ -100,12 +104,16 @@ export function createBundleRendererCreator (
         run(context).catch(err => {
           rewriteErrorTrace(err, maps)
           cb(err)
-        }).then(app => {
+        }).then(args => {
+          const { app, onRequestComplete } = getRenderArguments(args)
           if (app) {
             renderer.renderToString(app, context, (err, res) => {
               rewriteErrorTrace(err, maps)
+              onRequestComplete && onRequestComplete()
               cb(err, res)
             })
+          } else {
+            onRequestComplete && onRequestComplete()
           }
         })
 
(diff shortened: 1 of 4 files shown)


SacDin:
Is it going to be merged ?

ronald-d-rogers:
PR made in vue-router illustrating what would need to be done there as well: https://github.com/vuejs/vue-router/pull/2867

Again I am skeptical as to whether or not this is the right approach, but I am skeptical of everything I do so take that for what it's worth.

ronald-d-rogers:
Note I made another PR made in vue-router illustrating a way to fix the leak without requiring this PR: https://github.com/vuejs/vue-router/pull/2875

PR #9526, opened 2019-02-19
fix(#9511): avoid promise catch multiple times

### LINKED ISSUE #9511: Vue.config.errorHandler called twice when I return a rejected Promise from a handler of a component event
### Version
2.6.6

### Reproduction link
[https://codepen.io/anon/pen/qgQYra?editors=1011](https://codepen.io/anon/pen/qgQYra?editors=1011)

### Steps to reproduce
The problem seems to happen when I handle an event emitted from a component and return a rejected promise. To see the problem in action, open the codepen link, click the button and you will see in the console two logs of the same error, which means the Vue.config.errorHandler was called twice.

### What is expected?
I expect the errrorHandler called once.

the errorHandler function is called twice.

Shasharoman:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
fixed #9511 

In some cases, invokeWithErrorHandling will have nested calls, and Promise will independently catch multiple times, resulting in multiple calls to handleError.

Look at a sample code：

``` javascript
var p = Promise.reject('some err');

p.catch(() => console.log('1'));  // p = p.catch(...)
p.catch(() => console.log('2'));
```
The above code will output 1 and 2, but if you reassign the first catch call to p, it will not output 2.

### DIFF
--- a/src/core/util/error.js
+++ b/src/core/util/error.js
@@ -36,7 +36,9 @@ export function invokeWithErrorHandling (
   try {
     res = args ? handler.apply(context, args) : handler.call(context)
     if (res && !res._isVue && isPromise(res)) {
-      res.catch(e => handleError(e, vm, info + ` (Promise/async)`))
+      // issue #9511
+      // reassign to res to avoid catch triggering multiple times when nested calls
+      res = res.catch(e => handleError(e, vm, info + ` (Promise/async)`))
     }
   } catch (e) {
     handleError(e, vm, info)
--- a/test/unit/modules/util/invoke-with-error-handling.spec.js
+++ b/test/unit/modules/util/invoke-with-error-handling.spec.js
@@ -1,0 +1,23 @@
+import Vue from 'vue'
+import { invokeWithErrorHandling } from 'core/util/error'
+
+describe('invokeWithErrorHandling', () => {
+  if (typeof Promise !== 'undefined') {
+    it('nested calls do not trigger multiple errorHandler calls', done => {
+      let times = 0
+
+      Vue.config.errorHandler = function () {
+        times++
+      }
+
+      invokeWithErrorHandling(() => {
+        return invokeWithErrorHandling(() => {
+          return Promise.reject(new Error('fake error'))
+        })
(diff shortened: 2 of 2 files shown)


Shasharoman:
`npm run test:ssr` will also fail before my code is submitted. What should I do?

Shasharoman:
> Could you add a test case please?

I added a test case after your reminder, but I still don't know how to handle `npm run test:ssr` failure, because the original dev branch on my own computer can't pass ssr related tests.

Shasharoman:
> I think the test should go with test/unit/features/error-handling.spec.js

According to my understanding of the bug, the test case under test/unit/features/error-handling.spec.js might look like this:

``` javascript
it('config.errorHandler should capture async errors once', done => {
    let times = 0
    Vue.config.errorHandler = function () {
      times++
    }
    Vue.component('error-div', {
      mounted: function () {
        this.$emit('error')
      },
      render: function (h) {
        return h('div')
      }
    })
    const vm = new Vue({
      template: '<error-div @error="onError"></error-div>',
      methods: {
        onError: function() {
          return Promise.reject(new Error('async error'))
        }
      }
    }).$mount()

    Vue.nextTick(() => {
      expect(times).toBe(1)
      done()
    })
  })
```

The above test case will result in nested calls, but there are no other similar scenes in Vue that will cause such nested calls, which I don't know.

I think the above test case, other developers will look a bit confusing in the future, and in the `test/unit/modules/util/error/` directory, it will be easier to understand in the future. 

The essential reason for this problem is that Promise is not used correctly. This problem is exposed when the method is nested, so I think it is more reasonable to write test cases for the method.

If you still think that I should write test cases in test/unit/features/error-handling.spec.js, I will write the above test cases to this file and submit them again, or move the previously submitted test cases to this file.

My native language is not English, so I need to use google translation often, I hope you can understand. And I am also submitting code to Vue for the first time, not very familiar with the project, thank you very much for your review.

PR #9527, opened 2019-02-19
fix(utils): Handle more edgecases in `looseEqual()` util and add spec

jackmu95:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This PR makes the `looseEqual()` util more bulletproof and adds an extra spec for it.

The previous implementation failed at following tests:

```bash
Expected: false
Received: true

  65 |     // Files with slightly different dates
> 66 |     expect(looseEqual(file1, file3)).toBe(false)
```

```bash
Expected: false
Received: true

  102 |     // Different regex
> 103 |     expect(looseEqual(rx1, rx3)).toBe(false)
```

```bash
Expected: false
Received: true

> 153 |     expect(looseEqual(obj1, date1)).toBe(false)
```

Performance should also be equal or slightly better because early exits are used wherever possible:
https://jsperf.com/looseequal-comparison

Shoutout to @tmorehouse for helping with the tests and handling all the edge cases.

### DIFF
--- a/packages/vue-server-renderer/build.dev.js
+++ b/packages/vue-server-renderer/build.dev.js
@@ -41,10 +41,15 @@ function isPrimitive (value) {
   )
 }
 
+function isDate(obj) {
+  return obj instanceof Date
+}
+
 /**
  * Quick object check - this is primarily used to tell
  * Objects from primitive values when we know the value
  * is a JSON-compliant type.
+ * Note object could be a complex type like array, date, etc.
  */
 function isObject (obj) {
   return obj !== null && typeof obj === 'object'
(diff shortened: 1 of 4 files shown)


test/unit/util/loose-equal.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    // Dates with slightly different time (ms)
+    expect(looseEqual(date1, date3)).toBe(false)
+    // Dates with different year
+    expect(looseEqual(date1, date4)).toBe(false)
+  })
+
+  it('compares files correctly', () => {
+    const date1 = new Date(2019, 1, 2, 3, 4, 5, 6)
+    const date2 = new Date(2019, 1, 2, 3, 4, 5, 7)
+    const file1 = new File([''], 'filename.txt', { type: 'text/plain', lastModified: date1 })
```
posva: where is File coming from?

src/shared/util.js
```diff
@@ -283,38 +288,42 @@ export function genStaticKeys (modules: Array<ModuleOptions>): string {
...
-        return a.getTime() === b.getTime()
-      } else if (!isArrayA && !isArrayB) {
-        const keysA = Object.keys(a)
-        const keysB = Object.keys(b)
-        return keysA.length === keysB.length && keysA.every(key => {
-          return looseEqual(a[key], b[key])
-        })
-      } else {
-        /* istanbul ignore next */
+  if (a === b) {
```
posva: we like the one line 😉

packages/vue-server-renderer/build.dev.js
```diff
@@ -41,10 +41,15 @@ function isPrimitive (value) {
   )
 }
 
+function isDate(obj) {
+  return obj instanceof Date
+}
```
posva: make sure to remove any modifications to packages/* files

test/unit/util/loose-equal.spec.js
jackmu95: What do you mean with "coming from"?

jackmu95: https://stackoverflow.com/a/26181292

posva: sorry, I meant why is this a test?

jackmu95: Because the previous implementation failed here.
It treated two completely different files as same.

This test covers that.

jackmu95: The main issue was to use `Object.keys()` to compare the keys across two objects.
It returns an empty array  for e.g. File objects.

test/unit/util/loose-equal.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    const arr3 = [false, false, true]
+    const arr4 = [undefined, undefined, true]
+
+    expect(looseEqual(arr1, arr2)).toBe(true)
+    expect(looseEqual(arr2, arr1)).toBe(true)
+    expect(looseEqual(arr1, arr3)).toBe(false)
+    expect(looseEqual(arr3, arr1)).toBe(false)
+    expect(looseEqual(arr1, arr4)).toBe(true)
+    expect(looseEqual(arr4, arr1)).toBe(true)
+  })
```
tmorehouse: I would also add in the following test:

```js
    // This one is also sparse (missing index 1)
    const arr5 = []
    arr5[0] = arr5[2] = true

    // this one previously would return 'true', but shouldn't
    expect(looseEqual(arr1, arr5)).toBe(false)
    // this one previously would return 'false' (as it should)
    expect(looseEqual(arr5, arr1)).toBe(false)
```

jackmu95:
@posva @yyx990803 Since the same changes landed in Vue v3 it would be awesome to have them in Vue v2.7 too.

Ref: https://github.com/vuejs/vue-next/pull/379

PR #9550, opened 2019-02-22
test: Add a unit test for "$off event and this event added by $once"

Luxiaosong:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: add a unit test for "$off event and this event added by $once"

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/test/unit/features/instance/methods-events.spec.js
+++ b/test/unit/features/instance/methods-events.spec.js
@@ -56,7 +56,7 @@ describe('Instance methods events', () => {
     expect(spy).toHaveBeenCalledWith(1, 2, 3)
   })
 
-  it('$once', () => {
+  it('$off event added by $once', () => {
     vm.$once('test', spy)
     vm.$off('test', spy) // test off event and this event added by once
     vm.$emit('test', 1, 2, 3)
--- a/test/unit/features/instance/methods-events.spec.js
+++ b/test/unit/features/instance/methods-events.spec.js
@@ -58,7 +58,7 @@ describe('Instance methods events', () => {
 
   it('$once', () => {
     vm.$once('test', spy)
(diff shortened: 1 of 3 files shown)


Luxiaosong:
> Files in `packages` are automatically generated at build time, please do not include them in the PR. Thanks!

> Files in `packages` are automatically generated at build time, please do not include them in the PR. Thanks!
Thanks, I had revert file what I change in packages

Luxiaosong:
> Thanks for contributing. But I'm afraid this part cannot be removed.
> 
> After removing these two lines, if user call `vm.$off` before the event is triggered, it won't work.
> 
> eg.
> 
> ```js
> const fn = () => console.log('foo')
> this.$once('foo', fn)
> this.$off('foo', fn) // this won't work
> ```
> 
> Since we don't have unit test to cover this case, if you still willing to contribute, add a unit test for it would be appreciated.

Thanks, I will revert what I change and add a unit test for this.

test/unit/features/instance/methods-events.spec.js
```diff
@@ -56,6 +56,13 @@ describe('Instance methods events', () => {
     expect(spy).toHaveBeenCalledWith(1, 2, 3)
   })
 
+  it('$once', () => {
```
jkzing: Let's use a more specific description for this test case. eg. "$off event added by $once"

Luxiaosong: thanks, and I had do this.

jkzing:
```suggestion
  it('$off event added by $once', () => {
```

PR #9572, opened 2019-02-25
fix #9571, rendering async components after initial context was destroyed

### LINKED ISSUE #9571: Components created with async component factory break on route change
### Version
2.6.7

### Reproduction link
[https://codesandbox.io/s/239l9qnkxn](https://codesandbox.io/s/239l9qnkxn)

### Steps to reproduce
1. Open https://codesandbox.io/s/239l9qnkxn
2. Immediately (before the async component is resolved within 7 sec) click on the "Some Page" link
3. Click on the "Home" link (before 7s are elapsed)

### What is expected?
The AsyncReady component should render after 7 seconds.

The AsyncLoading component is shown forever

---
The problem is that the `currentRenderingInstance` is `Home`. After rendering the `SomePage` route, the initial `Home` instance (which is stored as the `owner` of the factory) is destroyed. As soon as the factory calls `forceRender`, triggering `$forceUpdate()` on the owner does nothing because the old `Home` instance already was destroyed and a new one was created when navigating back to `Home`.

This is problematic on very slow connections when users click a link before all dynamic components were resolved or in case of automatic redirects which are triggered immediately after a page is loaded (after checking auth for example).

See: https://github.com/vuejs/vue/blob/dev/src/core/vdom/helpers/resolve-async-component.js#L58-L74

maoberlehner:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No
- [ ] Maybe (I'm not sure)

**Other information:**

I have not heavily tested this in a lot of circumstances. The async part of it almost broke my brain ^^ But tests are passing, so... ¯\_(ツ)_/¯

### DIFF
--- a/src/core/vdom/helpers/resolve-async-component.js
+++ b/src/core/vdom/helpers/resolve-async-component.js
@@ -51,17 +51,26 @@ export function resolveAsyncComponent (
     return factory.resolved
   }
 
+  const owner = currentRenderingInstance
+  if (isDef(factory.owners) && factory.owners.indexOf(owner) === -1) {
+    // already pending
+    factory.owners.push(owner)
+  }
+
   if (isTrue(factory.loading) && isDef(factory.loadingComp)) {
     return factory.loadingComp
   }
 
-  const owner = currentRenderingInstance
-  if (isDef(factory.owners)) {
-    // already pending
-    factory.owners.push(owner)
-  } else {
+  if (!isDef(factory.owners)) {
     const owners = factory.owners = [owner]
     let sync = true
+
+    const removeOwner = (destroyedOwner) => {
+      const index = owners.indexOf(destroyedOwner)
+      if (index > -1) owners.splice(index, 1)
+    }
+
+    if (owner) owner.$on('hook:destroyed', () => removeOwner(owner))
 
     const forceRender = (renderCompleted: boolean) => {
       for (let i = 0, l = owners.length; i < l; i++) {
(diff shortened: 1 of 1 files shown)


src/core/vdom/helpers/resolve-async-component.js
```diff
@@ -51,18 +51,27 @@ export function resolveAsyncComponent (
...
     const owners = factory.owners = [owner]
     let sync = true
 
+    const removeOwner = (destroyedOwner) => {
+      const index = owners.indexOf(destroyedOwner)
+      if (index > -1) owners.splice(index, 1)
+    }
+
+    if (owner) owner.$on('hook:destroyed', () => removeOwner(owner))
+
```
maoberlehner: This stuff isn't mandatory. It doesn't break anything if destroyed owners are fore updated. But it might be cleaner to remove them.

src/core/vdom/helpers/resolve-async-component.js
```diff
@@ -51,18 +51,27 @@ export function resolveAsyncComponent (
...
   if (isTrue(factory.loading) && isDef(factory.loadingComp)) {
     return factory.loadingComp
   }
 
-  const owner = currentRenderingInstance
-  if (isDef(factory.owners)) {
-    // already pending
-    factory.owners.push(owner)
-  } else {
+  if (!isDef(factory.owners)) {
```
maoberlehner: Not too happy with double checking for `isDef(factory.owners)` here and above but cleaner than moving this block above the block which returns the `loadingComp` I guess?!

Issue #9577, opened 2019-02-26
Deep object as dynamic argument in custom directive [RFC-0003]


drozdzynski:
### Version
2.6.7

### Reproduction link
[https://jsfiddle.net/3qbky0z7/4/](https://jsfiddle.net/3qbky0z7/4/)

### Steps to reproduce
Create custom directive with a dynamic argument:
```v-test:[deep.arg]="value"```

### What is expected?
It should get the value from path `deep.arg`.

### What is actually happening?
Inside the binding object of a custom directive, `arg` key have value `[deep`.

posva:
For the moment use a computed property to get the nested value or use `v-test:[deep['arg']]="value"`

Shasharoman:
The workaround @posva mentioned will cause the console to print warn message:

![image](https://user-images.githubusercontent.com/5089251/53488787-49f35780-3aca-11e9-9b72-4a2221563bc9.png)

PR #9585, opened 2019-02-27
fix(#9577): support deep object as dynamic arguments

Shasharoman:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Fixed issue #9577.

The old modifier matching pattern is `/\.[^.]+/g`, It does not take into account that dynamic arguments may contain `.`，so I updated it and hope it is correct.

code example:

``` {javascript}
const before = /\.[^.]+/g;
const after = /\.[^.\]]+(?=\.|$)/g;

let s = 'name:[a.b].c.d';

console.log(s.match(before)); // [".b]", ".c", ".d"]
console.log(s.match(after));    // [".c", ".d"]
```

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -33,7 +33,7 @@ const dynamicArgRE = /^\[.*\]$/
 const argRE = /:(.*)$/
 export const bindRE = /^:|^\.|^v-bind:/
 const propBindRE = /^\./
-const modifierRE = /\.[^.\]]+(?=\.|$)/g
+const modifierRE = /(\.[^.\]]+)+$/g
 
 const slotRE = /^v-slot(:|$)|^#/
 
@@ -906,7 +906,7 @@ function parseModifiers (name: string): Object | void {
   const match = name.match(modifierRE)
   if (match) {
     const ret = {}
-    match.forEach(m => { ret[m.slice(1)] = true })
+    match[0].slice(1).split('.').forEach(item => ret[item] = true)
     return ret
   }
 }
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -33,7 +33,7 @@ const dynamicArgRE = /^\[.*\]$/
 const argRE = /:(.*)$/
 export const bindRE = /^:|^\.|^v-bind:/
 const propBindRE = /^\./
-const modifierRE = /\.[^.]+/g
+const modifierRE = /\.[^.\]]+(?=\.|$)/g
 
 const slotRE = /^v-slot(:|$)|^#/
 
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -33,7 +33,7 @@ const dynamicArgRE = /^\[.*\]$/
 const argRE = /:(.*)$/
 export const bindRE = /^:|^\.|^v-bind:/
 const propBindRE = /^\./
-const modifierRE = /(\.[^.\]]+)+$/g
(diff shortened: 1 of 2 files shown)


src/compiler/parser/index.js
```diff
@@ -33,7 +33,7 @@ const dynamicArgRE = /^\[.*\]$/
 const argRE = /:(.*)$/
 export const bindRE = /^:|^\.|^v-bind:/
 const propBindRE = /^\./
-const modifierRE = /\.[^.]+/g
+const modifierRE = /\.[^.\]]+(?=\.|$)/g
```
Justineo: `.a` in `'v-my:[deep.a.b]'` will be matched as a modifier.

test/unit/features/options/directives.spec.js
```diff
@@ -287,4 +287,28 @@ describe('Options directives', () => {
...
+          update(el, binding) {
+            expect(binding.arg).toBe('bar')
+            expect(binding.oldArg).toBe('foo')
+            done()
+          }
+        }
+      }
+    }).$mount()
+    vm.deep.key = 'bar'
+  })
```
Justineo: We may need more test case covering more dynamic arg expression. And maybe we shouldn't allow `]` inside the expression in the first place because that would make parsing a lot more complicated.

Thoughts? @yyx990803

src/compiler/parser/index.js
Shasharoman: Sorry, this is my mistake, I updated the regular expression again. And modified the `parseModifiers` method for new regular expressions. If there is any problem, please let me know.

test/unit/features/options/directives.spec.js
Shasharoman: I added a test case with dynamic arguments like `deep.a.b`.

src/compiler/parser/index.js
```diff
@@ -33,7 +33,7 @@ const dynamicArgRE = /^\[.*\]$/
 const argRE = /:(.*)$/
 export const bindRE = /^:|^\.|^v-bind:/
 const propBindRE = /^\./
-const modifierRE = /\.[^.\]]+(?=\.|$)/g
+const modifierRE = /(\.[^.\]]+)+$/g
```
Justineo: Use `/\.[^.\]]+(?=[^\]]*$)/g` instead so we can keep `parseModifiers` as is. I created a [benchmark](https://jsperf.com/vue-modifier-reg/) and it shows this is more performant.

test/unit/features/options/directives.spec.js
```diff
@@ -309,6 +309,32 @@ describe('Options directives', () => {
...
+          },
+          update(el, binding) {
+            expect(binding.arg).toBe('bar')
+            expect(binding.oldArg).toBe('foo')
+            done()
+          }
+        }
+      }
+    }).$mount()
+    vm.deep.a.b = 'bar'
```
Justineo: Maybe also `v-my:[deep.a.b].x.y`?

src/compiler/parser/index.js
Shasharoman: You are so amazing, thank you very much, I have modified the code according to your suggestion.

test/unit/features/options/directives.spec.js
Shasharoman: Added a test case for this.

yyx990803:
Thanks!

PR #9647, opened 2019-03-07
fix(scheduler): getNow detection can randomly fail (fix #9632)

### LINKED ISSUE #9632: Exception in IE 9 due to unsupported High Resolution Time API / getNow race-condition
### Version
2.6.8

### Reproduction link
[https://jsfiddle.net/re08ag7z/](https://jsfiddle.net/re08ag7z/)

### Steps to reproduce
When Vue.js is used in IE 9 (or IE 10/11 in IE 9 Emulation Mode), it sometimes crashes with an exception:

> [Vue warn]: Error in nextTick: "TypeError: Object doesn't support property or method 'now'"

This seems to be due to the usage of `performance.now()` which is not supported in IE 9, see:

https://caniuse.com/#feat=high-resolution-time

### What is expected?
It should not crash, since IE 9 is a supported browser for Vue 2.x.

It crashes sometimes and sometimes it works.

---
I think the problem is a race-condition in the detection code of [getNow](https://github.com/vuejs/vue/blob/v2.6.8/src/core/observer/scheduler.js#L44):

```javascript
var getNow = Date.now;
if (inBrowser && getNow() > document.createEvent('Event').timeStamp) {
  getNow = function () { return performance.now(); };
}
```

It looks like if `document.createEvent('Event').timeStamp` returns a unix timestamp, the result of this check is unpredictable, and is sometimes true and sometimes false.

Running the following in the IE 11 console when in IE 9 emulation mode sometimes returns true and sometimes false:

```javascript
Date.now() > document.createEvent('Event').timeStamp
```

This check can also fail in IE 11 in IE 10 emulation mode, so it probably affects IE 10 as well.

A possible workaround would be to check for something like `getNow() > document.createEvent('Event').timeStamp   1000`, though that still seems a bit brittle.

In my testing in a Windows 10 VM, the difference for `Date.now() - document.createEvent('Event').timeStamp` was usually  /- 2 ms, but sometimes I also got larger offsets, like Date.now() around 22 ms larger than timeStamp.

This problem could also be influenced by clock correction, when NTP or the VM host corrects the guest time between calls.

Generally comparisons of realtime clocks in computers is error prone and only monotonic timers should be trusted, which is probably one of the reasons why `performance.now()` was created…

felixbuenemann:
The previous detection code compared time stamps based on Date.now()
which are not monotonic, so the check could fail due to clock skew or
adjustments.

This fix changes the check to compare against performance.now() if it is
supported, because it is monotonic (strictly increasing).

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

See the discussion in #9632 for additional info.

This needs to be backported to the 2.6 branch.

### DIFF
--- a/src/core/observer/scheduler.js
+++ b/src/core/observer/scheduler.js
@@ -47,11 +47,13 @@ let getNow: () => number = Date.now
 // timestamp can either be hi-res (relative to page load) or low-res
 // (relative to UNIX epoch), so in order to compare time we have to use the
 // same timestamp type when saving the flush timestamp.
-if (inBrowser && getNow() > document.createEvent('Event').timeStamp) {
-  // if the low-res timestamp which is bigger than the event timestamp
-  // (which is evaluated AFTER) it means the event is using a hi-res timestamp,
-  // and we need to use the hi-res version for event listeners as well.
-  getNow = () => performance.now()
+if (inBrowser && window.performance && typeof performance.now === 'function') {
+  // if the event timestamp is bigger than the hi-res timestamp
+  // (which is evaluated AFTER) it means the event is using a lo-res timestamp,
+  // and we need to use the lo-res version for event listeners as well.
+  if (document.createEvent('Event').timeStamp <= performance.now()) {
+    getNow = () => performance.now()
+  }
 }
 
 /**
--- a/src/core/observer/scheduler.js
+++ b/src/core/observer/scheduler.js
@@ -47,13 +47,16 @@ let getNow: () => number = Date.now
 // timestamp can either be hi-res (relative to page load) or low-res
 // (relative to UNIX epoch), so in order to compare time we have to use the
 // same timestamp type when saving the flush timestamp.
-if (inBrowser && window.performance && typeof performance.now === 'function') {
+if (
+  inBrowser &&
+  window.performance &&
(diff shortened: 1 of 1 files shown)


yyx990803:
The check would lose the point if we compare the event timeStamp against a hi-res timeStamp. I think it should be good enough to only use `performance.now` when it's available since this only affects IE9.

felixbuenemann:
> The check would lose the point if we compare the event timeStamp against a hi-res timeStamp.

No, the point of the check is to see if the event timeStamp uses hi-res (like performance.now()) or lo-res (like Date.now()). So comparing if the event timeStamp is smaller or equal to performance.now() will work, because if it is lo-res it will be a lot bigger than performance.now().

Comparing to performance.now() fixes the problem, because it is not affected clock inaccuracies. 

> I think it should be good enough to only use performance.now when it's available since this only affects IE9.

No, it also affects IE 10/11 and any other browser that uses Unix timestamps (lo-res) for events.

yyx990803:
You are right. Reviewing PRs with a jetlag is a bad idea.

PR #9653, opened 2019-03-08
Fix #9641: v-bind object should be overridable by single bindings

### LINKED ISSUE #9641: camelCased props bound through an object take precedence over a kebab cased specific prop
### Version
2.6.8

### Reproduction link
[https://codesandbox.io/s/54v49n1nzn](https://codesandbox.io/s/54v49n1nzn)

### Steps to reproduce
The repro shows the issue directly.

```
<Child
      v-bind="{
        appleColor: 'objective red',
        bird: 'objective sparrow'
      }"
      :apple-color="`proper red`"
      :bird="`proper sparrow`"
    />
```

### What is expected?
I wanted to know what takes precedence when something is bound using `v-bind={}` vs set explicitly as a prop. I would expect this to always be one or the other, or to be resolved in order of declaration as when merging objects in JS ({...boundProps, specificProp })

The precedence is determined based on whether a prop has multiple words. duplicate keys are set by the object when they are declared in dash-case, and set by the prop/attribute if they are camelCase or a single word.

---
This can be made 'consistent' by declaring multiword props using camelcase, but that's against the recommendations in the styleguide which suggest using dash case for multiword props inside templates: https://vuejs.org/v2/style-guide/#Prop-name-casing-strongly-recommended.

I ran into this when an undefined value bound to a parent was overriding the same value explicitly set in the child, but it wasn't happening consistently depending on what the names of the props were.

Justineo:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/render-helpers/bind-object-props.js
+++ b/src/core/instance/render-helpers/bind-object-props.js
@@ -7,7 +7,8 @@ import {
   isObject,
   toObject,
   isReservedAttribute,
-  camelize
+  camelize,
+  hyphenate
 } from 'core/util/index'
 
 /**
@@ -45,12 +46,13 @@ export function bindObjectProps (
             : data.attrs || (data.attrs = {})
         }
         const camelizedKey = camelize(key)
-        if (!(key in hash) && !(camelizedKey in hash)) {
+        const hyphenatedKey = hyphenate(key)
+        if (!(camelizedKey in hash) && !(hyphenatedKey in hash)) {
           hash[key] = value[key]
 
           if (isSync) {
             const on = data.on || (data.on = {})
-            on[`update:${camelizedKey}`] = function ($event) {
+            on[`update:${key}`] = function ($event) {
               value[key] = $event
             }
           }
--- a/test/unit/features/directives/bind.spec.js
+++ b/test/unit/features/directives/bind.spec.js
@@ -234,7 +234,7 @@ describe('Directive v-bind', () => {
       template: `<test v-bind="test" data-foo="foo" dataBar="bar"/>`,
       components: {
         test: {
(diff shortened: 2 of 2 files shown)


posva:
Are you sure you are referencing the right issue on this one?

Justineo:
@posva Oops, wrong files. I’ll fix it ASAP.

test/unit/features/directives/bind.spec.js
```diff
@@ -234,7 +234,7 @@ describe('Directive v-bind', () => {
       template: `<test v-bind="test" data-foo="foo" dataBar="bar"/>`,
       components: {
         test: {
-          template: '<div :data-foo="dataFoo" :data-bar="dataBar"></div>',
+          template: '<div>{{ dataFoo }} {{ dataBar }}</div>',
```
Justineo: The old spec passed because `data-foo` fell through as an attribute and coincidentally appeared in the final output. Using plain text instead to ensure we got the correct value.

src/core/instance/render-helpers/bind-object-props.js
```diff
@@ -45,12 +46,13 @@ export function bindObjectProps (
             : data.attrs || (data.attrs = {})
         }
         const camelizedKey = camelize(key)
-        if (!(key in hash) && !(camelizedKey in hash)) {
+        const hyphenatedKey = hyphenate(key)
+        if (!(camelizedKey in hash) && !(hyphenatedKey in hash)) {
```
Justineo: We need to ensure both camelCased and kebab-cased keys are checked here.

Issue #9364, opened 2019-01-24
Transition not triggering @appear when used as the root of a component with appear attr


badz0:
### Version
2.5.22

### Reproduction link
[https://codepen.io/bogdan-muranets/pen/ErjOXy](https://codepen.io/bogdan-muranets/pen/ErjOXy)

### Steps to reproduce
Open repro, click on the "show" button and note that "animated on enter" text appeared.

### What is expected?
To see a text "animated on appear" which has to be added by `@appear` hook.

### What is actually happening?
The `@enter` hook is called instead of `@appear`.

---
Important is that `transition` is moved to separate component. Also, everything works as expected if `transition` isn't a root element in a separate component or the component is `functional`.

posva:
From my perspective, the behaviour is not consistent as the appear does trigger if the `transition` element is wrapped with a different element like a `div`

yyx990803:
closed via #9668

PR #9668, opened 2019-03-10
fix(transition): looking up `context` of transition

### LINKED ISSUE #9628: Transition component animates when wrapper is replaced
### Version
2.6.8

### Reproduction link
https://jsfiddle.net/shai/nujtes67/15/

### Steps to reproduce
Press the `Change screen` button to toggle between screen `1` and screen `2`.

### What is expected?
Nothing should fade, and nothing should be logged to console, because the `v-if` within the `<transition>` is always `true`.

On each press B fades in again and `enter: B (custom component)` gets logged to the console.

---
I've included A as well as B, to show that the bug only occurs when using a custom component with a `<transition>` and `<slot>`. A uses the `<transition>` directly, and doesn't exhibit the buggy behavior.

NB I've also noticed that if the custom component is switched to being functional instead, then it begins working correctly the same as A.

So it appears to only be reproducible when using a custom component and `<slot>`, even though this is what the docs recommend: https://vuejs.org/v2/guide/transitions.html#Reusable-Transitions

**Or, it's possible that I've misunderstood the expected behavior of a `<transition>` inside of a keyed or otherwise newly appearing component. In which case there's still a bug**, because in that case A and B should BOTH be fading and logging to console (though that seems weird to me: the `<transition>` isn't `appear=true`, so the fact that its parent is newly appearing shouldn't affect it…)

Shasharoman:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
fixed #9628 
Transition inside child component with v-if should be no enter-transition, it depend on appear attribute.

### DIFF
--- a/examples/modal/index.html
+++ b/examples/modal/index.html
@@ -10,7 +10,7 @@
   <body>
     <!-- template for the modal component -->
     <script type="text/x-template" id="modal-template">
-      <transition name="modal">
+      <transition name="modal" appear>
         <div class="modal-mask">
           <div class="modal-wrapper">
             <div class="modal-container">
--- a/src/platforms/web/runtime/modules/transition.js
+++ b/src/platforms/web/runtime/modules/transition.js
@@ -66,8 +66,8 @@ export function enter (vnode: VNodeWithData, toggleDisplay: ?() => void) {
   let context = activeInstance
   let transitionNode = activeInstance.$vnode
   while (transitionNode && transitionNode.parent) {
-    transitionNode = transitionNode.parent
     context = transitionNode.context
+    transitionNode = transitionNode.parent
   }
 
   const isAppear = !context._isMounted || !vnode.isRootInsert
--- a/test/unit/features/transition/transition.spec.js
+++ b/test/unit/features/transition/transition.spec.js
@@ -914,6 +914,84 @@ if (!isIE9) {
       }).then(done)
     })
 
+    it('transition inside nested child component with v-if', done => {
+      const vm = new Vue({
+        template: `
+          <div>
+            <foo v-if="ok" class="test"></foo>
+          </div>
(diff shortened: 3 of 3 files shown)


Shasharoman:
This pull request is also fixed #9364.

Shasharoman:
The test-e2e failed, I thought it was a problem with CircleCI, so I ignored it. Today I found that the test case was not updated. But also found some other hidden dangers, this PR does not seem to be compatible with the previous version, suppose there is a component of my-transition, his template is `<transition><div>xxx</div></transition>`, if used like this:

``` html
<my-transition v-if="showing"></my-transition>
```

When `showing` is changed from false to true, the previous version will have a transition effect, but the new PR will result in having to specify an appear attribute for the transition to have a transition effect.

katerlouis:
I disagree with that this change is not breaking things.
Yeah, the app builds etc.

But `enter` animations aren't called without applying `appear` inside a transition wrapping inside a component. Sure, just add `appear` – problem now is that the component's enter animation is triggered on first load also, which is not what I want.

See this issue: #10227

mrleblanc101:
@yyx990803 

Just found an example on the web why trying to understand how to use transition on root component.
I was wondering why the codepen example worked, but not in my project !
The example use Vue 2.6.4 which was before this patch was merged.

Vue 2.6.4: https://codepen.io/tahazsh/pen/mvKjbq
Vue 2.6.9+: https://codepen.io/mrleblanc/pen/eYgyZqR

I think the documentation should be improved on how to handle transition when they are on root element.
Currently if you do not specify `appear`, the leave transition work, but not the enter transition.

PR #9779, opened 2019-03-27
feat(v-model): add normalize modifier

bkuri:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] All tests are passing (I'm not 100% sure; **please see note below**)

- [x] A convincing reason for adding this feature

**Other information:**
Using `String.normalize()` is desirable when dealing with user-facing text inputs, considering that unicode strings that look the same on screen can sometimes be encoded differently and thus bring a bunch of problems later on (i.e. when trying to find duplicate strings).

Moreover, having to add `String.normalize()` every time is pretty much just as annoying as having to add `String.trim()` every time, so by that rationale I believe that adding this modifier to `v-model` makes just as much sense as having `trim` in there (which I very much appreciate!).

**Notes:**
- Sorry for not filing an issue first! I decided to just go for it since it's a pretty small modification. Hopefully this explanation is enough, but I can always add an issue later if needed.
- I'm pretty sure that I added all relevant tests and they all seem to have passed. I did experience an issue trying to run the ChromeHeadless instance on my PC, however, although I'm not sure if that test suite applies to this PR anyway (if it does please let me know and I'll try to get it fixed asap).
- I'm using the default `NFC` form of unicode normalization. It might be worth considering adding the other forms as well (`NFD`,`NFKC`,`NFKD`).

More informaton on `String.normalize()` and its usefulness can be found in the following links:
- [Why you need to normalize unicode strings](https://withblue.ink/2019/03/11/why-you-need-to-normalize-unicode-strings.html)
- [String.normalize() ­— MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize)

### DIFF
--- a/packages/vue-server-renderer/build.dev.js
+++ b/packages/vue-server-renderer/build.dev.js
@@ -3416,6 +3416,7 @@ function genComponentModel (
   modifiers
 ) {
   var ref = modifiers || {};
+  var normalize = ref.normalize;
   var number = ref.number;
   var trim = ref.trim;
 
@@ -3425,6 +3426,12 @@ function genComponentModel (
     valueExpression =
       "(typeof " + baseValueExpression + " === 'string'" +
       "? " + baseValueExpression + ".trim()" +
+      ": " + baseValueExpression + ")";
+  }
+  if (normalize) {
+    valueExpression =
+      "(typeof " + baseValueExpression + " === 'string'" +
+      "? " + baseValueExpression + ".normalize()" +
       ": " + baseValueExpression + ")";
   }
   if (number) {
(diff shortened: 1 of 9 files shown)


Justineo:
`String.prototype.normalize` is not supported by IE. Vue.js is compatible down to IE9.

bkuri:
Right. Totally forgot to check for IE compatibility. Bummer. :-1:

There's a [polyfill out there](https://github.com/walling/unorm), but I guess it's not worth the added weight, huh?

Thanks for your time regardless!

Justineo:
Thank you for your contribution anyway. 👍

PR #9884, opened 2019-04-17
feat: add warning when .native modifier is used on native HTML elements

NataliaTepluhina:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Currently if we have a `.native` modifier applied to event handler on native HTML element, it will break the handler (you can find the minimal reproduction [here](https://codesandbox.io/s/m718j536yp)).

We had a discussion about it with @chrisvfritz and it was decided it would be a good solution to add a console warning for this case. This way we can warn people they are using `.native` incorrectly: it should be used only on components not on elements.

An ideal solution would be also making `.native` work on elements while throwing a warning but it might be a subject for another PR.

### DIFF
--- a/src/core/vdom/create-element.js
+++ b/src/core/vdom/create-element.js
@@ -98,6 +98,12 @@ export function _createElement (
     ns = (context.$vnode && context.$vnode.ns) || config.getTagNamespace(tag)
     if (config.isReservedTag(tag)) {
       // platform built-in elements
+      if (process.env.NODE_ENV !== 'production' && isDef(data) && isDef(data.nativeOn)) {
+        warn(
+          `The .native modifier for v-on is only valid on components.`,
+          context
+        )
+      }
       vnode = new VNode(
         config.parsePlatformTagName(tag), data, children,
         undefined, undefined, context
--- a/src/core/vdom/create-element.js
+++ b/src/core/vdom/create-element.js
@@ -100,7 +100,7 @@ export function _createElement (
       // platform built-in elements
       if (process.env.NODE_ENV !== 'production' && isDef(data) && isDef(data.nativeOn)) {
         warn(
-          `The .native modifier for v-on is only valid on components.`,
+          `The .native modifier for v-on is only valid on components but it was used on <${tag}>.`,
           context
         )
       }
--- a/test/unit/features/directives/on.spec.js
+++ b/test/unit/features/directives/on.spec.js
@@ -470,7 +470,7 @@ describe('Directive v-on', () => {
     })
 
     triggerEvent(vm.$el, 'click')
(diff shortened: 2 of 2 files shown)


src/core/vdom/create-element.js
```diff
@@ -98,6 +98,12 @@ export function _createElement (
     ns = (context.$vnode && context.$vnode.ns) || config.getTagNamespace(tag)
     if (config.isReservedTag(tag)) {
       // platform built-in elements
+      if (process.env.NODE_ENV !== 'production' && isDef(data) && isDef(data.nativeOn)) {
+        warn(
+          `The .native modifier for v-on is only valid on components.`,
```
posva: I think we could add some extra information like the tag it was used on:

```suggestion
          `The .native modifier for v-on is only valid on components but it was used on <${tag}>.`,
```

test/unit/features/directives/on.spec.js
```diff
@@ -460,6 +460,20 @@ describe('Directive v-on', () => {
...
+    vm = new Vue({
+      el,
+      template: `
+        <button @click.native="foo"></button>
+      `,
+      methods: { foo: spy },
+    })
+
+    triggerEvent(vm.$el, 'click')
+    expect(`The .native modifier for v-on is only valid on components.`).toHaveBeenWarned()
```
posva:
```suggestion
    expect(`The .native modifier for v-on is only valid on components but it was used on <${tag}>.`).toHaveBeenWarned()
```

NataliaTepluhina: Agreed, more details is better :+1:

sdras:
lgtm!

test/unit/features/directives/on.spec.js
```diff
@@ -460,6 +460,20 @@ describe('Directive v-on', () => {
...
+    vm = new Vue({
+      el,
+      template: `
+        <button @click.native="foo"></button>
+      `,
+      methods: { foo: spy },
+    })
+
+    triggerEvent(vm.$el, 'click')
+    expect(`The .native modifier for v-on is only valid on components but it was used on <${tag}>.`).toHaveBeenWarned()
```
sirlancelot: Should the test assertion have `${tag}` replaced with `button`?

```suggestion
    expect(`The .native modifier for v-on is only valid on components but it was used on <button>.`).toHaveBeenWarned()
```

NataliaTepluhina: Thanks, you're totally right 👍

775976100:
很棒

PR #9912, opened 2019-04-23
fix(types): fix global namespace declaration for UMD bundle

ktsn:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

In the current declaration, `index.d.ts` has `export as namespace Vue` which exposes all exported types in `index.d.ts` in global namespace. So the user can access them by just writing like `Vue.ComponentOptions` without import statement.

But this declaration is actually not correct because the `Vue` constructor type is declared under `default` which is correct in ESModule bundle but UMD bundle.

```ts
new Vue({}) // This provides compilation error in current type.

new Vue.default({}) // No compilation error.
```

To declare Vue constructor as just `Vue` in global namespace, we need to write `export as namespace Vue` with a common js export. So I've add `umd.d.ts` and do that in that file.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,8 +1,7 @@
 import { Vue } from "./vue";
+import "./umd";
 
 export default Vue;
-
-export as namespace Vue;
 
 export {
   CreateElement,
--- a/types/test/tsconfig.json
+++ b/types/test/tsconfig.json
@@ -14,17 +14,9 @@
       "vue": ["../index.d.ts"]
     }
   },
-  "files": [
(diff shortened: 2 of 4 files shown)


yyx990803:
This should not affect existing usage with ES modules right?

ktsn:
Right, it should not affect. The entry point (`index.d.ts`) exports are the same and if a file is ES module, global `Vue` is not available.

PR #10154, opened 2019-06-14
polish: improve robustness of inject option

liximomo:
When a user provides a none-string object in inject array, it may fail at runtime because we try to make an implicit string conversion for the object in "Injection not found" warn. Objects like Symbol will fail at runtime because of such implicit conversion.

image

See the [reproduction](https://codepen.io/liximomo/pen/XLmZjX?editors=1012).

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/util/options.js
+++ b/src/core/util/options.js
@@ -338,6 +338,16 @@ function normalizeInject (options: Object, vm: ?Component) {
   const normalized = options.inject = {}
   if (Array.isArray(inject)) {
     for (let i = 0; i < inject.length; i++) {
+      if (typeof inject[i] !== 'string') {
+        if (process.env.NODE_ENV !== 'production') {
+          warn(
+            `Invalid value for option "inject": expected an Array of string, ` +
+              `but found ${toRawType(inject[i])}.`,
+            vm
+          );
+        }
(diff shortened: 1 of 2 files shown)


src/core/util/options.js
```diff
@@ -338,6 +338,16 @@ function normalizeInject (options: Object, vm: ?Component) {
...
+      if (typeof inject[i] !== 'string') {
+        if (process.env.NODE_ENV !== 'production') {
+          warn(
+            `Invalid value for option "inject": expected an Array of string, ` +
+              `but found ${toRawType(inject[i])}.`,
+            vm
+          );
+        }
+        continue;
+      }
```
Justineo: Let's refactor this like the following snippet (cache `inject[i]` ahead and reverse the if logic): 

https://github.com/vuejs/vue/blob/4d6e438d937dbe8f2a2fa0650855ff81ed9f62b9/src/core/util/options.js#L306-L312

So that we can get rid of the `continue` here.

PR #10199, opened 2019-06-26
Warn on colon shorthand usage on directive (fix #10191)

### LINKED ISSUE #10191: Warn if colon shorthand is used on v-if/v-html/etc.
I just spent way too long debugging something really weird until I realized I accidentally wrote `:v-if` instead of `v-if`. 

A warning when wrongly using shorthands like `:` on "native" vue attributes could prevent this bad experience easily. 

`:v-if="foo"`

--> console.warn("You specified v-bind:/ v-on: or a corresponding shorthand on a Vue attribute like v-if or similar. Usually this does not make sense.)

afontcu:
fix #10191

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

```html
<template>
  <div v-if="something">foo</div>  <!-- does not throw a warning -->
  <div :v-if="something">foo</div> <!-- throws a warning -->
</template>
```

There are several things I _decided_ that could be improved, so any advice or suggestion is greatly appreciated :)

- PR shows a warning for every bound attribute starting with `v-`. The reasoning is that "The v- prefix serves as a visual cue for identifying Vue-specific attributes in your templates." ([source](https://vuejs.org/v2/guide/syntax.html#Shorthands))
- Do we need more test cases?
- Is the warning message clear enough? If the attribute is not a valid Vue directive, should the message change?

Thanks!

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -30,6 +30,7 @@ export const forIteratorRE = /,([^,\}\]]*)(?:,([^,\}\]]*))?$/
 const stripParensRE = /^\(|\)$/g
 const dynamicArgRE = /^\[.*\]$/
 
+const colonDirRE = /^:v-/
 const argRE = /:(.*)$/
 export const bindRE = /^:|^\.|^v-bind:/
 const propBindRE = /^\./
@@ -761,6 +762,12 @@ function processAttrs (el) {
   for (i = 0, l = list.length; i < l; i++) {
     name = rawName = list[i].name
     value = list[i].value
+    // :v-if or similar
+    if (process.env.NODE_ENV !== 'production' && colonDirRE.test(name)) {
+      warn(
+        `A v-bind shorthand directive was used on another Vue directive. Did you want to write '${name.substr(1)}="${value}"'?`
+      )
+    }
     if (dirRE.test(name)) {
       // mark element as dynamic
       el.hasBindings = true
--- a/test/unit/modules/compiler/parser.spec.js
+++ b/test/unit/modules/compiler/parser.spec.js
@@ -530,6 +530,11 @@ describe('parser', () => {
     expect(ast.props[0].value).toBe('msg')
   })
 
+  it('v-bind expression on directive', () => {
+    parse('<div :v-if="foo"></div>', baseOptions)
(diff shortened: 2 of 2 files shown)


src/compiler/parser/index.js
```diff
@@ -761,6 +762,12 @@ function processAttrs (el) {
   for (i = 0, l = list.length; i < l; i++) {
     name = rawName = list[i].name
     value = list[i].value
+    // :v-if or similar
+    if (process.env.NODE_ENV !== 'production' && colonDirRE.test(name)) {
+      warn(
+        `A v-bind shorthand directive was used on another Vue directive. Did you want to write '${name.substr(1)}="${value}"'?`
```
posva: I think we could directly say which directive was used instead, something like:

```suggestion
        `A v-bind shorthand was used on another Vue Directive: ${name}. Did you meant '${name.substr(1)}="${value}"'?`
```

test/unit/modules/compiler/parser.spec.js
```diff
@@ -530,6 +530,11 @@ describe('parser', () => {
     expect(ast.props[0].value).toBe('msg')
   })
 
+  it('v-bind expression on directive', () => {
```
posva:
```suggestion
  it('warns when using v-bind shorthand on a directive', () => {
```

src/compiler/parser/index.js
afontcu: Makes sense! It might look a bit weird, though, given that `${name}` includes an initial colon:

> A v-bind shorthand was used on another Vue Directive: :v-if. Did you mean 'v-if="foo"'

what about going full explicit?

> A v-bind shorthand was used on a v-if directive. Did you mean v-if="foo", without a colon?




Thanks for the feedback! 🙌

posva: you are right, let's be even more explicit, I updated the suggestion and fixed the typos too

afontcu: Done! Didn't commit the suggestion since I needed to update the test as well 👍

afontcu:
np! Actually I was about to add that bit too 😂 but after not seeing it in the suggestion I restrained myself. Updated!

Vannsl:
Can't this PR be merged? :)

PR #10202, opened 2019-06-27
feat(events): add .emit event modifier (fix #9325)

### LINKED ISSUE #9325: Modifier to propagate events to parent
Currently, and as far as I know, if we want to propagate an event fired by a child component to the parent (the child's grandparent) we need to $emit the event again, and we need to pass all the arguments one more time. This can become a problem, for example, if the event has a variable number of arguments because we need to specify them manually or pass the whole array as a new argument.

The current way would be something like
```
@blur="$emit('blur')"
@create="$emit('create', arguments[0])"
@input="$emit('input', arguments[0], arguments[1])"
```

```
@blur.propagate
@create.propagate
@input.propagate
```
And if we want to both handle the event and propagate it to the parent, we would use
```
@input.propagate="someFunction"
```
----
Maybe .propagate isn't the best term, since it could be confused with .stopPropagation(), but I couldn't come out with anything better.

Vontus:
fix #9325

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/compiler/codegen/events.js
+++ b/src/compiler/codegen/events.js
@@ -42,6 +42,7 @@ const genGuard = condition => `if(${condition})return null;`
 const modifierCode: { [key: string]: string } = {
   stop: '$event.stopPropagation();',
   prevent: '$event.preventDefault();',
+  emit: '$emit(arguments.callee.name, ...arguments)',
   self: genGuard(`$event.target !== $event.currentTarget`),
   ctrl: genGuard(`!$event.ctrlKey`),
   shift: genGuard(`!$event.shiftKey`),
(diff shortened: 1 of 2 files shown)


danielcmm:
Hi! Has this been released already?

LinusBorg:
As this PR is still open and not merged: no. We're also hesitant to add new features to Vue 2 at this stage, and haven't discussed this specific proposal in any way.

IT should likely have gone through the RFC process in any case.

PR #10217, opened 2019-07-02
fix(warn): regular slot and scoped slot with same name shouldn't be allowed(fix #10149)

### LINKED ISSUE #10149: one scoped slot would render twice in some case
### Version
2.6.10

### Reproduction link
[https://codepen.io/lee88688/pen/jjPpBm?editors=1010](https://codepen.io/lee88688/pen/jjPpBm?editors=1010)

### Steps to reproduce
as seen in codepen,  "hello" component have only one slot. but rendered 2 div tag at last. I have a look at Vue's code [renderSlot ](https://github.com/vuejs/vue/blob/530ca1b2db315fbd0e360807b2031d26665c5d3d/src/core/instance/render-helpers/render-slot.js#L8) which seems renderSlot function would treat the normal slot as scoped slot. why it would be like this?

### What is expected?
render just one slot.

render one slot twice.

Kermit-Xuan:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/compiler/parser/index.js
+++ b/src/compiler/parser/index.js
@@ -201,6 +201,35 @@ export function parse (
     }
   }
 
+  function checkSlotConstraints (el: any, scopedSlots) {
+    scopedSlots = scopedSlots || {}
+    const children = el.children || []
+    if (!children.length) return
+    for (let i = 0; i < children.length; i++) {
+      const child = children[i]
+      if (child.tag !== 'slot') {
+        checkSlotConstraints(child, scopedSlots)
+        continue
+      }
+      const slotName = child.slotName || 'default'
+      const isScopedSlot = (child.attrs || []).concat(child.dynamicAttrs || []).map(attr => ({
+        name: camelize(attr.name),
+        value: attr.value,
+        dynamic: attr.dynamic
+      })).length > 0
+      if (isScopedSlot) {
+        scopedSlots[slotName] = true
+        continue
+      } else if (scopedSlots[slotName]) {
+        warnOnce(
+          'regular slot and scoped slot shouldn\'t have the same name: ' + slotName,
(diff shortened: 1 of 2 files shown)


src/compiler/parser/index.js
```diff
@@ -201,6 +201,33 @@ export function parse (
...
+      const slotName = child.slotName || 'default'
+      const isScopedSlot = (child.attrs || []).concat(child.dynamicAttrs || []).map(attr => ({
+        name: camelize(attr.name),
+        value: attr.value,
+        dynamic: attr.dynamic
+      })).length > 0
+      if (isScopedSlot) {
+        scopedSlots[slotName] = true
+        continue
+      } else if (process.env.NODE_ENV !== 'production' && scopedSlots[slotName]) {
```
posva: Even though the check is at compiler level, I think it will make sense to wrap the check so it only gets executed on non-prod mod

src/compiler/parser/index.js
```diff
@@ -306,6 +333,10 @@ export function parse (
         element.end = end
       }
       closeElement(element)
+      // check slot constraints
+      if (!stack.length && element === root) {
+        checkSlotConstraints(element, {})
```
posva: maybe the function could do the check and the warning could be executed here.
It would be better if the warning could also show some information about the Component where the parsing error was found.
Maybe using `warnOnce` is also a good idea

Kermit-Xuan: OK, I resolved it.

Kermit-Xuan:
@posva

PR #10229, opened 2019-07-04
fix(slot): add a function to return the slot fallback content

### LINKED ISSUE #10224: Default slot content should not be evaluated when the parent is providing it
### Version
2.6.10

### Reproduction link
[https://codesandbox.io/s/kk445knpx3](https://codesandbox.io/s/kk445knpx3)

### Steps to reproduce
In the `Child.vue` we have a slot with a default slot content. In the `Parent.vue` we override the default content of the child. 

### What is expected?
IMO the content of the slot in child component should not be evaluated since the content provided from the parent will be used.

The content of the parent it renders which is what I expect.

The default content of the `Child.vue` it's evaluated even if it not renders. That's why you see in the console: `Cannot read property 'length' of null`

---
Maybe that's not a bug and instead it's the intentional behavior, but it doesn't make sense to me, to evaluate something that will not render.

zrh122:
fixes #10224.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/render-helpers/render-slot.js
+++ b/src/core/instance/render-helpers/render-slot.js
@@ -7,7 +7,7 @@ import { extend, warn, isObject } from 'core/util/index'
  */
 export function renderSlot (
   name: string,
-  fallback: ?Array<VNode>,
+  fallbackRender: ?() => Array<VNode>,
   props: ?Object,
   bindObject: ?Object
 ): ?Array<VNode> {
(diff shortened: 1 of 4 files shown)


src/core/instance/render-helpers/render-slot.js
```diff
@@ -24,9 +24,9 @@ export function renderSlot (
       }
       props = extend(extend({}, bindObject), props)
     }
-    nodes = scopedSlotFn(props) || fallback
```
yyx990803, line 27: We need to account for already render vnodes here (from component compiled with an older version of Vue)

yyx990803 (changes requested):

posva:
I added the commit for retro compatibility at https://github.com/vuejs/vue/pull/12014 because I couldn't push to the branch

PR #10245, opened 2019-07-08
feat(keep-alive): includes/excludes keep-alive based on key and name (fix #8028)

### LINKED ISSUE #8028: keep-alive: include/exclude components by component key attribute
The include and exclude props allow components to be conditionally cached only by component name. If we want to reuse components but force replacement using the `key` attribute, there is no control over which components we want to keep-alive only matching components by their name.

https://jsfiddle.net/9nk92wuy/

RobbinBaauw:
The keep-alive component will now also use the key of the component when checking the includes or excludes properties.

On this way you can cache specific instances of a component instead of caching all instances of that component.

fix #8028

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

### DIFF
--- a/src/core/components/keep-alive.js
+++ b/src/core/components/keep-alive.js
@@ -9,16 +9,20 @@ function getComponentName (opts: ?VNodeComponentOptions): ?string {
   return opts && (opts.Ctor.options.name || opts.tag)
 }
 
-function matches (pattern: string | RegExp | Array<string>, name: string): boolean {
-  if (Array.isArray(pattern)) {
-    return pattern.indexOf(name) > -1
-  } else if (typeof pattern === 'string') {
-    return pattern.split(',').indexOf(name) > -1
-  } else if (isRegExp(pattern)) {
-    return pattern.test(name)
+function matches (pattern: string | RegExp | Array<string>, key: ?string, name: string): boolean {
+  function matchesValue(value: string) {
+    if (Array.isArray(pattern)) {
+      return pattern.indexOf(value) > -1
+    } else if (typeof pattern === 'string') {
+      return pattern.split(',').indexOf(value) > -1
+    } else if (isRegExp(pattern)) {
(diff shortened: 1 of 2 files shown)


qtttttttttting:
How to use it?Any example?

RobbinBaauw:
I've made a JSFiddle: https://jsfiddle.net/87t53pwd/18/. 

A small explanation: In the keep-alive block we have 3 usages of the same component with each a different state. We only want to keep the state cached of the first 2, and not the last.

The current version of keep-alive doesn't allow for this (because it only uses the name of the component, not the key). Now, it also looks at the key (so if either the name or the key matches something in the includes array, it caches it).

You can replace the newly built version of Vue by the old one and see the difference (it works vs it doesn't)

qtttttttttting:
Thank you for your answer.
But I install vue via git://github.com/vuejs/vue.git#dev, but i seems not to work .

RobbinBaauw:
Yes, because this pull request has not been merged yet. If you try dev on my fork it should work

qtttttttttting:
I have urgent need for this function in my project. When will it be merged?
Thank you .

RobbinBaauw:
I can't merge it, the vue authors need to do that. You can use that branch if you want, you'll see here when / if it will be merged

qtttttttttting:
OK. Thank you so much.

qtttttttttting:
I try to install git://github.com/RobbinBaauw/vue.git#dev, but it seems not work with vue-router

RobbinBaauw:
Can you make a fiddle? In the project I use it for it works, also with vue-router.

Also, how did you use it? You should build vue manually as well as its dist files are not yet updated.

qtttttttttting:
Sorry, it is my mistake.It works now.Thank you .

PR #10314, opened 2019-07-25
chore: remove computed cache:false option

HusamIbrahim:
As mentioned a while ago in another PR this option is deprecated in 2.x. If the maintainers have reservations against merging this PR then I think that at least we should provide a warning when _cache:false_ is used.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: remove deprecated feature

- [x] Yes
- [ ] No

Caching invalidation of computed properties will stop working in applications that use this feature. But as explained in the [migration guide](https://vuejs.org/v2/guide/migration.html#cache-false-deprecated) these applications should use normal methods instead, which should have the same result. 

**Other information:**

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -220,7 +220,7 @@ export function defineComputed (
     sharedPropertyDefinition.set = noop
   } else {
     sharedPropertyDefinition.get = userDef.get
-      ? shouldCache && userDef.cache !== false
+      ? shouldCache
         ? createComputedGetter(key)
         : createGetterInvoker(userDef.get)
       : noop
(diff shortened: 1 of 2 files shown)


LinusBorg:
Why?

HusamIbrahim:
@LinusBorg My understanding is that [this option is deprecated](https://vuejs.org/v2/guide/migration.html#cache-false-deprecated). The migration guide recommends using methods instead. Do you think a warning message is better than removing the option altogether?

Justineo:
“Deprecated” is different from “Removed” in the migration guide. Deprecated means it's still there but it's no longer recommended to use.

HusamIbrahim:
@Justineo Aha! That makes things clearer. Do you think this is deserving of a warning message then?

PR #10414, opened 2019-08-19
fix: add URL in ssr sandbox global

clarkdo:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Related to https://github.com/nuxt/nuxt.js/issues/4914, when using `runInNewContext: true` and URL in server side, the URL is not injected on global object.

### DIFF
--- a/src/server/bundle-renderer/create-bundle-runner.js
+++ b/src/server/bundle-renderer/create-bundle-runner.js
@@ -9,6 +9,7 @@ function createSandbox (context) {
   const sandbox = {
     Buffer,
     URL,
+    URLSearchParams:  global.URLSearchParams,
     console,
     process,
     setTimeout,
--- a/src/server/bundle-renderer/create-bundle-runner.js
+++ b/src/server/bundle-renderer/create-bundle-runner.js
@@ -8,6 +8,7 @@ const NativeModule = require('module')
 function createSandbox (context) {
   const sandbox = {
     Buffer,
+    URL,
     console,
     process,
     setTimeout,


pimlie:
Maybe we could add [`URLSearchParams`](https://nodejs.org/dist/latest-v12.x/docs/api/globals.html) as well?

Also please note that these classes are only available since Node v10, might be an issue.

allanchua101:
Looks good to me 👍

Atinux:
Any news about it?

pi0:
> Also please note that these classes are only available since Node v10, might be an issue.

I think so as well. Maybe we should either drop node 8 support or check if `URL` is not undefined and otherwise fallback to https://www.npmjs.com/package/whatwg-url?

posva:
@pi0 do you still need this one to go forward? If yes, we would have to make sure it works with node 8 since Vue 2 still supports that version of node even though it is unmaintained

pi0:
@posva I think we can close it to avoid breaking changes and let userland to polyfill. We can also conditionally add to context only if node.js version supports it like:

```js
URL: typeof URL === 'undefined'  ? undefined : URL
```

posva:
Okay, thanks a lot!

lzxb:
Has this been released?
https://github.com/vuejs/vue/blob/dev/src/server/bundle-renderer/create-bundle-runner.js#L9

I see the source code of this address. I don't see the merge

PR #10419, opened 2019-08-21
feat(core): allow users to create a Vue-instance without the "new" ke…

hassan-nteifeh:
Hi!

I believe that **not** enforcing the usage of the 'new' keyword when creating a vue-instance would be a tiny but good improvement for the code. All the tests are passing except for the one that goes against the purpose of of the code change, therefore i had to delete it.

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/instance/index.js
+++ b/src/core/instance/index.js
@@ -3,13 +3,12 @@ import { stateMixin } from './state'
 import { renderMixin } from './render'
 import { eventsMixin } from './events'
 import { lifecycleMixin } from './lifecycle'
-import { warn } from '../util/index'
 
 function Vue (options) {
   if (process.env.NODE_ENV !== 'production' &&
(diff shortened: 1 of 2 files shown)


posva:
Thanks for the proposal but Vue needs to be created as an instance with `new`. It also makes sense to instantiate it with the `new` operator as the variable name starts with an uppercase letter.

Calling just `Vue(option)` won't yield the desired result. Specially when calling it multiple times on the same application, so the warning there is intentional.

hassan-nteifeh:
@posva Yes but the pull request does include a code that makes sure that it gets called with the 'new' keyword in case it was omitted. Am i missing something?

PR #10459, opened 2019-08-30
bail out of event blocking for Adobe CEP bug (fix #10366)

### LINKED ISSUE #10366: Click Event Triggers on Complex Buttons are ignored in some environments
### Version
2.6.10

### Reproduction link
[https://jsfiddle.net/1d8nbv7h/2/](https://jsfiddle.net/1d8nbv7h/2/)

### Steps to reproduce
1. Configure one of the Adobe CEP [Sample Panels](https://github.com/Adobe-CEP/Samples). The [PProPanel](https://github.com/Adobe-CEP/Samples/tree/master/PProPanel) is a good starting point as it has very clear documentation on how to [set up the environment](https://github.com/Adobe-CEP/Samples/tree/master/PProPanel#2-enable-loading-of-unsigned-panels) for testing.
1. Replace the HTML/JavaScript/CSS contents of the panel project with the contents of the [linked JSFiddle](https://jsfiddle.net/1d8nbv7h/2/).
1. Open the panel.
1. Attach a debugger (with the default PProPanel setup this would mean browsing to `localhost:7777` in Chrome).
1. Set the "_Mouse > `click`_" **Event Listener Breakpoint** in the "Sources" tab of the Chrome Debugger.
1. Click the Vue icon in the center of the silver `div`.

### What is expected?
Method bound to the `@click` handler is triggered when the image embedded in the parent div is clicked. 

The method bound to the `@click` handler is _only_ triggered when clicking outside of the parent div.

---
This is a non-trivial bug to reproduce as the only place I've experienced it is in [Adobe CEP](https://github.com/Adobe-CEP/CEP-Resources/blob/master/CEP_9.x/Documentation/CEP%209.0%20HTML%20Extension%20Cookbook.md) extensions (which run NW.js under the hood). That said, it does reproduce 100% of the time there.

The debugger (CEP context) seems to show several funny things at around [this line](https://github.com/vuejs/vue/blob/d40b7ddb8177944d1dd50f4f780e6fd92c9455c2/src/platforms/web/runtime/modules/events.js#L69) in the Vue events.js file. Specifically:

1. The `e.timeStamp` value **does not change** between callbacks for _different buttons/elements_.
1. The `attachedTimestamp` is **_significantly_ larger** than the `e.timeStamp` value.
1. The `attachedTimestamp` value _does_ change when the component is updated (the `e.timeStamp` remains identical).

I should note that this affects at least [CEP 8.0 and CEP 9.0](https://github.com/Adobe-CEP/CEP-Resources/blob/master/CEP_9.x/Documentation/CEP%209.0%20HTML%20Extension%20Cookbook.md#chromium-embedded-framework-cef) (tested in Premiere Pro).

**Vue Versions Note:** This broke somewhere between versions 2.5.17 and 2.6.x. If we run a version of 2.5 (2.5.17 and some earlier versions verified), then this issue does not occur. In testing 2.6.x, we've found that this same issue occurs from 2.6.5 to 2.6.10 (latest). Versions of 2.6 prior to 2.6.5 are actually _worse_ in that the buttons basically don't work at all.

**Important Note:** I should _further_ note that apparently right-clicking to open the basic CEF [not CEP] context menu will cause the `e.timeStamp` values to begin reporting as expected. Once this occurs, the buttons will _also_ work as expected. That said, we shouldn't have to instruct users to right-click prior to interfacing with the extension.

ericdrobinson:
fix #10366

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Some [Adobe CEP](https://github.com/Adobe-CEP) [environments](https://github.com/Adobe-CEP/CEP-Resources/blob/master/CEP_9.x/Documentation/CEP%209.0%20HTML%20Extension%20Cookbook.md#applications-integrated-with-cep) have a broken `Event.timeStamp` implementation that breaks event blocking logic. The issue specifically affects host applications running on macOS with CEP version 9.3 and below. This workaround is OS-agnostic so as to maintain identical behavior across platforms. More details can be found in issue #10366.

### DIFF
--- a/src/core/util/env.js
+++ b/src/core/util/env.js
@@ -1,5 +1,7 @@
 /* @flow */
 
+import { isObject } from 'shared/util'
+
 // can we use __proto__?
 export const hasProto = '__proto__' in {}
 
@@ -16,7 +18,7 @@ export const isIOS = (UA && /iphone|ipad|ipod|ios/.test(UA)) || (weexPlatform ==
 export const isChrome = UA && /chrome\/\d+/.test(UA) && !isEdge
 export const isPhantomJS = UA && /phantomjs/.test(UA)
 export const isFF = UA && UA.match(/firefox\/(\d+)/)
-export const isCEP = inBrowser && window.__adobe_cep__ !== undefined
+export const isCEP = inBrowser && isObject(window.__adobe_cep__)
 
 // Firefox has a "watch" function on Object.prototype...
 export const nativeWatch = ({}).watch
--- a/src/core/util/env.js
+++ b/src/core/util/env.js
@@ -16,6 +16,7 @@ export const isIOS = (UA && /iphone|ipad|ipod|ios/.test(UA)) || (weexPlatform ==
 export const isChrome = UA && /chrome\/\d+/.test(UA) && !isEdge
 export const isPhantomJS = UA && /phantomjs/.test(UA)
 export const isFF = UA && UA.match(/firefox\/(\d+)/)
+export const isCEP = inBrowser && window.__adobe_cep__ !== undefined
 
 // Firefox has a "watch" function on Object.prototype...
 export const nativeWatch = ({}).watch
--- a/src/platforms/web/runtime/modules/events.js
+++ b/src/platforms/web/runtime/modules/events.js
@@ -2,7 +2,7 @@
 
 import { isDef, isUndef } from 'shared/util'
 import { updateListeners } from 'core/vdom/helpers/index'
-import { isIE, isFF, supportsPassive, isUsingMicroTask } from 'core/util/index'
+import { isIE, isFF, isCEP, supportsPassive, isUsingMicroTask } from 'core/util/index'
 import { RANGE_TOKEN, CHECKBOX_RADIO_TOKEN } from 'web/compiler/directives/model'
 import { currentFlushTimestamp } from 'core/observer/scheduler'
 
@@ -44,6 +44,14 @@ function createOnceHandler (event, handler, capture) {
 // safe to exclude.
 const useMicrotaskFix = isUsingMicroTask && !(isFF && Number(isFF[1]) <= 53)
 
+// #10366: CEP <= 9.3.x has a buggy Event.timeStamp implementation. While the
+// issue is restricted to macOS, the fix is OS-agnostic to keep behavioral
+// differences to a minimum.
+const isCEP93orEarlier = isCEP && ((maxBadMajor, maxBadMinor) => {
+  const version = JSON.parse(window.__adobe_cep__.getCurrentApiVersion())
+  return version.major <= maxBadMajor && version.minor <= maxBadMinor
+})(9, 3)
+
 function add (
   name: string,
   handler: Function,
@@ -71,6 +79,9 @@ function add (
         // #9462 iOS 9 bug: event.timeStamp is 0 after history.pushState
         // #9681 QtWebEngine event.timeStamp is negative value
         e.timeStamp <= 0 ||
+        // #10366 Adobe CEP bug: event.timeStamp is not reliable on macOS for
+        // host applications with CEP versions prior to 9.4.x.
+        isCEP93orEarlier ||
         // #9448 bail if event is fired in another document in a multi-page
         // electron/nw.js app, since event.timeStamp will be using a different
         // starting reference
--- a/src/platforms/web/runtime/modules/events.js
+++ b/src/platforms/web/runtime/modules/events.js
@@ -47,10 +47,12 @@ const useMicrotaskFix = isUsingMicroTask && !(isFF && Number(isFF[1]) <= 53)
 // #10366: CEP <= 9.3.x has a buggy Event.timeStamp implementation. While the
 // issue is restricted to macOS, the fix is OS-agnostic to keep behavioral
 // differences to a minimum.
-const isCEP93orEarlier = isCEP && ((maxBadMajor, maxBadMinor) => {
+const isCEP93orEarlier = isCEP &&
+      (typeof window.__adobe_cep__.getCurrentApiVersion !== 'function' ||
+      ((maxBadMajor, maxBadMinor) => {
   const version = JSON.parse(window.__adobe_cep__.getCurrentApiVersion())
   return version.major <= maxBadMajor && version.minor <= maxBadMinor
-})(9, 3)
+})(9, 3))
 
 function add (
   name: string,


derrickb:
One note:
`var isCEP = inBrowser && window.__adobe_cep__ !== 'undefined';`
I believe you're missing `typeof`, or change `'undefined'` to `undefined`.

src/core/util/env.js
```diff
@@ -16,6 +16,7 @@ export const isIOS = (UA && /iphone|ipad|ipod|ios/.test(UA)) || (weexPlatform ==
 export const isChrome = UA && /chrome\/\d+/.test(UA) && !isEdge
 export const isPhantomJS = UA && /phantomjs/.test(UA)
 export const isFF = UA && UA.match(/firefox\/(\d+)/)
+export const isCEP = inBrowser && window.__adobe_cep__ !== 'undefined'
```
derrickb: `undefined` shouldn't be a string here.

ericdrobinson: Good catch! I'll get that cleaned up!

ericdrobinson:
@derrickb Fixed! Thanks for catching that!! 😅

src/core/util/env.js
```diff
@@ -16,6 +16,7 @@ export const isIOS = (UA && /iphone|ipad|ipod|ios/.test(UA)) || (weexPlatform ==
 export const isChrome = UA && /chrome\/\d+/.test(UA) && !isEdge
 export const isPhantomJS = UA && /phantomjs/.test(UA)
 export const isFF = UA && UA.match(/firefox\/(\d+)/)
+export const isCEP = inBrowser && window.__adobe_cep__ !== undefined
```
posva: I think checking `window.__adobe_cep__` is enough since we know `window` is defined

src/platforms/web/runtime/modules/events.js
```diff
@@ -44,6 +44,14 @@ function createOnceHandler (event, handler, capture) {
 // safe to exclude.
 const useMicrotaskFix = isUsingMicroTask && !(isFF && Number(isFF[1]) <= 53)
 
+// #10366: CEP <= 9.3.x has a buggy Event.timeStamp implementation. While the
+// issue is restricted to macOS, the fix is OS-agnostic to keep behavioral
+// differences to a minimum.
+const isCEP93orEarlier = isCEP && ((maxBadMajor, maxBadMinor) => {
+  const version = JSON.parse(window.__adobe_cep__.getCurrentApiVersion())
```
posva: are we sure this function exist in **all** versions of adobe_cep?

posva: If it exist in all version till today, I think we should check if the function exists and if it doesn't return false to prevent breaking changes with adobe_cep api in the future

derrickb: It has existed in all CEP versions since its inception in [version 4.x ](https://github.com/Adobe-CEP/CEP-Resources/blob/5880e20b7cb2dc09e6b226f15a33165a08c0d831/CEP_4.x/CSInterface.js#L812)

To my knowledge, Adobe has never removed functions from their CEP API, even deprecated ones. However, a check for the function is still a good idea.

src/core/util/env.js
ericdrobinson: Can you clarify? Which of the following do you mean:

1. ```js
    export const isCEP = window.__adobe_cep__ !== undefined
    ```
2. ```js
    export const isCEP = inBrowser && window.__adobe_cep__
    ```

The first would cause issues where `window` is not defined (i.e. not in `inBrowser` contexts) while the second seems a bit ambiguous and doesn't follow the `undefined` check style of other options.

Or did you mean something  else?

src/platforms/web/runtime/modules/events.js
ericdrobinson: >are we sure this function exist in all versions of adobe_cep?

Not 100%, no. The table shows that the earliest version of  **HTML**-based CEP was "CEP 4". Adobe has an extraordinarily bad habit of not publishing details of the CEP minor versions for their apps so it is entirely possible that "version 4" means "version 4.2." That said, the opposite is also possible.

>If it exist in all version till today, I think we should check if the function exists and if it doesn't return false to prevent breaking changes with adobe_cep api in the future

@bbb999 might be able to shed more light on this but my understanding is that the upcoming CEP 9.4.x is likely to be the last version of CEP. Adobe is _beginning_ to transition their apps to a new technology (known as UXP) and are starting to phase out CEP. There may be bugfix releases but I'd be willing to bet that we won't see any core API changes before CEP's final death knell.

>However, a check for the function is still a good idea.

Agreed. I can make that change for sure.

bbb999: Adobe has made no announcements regarding [above], but Eric's speculation is entirely plausible. :)

ericdrobinson: @bbb999 I'm going to add a check for the existence of the `getCurrentApiVersion` function. It's non-existence, however, can mean one of two things:

1. This is an archaic version of CEP that pre-dates that function.
2. This is a future version of CEP that redefined that function as something else.

Given the trajectory of CEP over its lifetime, it seems that option 2 is far, far less likely than option 1. Would it be safe to assume, then, that if:

```js
typeof __adobe_cep__.getCurrentApiVersion !== 'function'
```

then we're more likely to be working with an archaic version of CEP than an as-yet-unreleased one?

ericdrobinson: Spoke with @bbb999 in another forum (Adobe Devs community Slack) and the response was:

>[Option 1], definitely.

src/core/util/env.js
ericdrobinson: I spoke with @posva in a separate forum (Discord direct chat) and it was determined that using the `isObject` utility function was the best approach to validating `window.__adobe_cep__`.

ericdrobinson:
@posva Fixes applied.  Please let me know if there is anything further!

ericdrobinson:
For the record, [the decision](https://github.com/vuejs/vue/pull/10459#discussion_r322929456) to assume that:

   >`__adobe_cep__.getCurrentApiVersion === undefined` => super old CEP version

was correct.

I just ran across this statement in [the CEP Cookbook](https://github.com/Adobe-CEP/CEP-Resources/blob/master/CEP_9.x/Documentation/CEP%209.0%20HTML%20Extension%20Cookbook.md#api-version):

>Please note [that the `getCurrentApiVersion`] API itself is available only since 4.2.0. If you get an error saying `getCurrentApiVersion` is undefined, then you are running in CEP 4.0 or 4.1. Otherwise, the value returned will tell you the version of CEP integrated by the Adobe product.

Awesome. 💪

posva (approved):

bbb999 (approved):

PR #10491, opened 2019-09-06
feat(web): Integrate trusted types into Vue

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG , with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

### Integrating Trusted Types to Vue
The integration is fairly simple as there are not many places which break when using Vue with Trusted Types. There one place though, which I am not sure if it's even called (couldn't trigger it myself). It's marked in a code with `TT_TODO: ...` comment. I hope someone can clarify this for me. **The PR doesn't introduce any breaking changes**. All of the code is applied only when the trustedTypes API is available in the global object.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The PR adds support for Trusted Types only on client side, but it might be beneficial to use them also on server side. There is no DOM on server side, however we can check if the values are trusted before creating markup with dangerous attributes. This also makes the applications render the same thing on the server and on client. There is already a similar [PR for this for React](https://github.com/facebook/react/pull/16555).

If you would like to try this change there is a [really simple application](https://github.com/Siegrift/vue-example-app) with the instructions how to setup vue locally in which you can play with Trusted Types. Vue cli uses webpack for development which provides features like hot-reloading. This currently breaks, but the [PR is already fired and we are waiting for review](https://github.com/webpack/webpack/pull/9426).

### DIFF
--- a/src/platforms/web/runtime/modules/dom-props.js
+++ b/src/platforms/web/runtime/modules/dom-props.js
@@ -2,7 +2,7 @@
 
 import { isDef, isUndef, extend, toNumber } from 'shared/util'
 import { isSVG } from 'web/util/index'
-import {convertToTrustedType} from 'web/security'
+import {maybeCreateDangerousSvgHTML} from 'web/security'
 
 let svgContainer
 
@@ -53,7 +53,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
     } else if (key === 'innerHTML' && isSVG(elm.tagName) && isUndef(elm.innerHTML)) {
       // IE doesn't support innerHTML for SVG elements
       svgContainer = svgContainer || document.createElement('div')
-      svgContainer.innerHTML = convertToTrustedType(`<svg>${cur}</svg>`)
+      svgContainer.innerHTML = maybeCreateDangerousSvgHTML(cur)
       const svg = svgContainer.firstChild
       while (elm.firstChild) {
         elm.removeChild(elm.firstChild)
--- a/src/platforms/web/runtime/modules/dom-props.js
+++ b/src/platforms/web/runtime/modules/dom-props.js
@@ -2,6 +2,7 @@
 
 import { isDef, isUndef, extend, toNumber } from 'shared/util'
 import { isSVG } from 'web/util/index'
+import {convertToTrustedType} from 'web/security'
 
 let svgContainer
 
@@ -20,6 +21,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
 
   for (key in oldProps) {
     if (!(key in props)) {
+      // TT_TODO: when (how) is this even called
       elm[key] = ''
     }
   }
@@ -51,7 +53,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
     } else if (key === 'innerHTML' && isSVG(elm.tagName) && isUndef(elm.innerHTML)) {
       // IE doesn't support innerHTML for SVG elements
       svgContainer = svgContainer || document.createElement('div')
-      svgContainer.innerHTML = `<svg>${cur}</svg>`
+      svgContainer.innerHTML = convertToTrustedType(`<svg>${cur}</svg>`)
       const svg = svgContainer.firstChild
       while (elm.firstChild) {
         elm.removeChild(elm.firstChild)
--- a/src/platforms/web/security.js
+++ b/src/platforms/web/security.js
@@ -8,24 +8,50 @@ type TrustedTypePolicy = {
 };
 
 let policy: ?TrustedTypePolicy
-// we need this function to clear the policy in tests
-Vue.prototype.$clearTrustedTypesPolicy = function() {
-  policy = undefined
-}
-
-export function maybeCreateDangerousSvgHTML(value: any): string {
-  // create policy lazily to simplify testing
+// create policy lazily to simplify testing
+function getOrCreatePolicy() {
   const tt = getTrustedTypes()
   if (tt && !policy) {
     policy = tt.createPolicy(Vue.config.trustedTypesPolicyName, {createHTML: (s) => s});
   }
 
-  if (!tt) return `<svg>${value}</svg>`;
-  else if (!isTrustedValue(value)) throw new Error('Expected svg innerHTML to be TrustedHTML!');
-  // flow complains 'policy' may be undefined
-  else return (policy: any).createHTML(`<svg>${value}</svg>`);
+  return policy
+}
+
+if (process.env.NODE_ENV !== 'production') {
+  // we need this function to clear the policy in tests
+  Vue.prototype.$clearTrustedTypesPolicy = function() {
+    policy = undefined
+  }
+}
+
+// TODO: remove once https://github.com/WICG/trusted-types/issues/36 is resolved.
+function isApplicationUsingTrustedTypes() {
+  // This check actually checks whether Trusted Types are enforced or not.
+  // However application might still use them in report only mode.
+  // This has also a side effect that CSP violation will be triggered.
+  try {
+    document.createElement('div').innerHTML = 'string';
+    return false;
+  } catch (e) {
+    return true;
+  }
+}
+
+export function maybeCreateDangerousSvgHTML(value: any): string {
+  const tt = getTrustedTypes()
+
+  if (!tt || !isApplicationUsingTrustedTypes()) return `<svg>${value}</svg>`;
+  else if (!isTrustedValue(value)) {
+    throw new Error('Expected svg innerHTML to be TrustedHTML!');
+  }
+  // flow complains that 'getOrCreatePolicy()' might return null.
+  else return (getOrCreatePolicy(): any).createHTML(`<svg>${value}</svg>`);
 }
 
 export function getTrustedShouldDecodeInnerHTML(href: boolean): string {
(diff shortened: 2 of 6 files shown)


Siegrift:
@yyx990803 Are the broken tests caused by my feature? They seems pretty unrelated...

src/platforms/web/runtime/modules/dom-props.js
```diff
@@ -20,6 +21,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
 
   for (key in oldProps) {
     if (!(key in props)) {
+      // TT_TODO: when (how) is this even called
```
yyx990803: This line is called when users use object bindings, e.g. `v-bind="obj"`, and the new object has different keys from the old object. For example, on first render `obj` could be `{ id: 'foo' }` and later updated to `{ innerHTML: 'bar' }` - in this case `id` needs to be unset. This can also happen in manually written render functions.

src/platforms/web/security.js
```diff
@@ -0,0 +1,31 @@
...
+import Vue from 'core/index'
+import {getTrustedTypes, isTrustedValue} from 'shared/util'
+
+type TrustedTypePolicy = {
+  // value returned is actually an object with toString method returning the wrapped value
+  createHTML: (value: any) => string;
+};
+
+let policy: ?TrustedTypePolicy
+// we need this function to clear the policy in tests
```
yyx990803: Is this strictly for testing only and there would be no legit use case for it by end users? In that case we can make it exposed in dev mode only by wrapping it with `if (process.env.NODE_ENV !== 'production')`

src/shared/util.js
```diff
@@ -79,15 +79,31 @@ export function isPromise (val: any): boolean {
   )
 }
 
+export function getTrustedTypes() {
+  // TrustedTypes have been renamed to trustedTypes https://github.com/WICG/trusted-types/issues/177
+  return typeof window !== 'undefined' && (window.trustedTypes || window.TrustedTypes);
```
yyx990803: might want to cache the result here since this will be called a lot (e.g. in `toString`)

src/platforms/web/runtime/modules/dom-props.js
```diff
@@ -51,7 +53,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
     } else if (key === 'innerHTML' && isSVG(elm.tagName) && isUndef(elm.innerHTML)) {
       // IE doesn't support innerHTML for SVG elements
       svgContainer = svgContainer || document.createElement('div')
-      svgContainer.innerHTML = `<svg>${cur}</svg>`
+      svgContainer.innerHTML = maybeCreateDangerousSvgHTML(cur)
```
yyx990803: This file actually handles the setting of normal `innerHTML` as well (down at line 72) - it's just we didn't have a dedicated branch to it.

src/shared/util.js
```diff
@@ -79,15 +79,31 @@ export function isPromise (val: any): boolean {
...
+}
+
+export function isTrustedValue(value: any): boolean {
+  const tt = getTrustedTypes();
+  if (!tt) return false;
```
yyx990803: Is `window.trustedTypes` only present when the relevant CSP is enabled?

src/platforms/web/security.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  if (!tt) return `<svg>${value}</svg>`;
+  else if (!isTrustedValue(value)) throw new Error('Expected svg innerHTML to be TrustedHTML!');
+  // flow complains 'policy' may be undefined
+  else return (policy: any).createHTML(`<svg>${value}</svg>`);
+}
+
+export function getTrustedShouldDecodeInnerHTML(href: boolean): string {
+  return href ? `<a href="\n"/>` : `<div a="\n"/>`
```
yyx990803: Shouldn't this also be a trustedType if necessary?

src/platforms/web/runtime/modules/dom-props.js
Siegrift: Thanks, however I don't think I understand.

The `props` and `oldProps` are domProps from [render function](https://vuejs.org/v2/guide/render-function.html) and `v-bind` binds values from [data property of the Vue component](https://vuejs.org/v2/guide/components.html#data-Must-Be-a-Function). I am still unable to trigger that part of the code, but I was trying to do something like [this codesandbox](https://codesandbox.io/embed/vue-template-dndzq). I also had a look at the original issue [here](https://github.com/vuejs/vue/issues/9650), but their example is different from what I want to see.

The important question is: **Can the code (where the TODO comment is) be called for dangerous properties like `innerHTML`?**

If yes, I need to add integration on this place as well, but I don't wanna include it on places which don't need it.

src/shared/util.js
Siegrift: No, it will be available natively in browsers (for now just chrome with `Experimental Web Platform features` enabled) or when there is a [polyfill](https://github.com/WICG/trusted-types).

For this function it's logical to return `false` when Trusted Types are not available. However, I think you meant the case when application is running in environment that has Trusted Types, but doesn't want to use them. In this case, the application won't have the CSP header present and in that case Trusted Types act as noop. However, we might add a parameter to Trusted Types to tell if the application wants to use them or not. This is already [tracked issue](https://github.com/WICG/trusted-types/issues/36).

Siegrift: EDIT: The code before was not compliant with what I said as I would throw and error in a specific case (innerHTML in svg in IE with trustedTypes available). I've fixed the code not to throw in this case.

src/platforms/web/security.js
Siegrift: Yeah, I was experimenting how to write this function in a secure way and I accidentally deleted the secure part :cry:

src/platforms/web/security.js
```diff
@@ -0,0 +1,57 @@
...
+
+if (process.env.NODE_ENV !== 'production') {
+  // we need this function to clear the policy in tests
+  Vue.prototype.$clearTrustedTypesPolicy = function() {
+    policy = undefined
+  }
+}
+
+// TODO: remove once https://github.com/WICG/trusted-types/issues/36 is resolved.
+function isApplicationUsingTrustedTypes() {
```
koto: Please avoid this `try {} catch` block, it's for example possible that a default policy will prevent the exception from being thrown, and TT are still enforced. Why do we need to know if application is using TT here? It should be ok to remove strinigificaton for trusted values regardless of whether the application has a CSP that enforces TT.

Siegrift: It's because of the way I implemented `maybeCreateDangerousSvgHTML`. But it's true I forgot about the default policy preventing my hack. I can check if there is a default policy and if `defPolicy.createHTML` doesn't throw.

Siegrift: We actually created a implementation that works without throwing and without ugly hacks :)

aKzenT:
Are there any plans to integrate this into vue?

kazuma0129:
@Siegrift Is it still in progress?
I arrived at this PR from the following description.
https://github.com/w3c/trusted-types/wiki/Integrations#:~:text=jQuery%3A%20bug-,VueJS%3A%20Pull%20request,-Next.js%3A

Siegrift:
Hi, thanks for the interesest. 

This PR got quite stale since I first opened it. The integration implementation would likely need to be revisited now. I am not sure whether Vue is interested in having this change merged.

Siegrift:
[I think it's best](https://github.com/vuejs/vue/pull/10491#issuecomment-1812063235) to close this PR and re-implement it.

Siegrift:
No idea, sorry.

Issue #1842, opened 2015-11-21
Suggestion: "$last", "$even", "$odd" etc (v-for special properties)


furey:
Would it be relatively inexpensive to add other special properties to `v-for` loops?

For example ([from the Angular docs](https://docs.angularjs.org/api/ng/directive/ngRepeat)):

![ngrepeat](https://cloud.githubusercontent.com/assets/1914481/11316439/4d9defde-905b-11e5-90aa-f17d6faa1912.png)

I'm currently writing a breadcrumb component and a property like `$last` would simplify:

```javascript
<li :class="{ active: $index + 1 == crumbs.length }">
```

into:

```javascript
<li :class="{ active: $last }">
```

Interested to hear your thoughts!

yyx990803:
Unfortunately, it is in fact relatively expensive to add these aliases, because in Vue these properties need to be reactive, so each of them would have extra initialization and memory costs. It would be wasteful to always define them on every repeated fragment.

You can define custom methods like `isLast` and just use it like

``` html
<li :class="{ active: isLast($index) }">
```

furey:
Understood. Cheers.

PR #10494, opened 2019-09-08
feat(v-for): add 'loop' iterator to v-for 

oguzhaninan:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

When using v-for directive, no additional statement is needed to determine whether the current element is the first or the last element, or whether the sequence is even or odd. Also added 'remaining' variable, which tells you how many elements are left before the end of the loop.

### DIFF
--- a/flow/compiler.js
+++ b/flow/compiler.js
@@ -148,6 +148,7 @@ declare type ASTElement = {
   alias?: string;
   iterator1?: string;
   iterator2?: string;
+  iterator3?: string;
 
   staticClass?: string;
   classBinding?: string;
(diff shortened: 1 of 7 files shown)


posva:
Hey, thanks a lot for the initiative but this is something we want to avoid for performance reasons (https://github.com/vuejs/vue/issues/1842)
It can however be added with custom functions.

I'm sorry you spent time building this but it's in general a bad idea to submit a non-small feature that hasn't been accepted. Something like this should have even gone through an RFC but it does have a performance impact we want to avoid

PR #10575, opened 2019-09-27
fix(ts): force mandatory options

### LINKED ISSUE #10573: typesafe Vue.use
**Reproduction link**
Please see: https://tinyurl.com/y5mlgqqh

**Steps to reproduce**

1. write a plugin
2. overload Vue.use ( see the given link above for detail )

**What is expected?**
During compiling it fails, because the types are not matching.

**What is actually happening?**
It compiles without errors

A possible solution is to change the type in vue/types/vue VueConstructor to: 

` use(plugin: PluginObject<unknown> | PluginFunction<unknown>, ...options: unknown[]): VueConstructor<V>;`

See: https://tinyurl.com/y6anfs8b

bitPogo:
Closes #10573

- [ ] Bugfix
- [ x ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ x ] No

- [ x ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**
 Example: https://tinyurl.com/y2uzn6k2

### DIFF
--- a/types/test/plugin-test.ts
+++ b/types/test/plugin-test.ts
@@ -14,10 +14,13 @@ const plugin: PluginObject<Option> = {
   }
 }
 const installer: PluginFunction<Option> = function(Vue, option) { }
+function NoOptions( _vue: typeof Vue ){};
+function OptionalOption( _vue: typeof Vue, options?: Option) {};
 
-Vue.use(plugin);
 Vue.use(plugin, new Option);
-Vue.use(plugin, new Option, new Option);
-Vue.use(installer);
 Vue.use(installer, new Option);
 Vue.use(installer, new Option, new Option, new Option);
+
+Vue.use(NoOptions);
+Vue.use(OptionalOption);
+Vue.use(OptionalOption, { prefix: '', suffix: '' });
--- a/types/test/plugin-test.ts
+++ b/types/test/plugin-test.ts
@@ -1,4 +1,4 @@
-import Vue from "../index";
+import Vue, { VueConstructor } from "../index";
 import { PluginFunction, PluginObject } from "../index";
 
 class Option {
@@ -14,8 +14,8 @@ const plugin: PluginObject<Option> = {
   }
 }
 const installer: PluginFunction<Option> = function(Vue, option) { }
-function NoOptions( _vue: typeof Vue ){};
-function OptionalOption( _vue: typeof Vue, options?: Option) {};
+function NoOptions( _Vue: VueConstructor<Vue> ){};
+function OptionalOption( _Vue: VueConstructor<Vue>, options?: Option) {};
 
 Vue.use(plugin, new Option);
 Vue.use(installer, new Option);
--- a/types/test/plugin-test.ts
+++ b/types/test/plugin-test.ts
@@ -15,6 +15,9 @@ const plugin: PluginObject<Option> = {
 }
 const installer: PluginFunction<Option> = function(Vue, option) { }
 
+Vue.use(plugin);
 Vue.use(plugin, new Option);
+Vue.use(plugin, new Option, new Option);
+Vue.use(installer);
 Vue.use(installer, new Option);
 Vue.use(installer, new Option, new Option, new Option);
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -111,8 +111,8 @@ export interface VueConstructor<V extends Vue = Vue> {
   component<Props>(id: string, definition: FunctionalComponentOptions<Props, RecordPropsDefinition<Props>>): ExtendedVue<V, {}, {}, {}, Props>;
   component(id: string, definition?: ComponentOptions<V>): ExtendedVue<V, {}, {}, {}, {}>;
 
-  use<T>(plugin: PluginObject<T> | PluginFunction<T>, options?: T): VueConstructor<V>;
-  use(plugin: PluginObject<any> | PluginFunction<any>, ...options: any[]): VueConstructor<V>;
+  use<T>(plugin: PluginObject<T> | PluginFunction<T>, ...options: T[]): VueConstructor<V>;
+  use(plugin: PluginObject<unknown> | PluginFunction<unknown>, ...options: unknown[]): VueConstructor<V>;
   mixin(mixin: VueConstructor | ComponentOptions<Vue>): VueConstructor<V>;
   compile(template: string): {
     render(createElement: typeof Vue.prototype.$createElement): VNode;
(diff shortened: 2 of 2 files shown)


types/vue.d.ts
```diff
@@ -111,8 +111,8 @@ export interface VueConstructor<V extends Vue = Vue> {
   component<Props>(id: string, definition: FunctionalComponentOptions<Props, RecordPropsDefinition<Props>>): ExtendedVue<V, {}, {}, {}, Props>;
   component(id: string, definition?: ComponentOptions<V>): ExtendedVue<V, {}, {}, {}, {}>;
 
-  use<T>(plugin: PluginObject<T> | PluginFunction<T>, options?: T): VueConstructor<V>;
-  use(plugin: PluginObject<any> | PluginFunction<any>, ...options: any[]): VueConstructor<V>;
+  use<T>(plugin: PluginObject<T> | PluginFunction<T>, ...options: T[]): VueConstructor<V>;
+  use(plugin: PluginObject<unknown> | PluginFunction<unknown>, ...options: unknown[]): VueConstructor<V>;
```
posva: Could you add a comment on top of this stating that this makes options mandatory when they are defined

types/test/plugin-test.ts
```diff
@@ -15,6 +15,9 @@ const plugin: PluginObject<Option> = {
 }
 const installer: PluginFunction<Option> = function(Vue, option) { }
 
+Vue.use(plugin);
```
posva: this one should yield an error though 🤔 

The tests that must be added are:

```ts
function NoOptions(_Vue: Vue): void {}
function OptionalOption(_Vue: Vue, options?: Option): void {}

Vue.use(NoOptions)
Vue.use(OptionalOption)
Vue.use(OptionalOption, { prefix: '', suffix: '' })
```

I'm realizing that the types you put in vue.d.ts are different from what you tried at https://www.typescriptlang.org/play/index.html#code/KYDwDg9gTgLgBDAnmYcAKAbArgcwJYB2AYlgQMYx4QEA8AKgHxwC8cAFAG5bABccAatwA0cCGErUAzgH4+dAJQsmHCHgAmAbgBQWwjGBQAZgEMyqALLBJk4zmAAlYJDgBvLXA9w7MNnAC2VjZ2ANLAiHySMFCEOHDyEVEx2gC+OnoGJmZwlta2Vpi4hADy4lQEkq5aAJABuXaSfDlBDk4QKTpIKNmBeZIF+MSkFGUs7HBcvALC-j12jbP52AMlEuVx2oZDq911i4UEbO6eE3yCwELVYqsNO819S8WlUiJa8eOqapWpWpvk2wByEBWZUkbAA+mdTtw3ip1K5Ur9htQ4MDqMYMKiDhDuFDzqInuVZLdev1HtcYR94WkCPojKZUGdKp5PFhJMB6Aw2GAHgQ+KTBn8yhyRFcQUSFHxYZojszWcBfNz9nyeSRBdQaKQANYECAAdwIDBEADoTaKpHwtTr9QBtAC6cUlH20zI8coVPPmu3u+1VSIIIoJnru-MxFQp6naOjIUngE1GLmScGMFTO2mj5XgtWajmcrATSYqTTyObaOiL9X5vgmIhcMy9fCzxdacET8g0cAA9B2vBAIGotBMjW7iRWeTW63cGwsSy31p3uzhe-3B27AaH5APuEO2WxMejMRuVzu9xiCePG-Up7sZ8kN1ouwgABaoQwQDAYPUxOCSR8QLAYT4ACNUFAbk+1QQgn1QaM-DAPAMAMTdgG3eURz2AZ1iQlDfHLdDCBrVttCPVDcO9AZx0MQwr2zZtCKw4dSP5c8FhuAByV8IFY2c2yAA

types/test/plugin-test.ts
```diff
@@ -14,7 +14,13 @@ const plugin: PluginObject<Option> = {
   }
 }
 const installer: PluginFunction<Option> = function(Vue, option) { }
+function NoOptions( _vue: typeof Vue ){};
```
posva: The typeof shouldn't be necessary, also it's better to use an uppercase version like `_Vue` or `Vue`

types/test/plugin-test.ts
bitPogo: It passes also in current dev

Thanks for that advises.

Indeed is a minor difference to the mentioned playground. The reason is 
`Vue.use(installer, new Option, new Option, new Option);`
Would fail, since the definition hits in the master the `any` case of use, which would not work with `unknown`.  (I am even not sure if this was intended in the master)

types/test/plugin-test.ts
bitPogo: Since is defined as `PluginFunction<T> = (Vue: typeof _Vue, options?: T) => void;` the compiler will complain on a direct usage of Vue as a type.

types/test/plugin-test.ts
bitPogo: Please let me know, if the `any` fallback was intended on `Vue.use<PluginOption>`. This could behavior could be restored by a additional definition  ` use<T>(plugin: PluginObject<T> | PluginFunction<T>, option1: any, option2: any, ...options: any[] ): VueConstructor<V>;`. It looks perhaps a bit odd, but `use` and a given type should fallback at n+1( n>1 ) arguments to`any` according to master.
( Please see: https://tinyurl.com/y3lcapvv )

types/vue.d.ts
bitPogo: That is done I guess

types/vue.d.ts
```diff
@@ -111,8 +111,9 @@ export interface VueConstructor<V extends Vue = Vue> {
   component<Props>(id: string, definition: FunctionalComponentOptions<Props, RecordPropsDefinition<Props>>): ExtendedVue<V, {}, {}, {}, Props>;
   component(id: string, definition?: ComponentOptions<V>): ExtendedVue<V, {}, {}, {}, {}>;
 
-  use<T>(plugin: PluginObject<T> | PluginFunction<T>, options?: T): VueConstructor<V>;
-  use(plugin: PluginObject<any> | PluginFunction<any>, ...options: any[]): VueConstructor<V>;
```
pkf1994, line 115: Test

posva:
@pkf1994 This is the 3rd time you spam with tests. Don't do it again or we will have to block you.

bitPogo:
Since this is now ancient...I am closing it - happy new year!

PR #10595, opened 2019-10-03
(v-model): trigger change event and reset value of select element if currently selected option was removed (fix #10527)

### LINKED ISSUE #10527: Value of select element is not updated when options are changed
### Version
2.6.10

### Reproduction link
[https://jsfiddle.net/rftv5zs4/](https://jsfiddle.net/rftv5zs4/)

### Steps to reproduce
Render a select element bound with v-model to some value. Options should be dynamically generated by iterating with v-for over a computed.
Select an item in the select.
Somehow trigger the computed to return new list of options.

### What is expected?
Value should be updated to undefined, watcher should be triggered if new options do not contain previously selected value.

Previous selected value is kept in model even if updated list of options does not contain selected value.

---
This works fine in 2.0.2 and was broken in 2.0.3 and later up to the latest 2.6.10 version.

RoboBurned:
When a select element is bound with v-model and currently selected option is removed then browser sets select's value to `undefined` but vue's model is not updated and still has previous value making model out of sync with DOM.
This will resolve this situation by triggering "change" event the same way how it is now handled for selects with `multiple` attribute.

close #10527 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

I think it cannot be considered as a breaking change. This PR fixes an issue introduced in a minor version (2.0.3)

**Other information:**

### DIFF
--- a/src/platforms/web/runtime/directives/model.js
+++ b/src/platforms/web/runtime/directives/model.js
@@ -63,7 +63,7 @@ const directive = {
         // no matching option found for at least one value
         const needReset = el.multiple
           ? binding.value.some(v => hasNoMatchingOption(v, curOptions))
-          : binding.value !== binding.oldValue && hasNoMatchingOption(binding.value, curOptions)
+          : binding.value !== el.value && hasNoMatchingOption(binding.value, curOptions)
         if (needReset) {
           trigger(el, 'change')
         }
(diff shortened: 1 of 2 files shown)


afwn90cj93201nixr2e1re:
It's not right.

posva:
Thanks for the PR but as said in #10527, this is intended

PR #10615, opened 2019-10-10
fix #10614: the select's v-model set to `undefined` in some corner case

### LINKED ISSUE #10614: when <select> model and the option list changed at the same time, model may incorrectly set to `undefined`
### Version
2.6.10

### Reproduction link
[https://jsfiddle.net/4fyrj95L/](https://jsfiddle.net/4fyrj95L/)

### Steps to reproduce
set model binded to select element and the array that iterated the option list at the same time

make sure the new model value do not match any of the new options

the model will be set to `undefined`

### What is expected?
The model should reverses the new given value

model's value set to `undefined`

---
This bug only appears when model and the options changed at the same and the new model value does not match any option, other situations behaves correctly as far as I tested.

xieranmaya:
fix(select-model): fix the issue of select's v-model set to `undefined` in some corner case
fix #10614

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/platforms/web/compiler/directives/model.js
+++ b/src/platforms/web/compiler/directives/model.js
@@ -118,7 +118,7 @@ function genSelect (
     `.map(function(o){var val = "_value" in o ? o._value : o.value;` +
     `return ${number ? '_n(val)' : 'val'}})`
 
-  const assignment = '$event.target.multiple ? $$selectedVal : $$selectedVal[0]'
+  const assignment = `$event.target.multiple ? $$selectedVal : ($$selectedVal.length > 0 ? $$selectedVal[0] : ${value})`
   let code = `var $$selectedVal = ${selectedVal};`
   code = `${code} ${genAssignmentCode(value, assignment)}`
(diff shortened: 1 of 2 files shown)


src/platforms/web/compiler/directives/model.js
```diff
@@ -118,7 +118,7 @@ function genSelect (
     `.map(function(o){var val = "_value" in o ? o._value : o.value;` +
     `return ${number ? '_n(val)' : 'val'}})`
 
-  const assignment = '$event.target.multiple ? $$selectedVal : $$selectedVal[0]'
```
xieranmaya: $$selectedVal may be an empty array, this way we should not return its first element
reverse the model's previous value is a reasonable chose

despreston:
@xieranmaya I'd like to get a fix for this bug into the next release. Do you mind if I open a PR off this branch that will fix this failing test?

PR #10627, opened 2019-10-13
Apply :duration property for list transition, fixes #10626

### LINKED ISSUE #10626: transition-group with duration property doesn't work
### Version
2.6.10

### Reproduction link
[https://codesandbox.io/s/vue-template-lcrsy](https://codesandbox.io/s/vue-template-lcrsy)

### Steps to reproduce
Click "Move" button.

### What is expected?
Both two lists move with animation.

Only second list which is applied duration with css has animation.

Monchi:

- [x] Bugfix
- [] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
`duration` property for `transition-group` component accepts a value with type `object`, `string`, or `number`.
If it's `object`, you can apply like `{ enter: 500, leave: 800, move: 300 }` in ms. In case of string or number, you can just specify duration like `500` as same as `transition` component.

### DIFF
--- a/src/platforms/web/runtime/components/transition-group.js
+++ b/src/platforms/web/runtime/components/transition-group.js
@@ -15,6 +15,7 @@ import { warn, extend } from 'core/util/index'
 import { addClass, removeClass } from '../class-util'
 import { transitionProps, extractTransitionData } from './transition'
 import { setActiveInstance } from 'core/instance/lifecycle'
+import { isObject } from 'shared/util'
 
 import {
   hasTransition,
@@ -117,7 +118,9 @@ export default {
         const el: any = c.elm
         const s: any = el.style
         addTransitionClass(el, moveClass)
-        s.transform = s.WebkitTransform = s.transitionDuration = ''
+        s.transform = s.WebkitTransform = ''
+        const duration = isObject(c.data.transition.duration) ? c.data.transition.duration.move : c.data.transition.duration
+        s.transitionDuration = duration ? `${duration}ms` : ''
         el.addEventListener(transitionEndEvent, el._moveCb = function cb (e) {
           if (e && e.target !== el) {
             return


Monchi:
@posva Done!

src/platforms/web/runtime/modules/transition.js
```diff
@@ -280,7 +280,7 @@ export function leave (vnode: VNodeWithData, rm: Function) {
 }
 
 // only used in dev mode
-function checkDuration (val, name, vnode) {
+export function checkDuration (val: number, name: string, vnode: VNode) {
   if (typeof val !== 'number') {
     warn(
       `<transition> explicit ${name} duration is not a valid number - ` +
```
Justineo: This warning should be updated because now it should work on both `<transition>`s and `<transition-group>`s.

src/platforms/web/runtime/components/transition-group.js
```diff
@@ -117,7 +119,16 @@ export default {
...
+        s.transform = s.WebkitTransform = ''
+        const duration: any = toNumber(
+          isObject(c.data.transition.duration)
+          ? c.data.transition.duration.move
+          : c.data.transition.duration
+        )
+        if (process.env.NODE_ENV !== 'production' && duration != null) {
+          checkDuration(duration, 'move', c)
+        }
+        s.transitionDuration = duration ? `${duration}ms` : ''
```
Justineo: No need to override the value if `duration` isn't set. (`0` should be considered as set.)

Monchi: @Justineo 
Thank you for reviewing, but original code is overriding the value at line 120. IMO, we shouldn't change it unless there is a specific reason.

Justineo: Let's keep the current behavior then :)

PR #10741, opened 2019-10-21
fix #10734[, ##10734]

### LINKED ISSUE #10734: Symbol properties non-reactive
### Version
2.6.10

### Reproduction link
[https://jsfiddle.net/dansebcar/b75n0uw6/16/](https://jsfiddle.net/dansebcar/b75n0uw6/16/)

### Steps to reproduce
- Press update to change each item's name to "b"

### What is expected?
The list re-renders with the new name

The list does not change

---
If we instead define "symbol" as any non-empty string, the expected re-render is observed, so I think the cause is just the parent list (items) is accessed through a symbol.

Nesting the symbol property an additional level (IE in config, rather than directly on data) is necessary to render the list at all; I guess symbols from data aren't directly transferred?

I noticed in the devtools when I pressed "refresh" after mutating the items that I could see the new name, so I assumed something about the way I was setting the property didn't trigger the update, but even Vue.set did not help.

xx1124961758:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -62,7 +62,7 @@ export class Observer {
    * value type is Object.
    */
   walk (obj: Object) {
-    const keys = Object.keys(obj)
+    const keys = Object.keys(obj).concat(Object.getOwnPropertySymbols(obj))
     for (let i = 0; i < keys.length; i++) {
       defineReactive(obj, keys[i])
     }


Tipwheal:
Take care about methods like `initData` that use `Object.keys` to make attributes directly declared in data() reactive, there're many `Object.keys` calls in the code, ensure that this feature will work properly in mixins,  extends or any other places where merging data and so on. The `provide/inject` option supports symbols by using `Reflect.ownKeys` and there's a function named `hasSymbol` in util to check whether symbol  and Reflect are available in runtime environment. Also, be aware that the type of param `key` in defineReactive is `string`, not `string | symbol`.

xx1124961758:
> lable in runtime environment. Also, be aware that the type of param `key` in defineReactive is `string`, not `string | symbol`.

Symbols are not currently supported by Flow。Reflect.ownkeys returns all the properties on the object, including those that cannot be enumerated. On the other hand, if there is an enumerable "__ob__" attribute on the responsive object, the stack will overflow.

guda-art:
你的意思是, vue2在对data中的对象进行响应式处理的时候, 并不会处理key为symbol的属性. 但是用$set可以解决这个问题

> > lable in runtime environment. Also, be aware that the type of param `key` in defineReactive is `string`, not `string | symbol`.
> 
> Symbols are not currently supported by Flow。Reflect.ownkeys returns all the properties on the object, including those that cannot be enumerated. On the other hand, if there is an enumerable "__ob__" attribute on the responsive object, the stack will overflow.
> 
> yes, ensure that this feature will work properly everywhere, those places also need to be dealt with.

guda-art:
> > lable in runtime environment. Also, be aware that the type of param `key` in defineReactive is `string`, not `string | symbol`.
> 
> Symbols are not currently supported by Flow。Reflect.ownkeys returns all the properties on the object, including those that cannot be enumerated. On the other hand, if there is an enumerable "__ob__" attribute on the responsive object, the stack will overflow.
> 
> yes, ensure that this feature will work properly everywhere, those places also need to be dealt with.
还是vue3牛逼啊, 虽然我用$set可以让一个对象中symbol类型的key响应, 但是如果data里面定义的是一个数组, 数组的元素是对象, 对象有symbol类型的key呢, 又该怎么办

PR #10898, opened 2019-12-05
chore : add COMMIT_EDITMSG

getspooky:


posva:
I don't think we want to change this.

Note: when opening a PR don't remove the template, instead fill it, thank you

PR #10958, opened 2019-12-27
fix(compiler): event handler with modifiers swallowing arguments

### LINKED ISSUE #10867: Multiple arguments in custom event with a modifier are not passed to handler
### Version
2.6.10

### Reproduction link
[https://jsfiddle.net/50wL7mdz/30115/](https://jsfiddle.net/50wL7mdz/30115/)

### Steps to reproduce
Add a modifier to one of the event listeners. e.g. @emission.once

### What is expected?
That all arguments will be passed to the function

Only the first argument is passed

---
My guess is that it lies in the template compiler which creates a function($event){...}  and calling the handler with $event instead of ...arguments?

despreston:
Pass full parent argument list to the event handler when there are event
modifiers present.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Fix #10867

### DIFF
--- a/src/compiler/codegen/events.js
+++ b/src/compiler/codegen/events.js
@@ -148,9 +148,9 @@ function genHandler (handler: ASTElementHandler | Array<ASTElementHandler>): str
       code += genModifierCode
     }
     const handlerCode = isMethodPath
-      ? `return ${handler.value}($event)`
+      ? `return ${handler.value}.apply(null, arguments)`
       : isFunctionExpression
-        ? `return (${handler.value})($event)`
+        ? `return (${handler.value}).apply(null, arguments)`
         : isFunctionInvocation
           ? `return ${handler.value}`
           : handler.value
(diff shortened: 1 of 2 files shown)


despreston:
Hm. This is kind of strange b/c I can successfully run the todomvc app using my build w/o issue. I'm not very familiar with nightwatch; perhaps someone with a better eye can help me debug this one.

posva:
It's probably necessary to avoid the `spread` operator to support more browsers and use `apply` instead

despreston:
@posva Thanks again!

despreston:
@posva Do you know if there's a plan to push this into production soon? I'm actually being bit by this in a project now. The workaround works but obviously I'd prefer the solution I helped fix 😝.

PR #10990, opened 2020-01-07
perf: preinitialize typeCheck RegExp

SebastianSpeitel:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Performance improvement

- [ ] Yes
- [x] No

**Other information:**

After looking at the performance of one of my apps, I noticed, the function, that took the most time summed up was `getType()` , defined in [1] taking 200ms. I noticed the regular expression getting created for each call and measured again including the proposed improvement, which took the time down to 150ms. To make sure I changed it back to confirm and it again timed at 200ms.

[1]
https://github.com/vuejs/vue/blob/b07087deae060244fe5d9651b38dc5f1e79667a9/src/core/util/props.js#L182-L185

### DIFF
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -174,13 +174,15 @@ function assertType (value: any, type: Function): {
   }
 }
 
+const functionTypeCheckRE = /^\s*function (\w+)/
+
 /**
  * Use function string name to check built-in types,
  * because a simple equality check will fail when running
  * across different vms / iframes.
  */
 function getType (fn) {
-  const match = fn && fn.toString().match(/^\s*function (\w+)/)
+  const match = fn && fn.toString().match(functionTypeCheckRE)
   return match ? match[1] : ''
 }
 


SebastianSpeitel:
Maybe further improvements could be made by removing the regular expression completely and using standard string manipulation methods.

SebastianSpeitel:
Is there anything else, that needs to be done before merging? @posva

posva:
@SebastianSpeitel You don't need to do anything else, thank you :)

SebastianSpeitel:
I made this PR 2 months ago. Is there a reasons for this taking so long?

lefreet:
@SebastianSpeitel   cached it may be better ?

```js
  const getType = cached(function getType (fn) {
    var match = fn && fn.toString().match(/^\s*function (\w+)/);
    return match ? match[1] : ''
  })
```

SebastianSpeitel:
> @SebastianSpeitel   cached it may be better ?
> 
> ```js
>   const getType = cached(function getType (fn) {
>     var match = fn && fn.toString().match(/^\s*function (\w+)/);
>     return match ? match[1] : ''
>   })
> ```

Both could be optimal.
Compile the regex only once *and* compute the type of each object only once.

lefreet:
but the `fn.toString` is a expense also, for every function  @SebastianSpeitel   sorry my later comment

Issue #7787, opened 2018-03-10
SSR - classes are not re-applied on hydration without a ref


KaelWD:
### Version
2.5.14

### Reproduction link
No ref: https://codepen.io/anon/pen/yKyBYZ
With ref: https://codepen.io/anon/pen/PRwYPO

### Steps to reproduce
Open the codepen

### What is expected?
`.has-padding` should be applied after 2sec

### What is actually happening?
Nothing changes on hydration unless the element has a ref

---
I opened a similar issue a while ago but for styles - #7063 

Same thing happens with custom attributes btw, I'd expect this to give a "mismatched nodes" error at the very least:

No ref: https://codepen.io/anon/pen/qoExmb?editors=1010
With ref: https://codepen.io/anon/pen/oqgEWL?editors=1010

Text is fine though: https://codepen.io/anon/pen/OvPQQv?editors=1010

yyx990803:
You should do it in `mounted`. Mutating state in `created` will be affecting the output in SSR so it would have no effect.

KaelWD:
Why does it work with a ref then? This is basically the same as #7063 but with a different attribute.

johnleider:
Is there a particular reason that #7063 qualifies as a bug but this does not?

iranagame:
This issue still exist and has no workaround expect adding a ref to every single element which is not very elegant way!
Any ideas!?

johnleider:
We kindly ask users to not comment on closed/resolved issues. If you believe that this issue has not been correctly resolved, please create a new [issue](https://issues.vuetifyjs.com) showing the regression or create a new [discussion](https://discussions.vuetifyjs.com).

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

Issue #7063, opened 2017-11-15
SSR - mismatched styles are not updated in the DOM unless the element has a ref


KaelWD:
### Version
2.5.3

### Reproduction link
[https://codepen.io/anon/pen/GOvGBJ?editors=1010](https://codepen.io/anon/pen/GOvGBJ?editors=1010)

### Steps to reproduce
Load the codepen. 
This is obviously not an actual SSR app, but is emulating one. The same thing happens with actual SSR that sets the padding based on window width. 

### What is expected?
The inline style should be updated when the page hydrates (see the codepen below)

### What is actually happening?
The style only updates if the element has a `ref`
See https://codepen.io/anon/pen/aVyjoM?editors=1010

KaelWD:
This doesn't seem to have been fixed for classes? 

No ref: https://codepen.io/anon/pen/yKyBYZ
With ref: https://codepen.io/anon/pen/PRwYPO

Should I open a new issue or is this expected behaviour?

PR #10993, opened 2020-01-09
fix: clear extra styles when hydrating/mounting element

dsanders11:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Inline styles already existing on an element which Vue is mounted on, or server-side rendered DOM, may be removed if the component does not render those styles. Developers should not rely on extra inline styles being left untouched.

**Other information:**

There's currently a somewhat sharp gotcha with SSR where an inline style rendered for an element on the server, but not rendered on the client, will end up stuck on the element, leading to partially broken markup.

The styles may differ between the SSR DOM and the client DOM due to practices such as using the screen size as a prop value, which causes a style to be applied on the server which is not applied on the client when loaded on mobile clients. See vuetifyjs/vuetify#10163 for a real-world example of this.

I tried to keep this PR concise and focused, but it's worth discussing potential side effects of this change. Ideally it would be scoped just to hydration to limit the surface area for side effects, but that wasn't as simple as limiting it to creation, so I opted for the simpler route for the PR.

Some users may be relying on the current behavior (perhaps without realizing it), so this would break behavior for them. However, considering the current gotcha for otherwise sane looking code, it seems to be a reasonable trade-off to potentially break a tiny amount of code to remove a hard to debug issue which can lead to messed up markup.

An argument could be made that much as the DOM structure needing to match when hydrating, perhaps styles should also be required to match, which would avoid this issue. However, it seems like a tall order to ensure all styles are rendered identically between server and client since more style-oriented information is available on the client (screen size, device capabilities, etc). So, that seems like an overly burdensome requirement to put on developers for SSR. However, if that's preferred, this PR could be turned into issuing a warning on style-mismatch instead.

### DIFF
--- a/src/platforms/web/runtime/modules/style.js
+++ b/src/platforms/web/runtime/modules/style.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { getStyle, normalizeStyleBinding } from 'web/util/style'
+import { getStyle, normalizeStyleBinding, parseStyleText } from 'web/util/style'
 import { cached, camelize, extend, isDef, isUndef, hyphenate } from 'shared/util'
 
 const cssVarRE = /^--/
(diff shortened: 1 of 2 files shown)


KaelWD:
Related: #7063, #7787

Issue #11104, opened 2020-02-13
Global event bus sets listeners to null on bad code


hawkeye64:
### Version
2.6.11

### Reproduction link
[https://codepen.io/Hawkeye64/pen/bGdENZo?editors=1010](https://codepen.io/Hawkeye64/pen/bGdENZo?editors=1010)

### Steps to reproduce
The link has the steps. Basically, you need two components that are subsequently created and destroyed. The first component registers an event with the global event bus and de-registers it when destroyed. The second component, only has the de-registration in it's BeforeDestroy. When going from 1st component, to second component, then back to first component, the first component is created, the event is registered, the second component is destroyed and array of listeners for that event is set to null when the de-registration is done.
I know this is bad coding. But could happen to anyone. This took us two days to figure out. I wanted to learn more about the issue, so wrote the codepen to illustrate the issue and to test various versions of Vue.
This issue does not occur in v2.5.2 or earlier. It started with v2.5.3 and is present in latest (v2.6.11).

### What is expected?
Vue should keep track of $on/$off listeners and ignore $off if no corresponding $on - at least don't set the listeners array to null.
With the link to reproduce the bug, I have done the latest (2.6.11) and back to 2.5.3 - this is NOT a bug in 2.5.2 as the reproduction works in that version, so quite possibly something changed.

### What is actually happening?
The listeners array for the event is set to null. Therefore, no other listeners can receive the event messages from the global event bus.

---
Before a jr dev left our company, he was doing some clean up. In the mounted() he remove some code `this.$root.$on('alarms:new', this.myFunc)` but forgot to remove in the `BeforeDestroy` this `this.$root.$off('alarms:new', this.myFunc)`. Subsequently, we stopped getting this particular event with other components. 
What we found is that the innocuous code, when calling $root.$off does not have a matching $root.$on, so the event array for the listeners is set to null.

jesusvilla:
This occurs because the event is deleted when there is no function declared in [vm.$off](https://github.com/vuejs/vue/blob/dev/src/core/instance/events.js#L101-L103) but it should not be taken into account when it is $root.
Look: https://codepen.io/jvillanueva/pen/ExjVdGg

hawkeye64:
@jesusvilla Good fix! Hopefully the Vue folks incorporate that.

hawkeye64:
Here is what happened in dev tools: 
[...]

PR #11105, opened 2020-02-13
fix(core): remove event only when it is not $root

jesusvilla:

- [ x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Look #11104

### DIFF
--- a/src/core/instance/events.js
+++ b/src/core/instance/events.js
@@ -98,7 +98,7 @@ export function eventsMixin (Vue: Class<Component>) {
     if (!cbs) {
       return vm
     }
-    if (!fn) {
+    if (!fn && vm !== vm.$root) {
       vm._events[event] = null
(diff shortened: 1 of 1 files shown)


posva:
See https://github.com/vuejs/vue/issues/11104#issuecomment-585821293

I'm concerned this code doesn't break any test. The Vue instance being the `$root` doesn't mean it shouldn't remove all event listeners when no function is provided. Maybe you want to add a test case for that? 🙂 I haven't checked but I guess there are tests for instances but none using a plain Vue instance as a bus

jesusvilla:
@posva but it passed the tests, and these are oriented to plain Vue instance

PR #11111, opened 2020-02-15
fix(ssr): ignore empty text vnode when hydrating (fix #11109)

### LINKED ISSUE #11109: Hydration fails when write like "<div><span>123</span>{{value}}</div>" in the template and the data "value" is a empty string
### Version
2.6.11

### Reproduction link
[https://codepen.io/deqwin/pen/wvaMxoO](https://codepen.io/deqwin/pen/wvaMxoO)

### Steps to reproduce
open the reproduction above and you will see a red vue warning in the console

### What is expected?
a successful hydration

the hydration fails and the app re-renders absolutely

---
this looks like the problem caused by the empty text node (src/core/vdom/patch.js line:646)

deqwin:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Closes #11109

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -645,8 +645,9 @@ export function createPatchFunction (backend) {
             let childNode = elm.firstChild
             for (let i = 0; i < children.length; i++) {
               const child = children[i]
-              // ignore empty text vnode
+              // ignore empty text vnode but create a empty textNode for next patch
               if (isUndef(child.tag) && isFalse(child.isComment) && child.text === '') {
+                createElm(child, insertedVnodeQueue, elm, childNode)
                 continue
               }
               if (!childNode || !hydrate(childNode, child, insertedVnodeQueue, inVPre)) {
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -23,6 +23,7 @@ import {
   isDef,
   isUndef,
   isTrue,
+  isFalse,
   makeMap,
   isRegExp,
   isPrimitive
@@ -643,7 +644,12 @@ export function createPatchFunction (backend) {
             let childrenMatch = true
             let childNode = elm.firstChild
             for (let i = 0; i < children.length; i++) {
-              if (!childNode || !hydrate(childNode, children[i], insertedVnodeQueue, inVPre)) {
+              const child = children[i]
+              // ignore empty text vnode
+              if (isUndef(child.tag) && isFalse(child.isComment) && child.text === '') {
+                continue
+              }
+              if (!childNode || !hydrate(childNode, child, insertedVnodeQueue, inVPre)) {
                 childrenMatch = false
                 break
               }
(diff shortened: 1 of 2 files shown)


test/unit/modules/vdom/patch/hydration.spec.js
```diff
@@ -388,4 +388,18 @@ describe('vdom patch: hydration', () => {
       expect(dom.children[0].className).toBe('bar')
     }).then(done)
   })
+
+  // #11109
+  it('should not warn failed hydration when there are some empty text vnode in children', () => {
```
posva:
```suggestion
  it('should not fail hydration with empty text vnodes children', () => {
```

posva:
for GitHub: this closes #11109

src/core/vdom/patch.js
```diff
@@ -643,7 +644,12 @@ export function createPatchFunction (backend) {
             let childrenMatch = true
             let childNode = elm.firstChild
             for (let i = 0; i < children.length; i++) {
-              if (!childNode || !hydrate(childNode, children[i], insertedVnodeQueue, inVPre)) {
+              const child = children[i]
+              // ignore empty text vnode
+              if (isUndef(child.tag) && isFalse(child.isComment) && child.text === '') {
+                continue
```
contribu: This results in child.elm === undefined and cause a crash in next patch.
Reproduction: https://github.com/deqwin/vue/compare/fix_ssr_hydration...contribu:fix_ssr_hydration_crash

deqwin: Thanks for your suggestions. i will see how to modify.

deqwin: The fact is not that the vnode tree has an extra empty text node, but the DOM tree is missing a necessary empty text node.

deqwin: @posva I have modified the code to deal with this problem.

deqwin: @posva I have modified the code to deal with this problem and also updated the test.

contribu: Thanks!
I am unsure if this causes repainting during hydration and if the existing hydration code allows it.
I think it shouldn't cause repainting if the existing hydration code disallows repainting.

deqwin: Yeah, it‘s worth exploring. However, it is necessary to keep the vnode structure and DOM structure consistent at all times. So I think it's ok to do this step here.

contribu: Another solution is rendering empty text nodes as comment nodes in SSR like async placeholder or v-if="false".
I am unsure which solution is better.

OllysCoding:
Hi All, is there any update on this PR being merged? This is causing us issues in our of SSR apps

RomainLanz:
Hey there! 👋🏻 

Small bump to see if there is any way to see this being merged and released for Vue 3?

posva:
This is only for Vue 2. Check in the core repo if the issue exists

PR #11152, opened 2020-02-27
fix(compiler): Allow BigInt usage in templates

### LINKED ISSUE #11126: Allow BigInt usage in templates
It would allow us to use numbers which is larger than MAX_SAFE_INTEGER. 

In HTML, `{{BigInt("100")}}`, `{{100n}}`, `{{2n * 50n}}` or `{{50n + 50n}}` should show "100" (it can be just simply string by using .toString()) at frontend.
At the moment using BigInt in "Mustache" syntax just throw error.
Sure, backup solution would be using external functions or converted into string and save it to variable. But using BigInt directly in HTML would be nice to have.
More Information about BigInt -> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

shadowings-zy:

- [ ] Bugfix
- [ x ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ x ] No

- [ x ] It's submitted to the `dev` branch for v2.x (or to a previous version branch), _not_ the `master` branch

- [ x ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**
Hello!
This pull request is related to issue #11126 . Now we can use BigInt in "Mustache" syntax.
I had already added test for this feature in "test/unit/features/filter/filter.spec.js".
And all tests are passing.

Close #11126


posva:
Hi, thanks for the PR but this is not what was mentioned on the issue. The idea is to allow the use of BigInt when it's available and fail if it's not. See https://jsfiddle.net/syrnaefq/. This should work on browsers supporting `BigInt`. It's similar to allowing `Number` in templates. The syntax `100n` seems to already work when it's supported by the browser.
The goal is not to polyfill the feature

shadowings-zy:
That makes me a little confused...
In my pull request," {{ BigInt(100) }}" and "{{ 100n }}" both work when BigInt is available, and fail if it is not ( and warn 'BigInt is not support! ' )

It seems that the idea of this issue is the same to my PR🤦‍♂️？

shadowings-zy:
Did you mean: 'We need to "{{ BigInt(100) }}" and "{{ 100n }}" both work when BigInt is available, and fail if it is not' like ' "{{ Number(100) }}" and "{{ 100n }}" both work'?

If so, I get your point and I will fix it in this way.

shadowings-zy:
Hi @posva !
I revert the previous commits and I rewrite the code to allow BigInt usage in templates.
As you see it just like allow `Number` in template.

截屏2020-02-2721 47 18

If it's OK, please merge this PR, if not, let's discuss further : )

src/core/instance/proxy.js
```diff
@@ -6,11 +6,16 @@ import { warn, makeMap, isNative } from '../util/index'
...
+    (typeof window !== 'undefined' && typeof window.BigInt === 'function') ||
+    (typeof global !== 'undefined' && typeof global.BigInt === 'function')
+
   const allowedGlobals = makeMap(
     'Infinity,undefined,NaN,isFinite,isNaN,' +
     'parseFloat,parseInt,decodeURI,decodeURIComponent,encodeURI,encodeURIComponent,' +
     'Math,Number,Date,Array,Object,Boolean,String,RegExp,Map,Set,JSON,Intl,' +
-    'require' // for Webpack/Browserify
+    'require,' + // for Webpack/Browserify
+    (supportBigInt ? 'BigInt' : '') // for BigInt support issue #11126
```
posva: I think it's fine to not use a conditional, the same way we allow `Intl` even though it's not supported before IE 11

shadowings-zy: OK，That sounds right，and I will add `BigInt` directly in makeMap()
: )

src/core/instance/proxy.js
```diff
@@ -10,7 +10,8 @@ if (process.env.NODE_ENV !== 'production') {
     'Infinity,undefined,NaN,isFinite,isNaN,' +
     'parseFloat,parseInt,decodeURI,decodeURIComponent,encodeURI,encodeURIComponent,' +
     'Math,Number,Date,Array,Object,Boolean,String,RegExp,Map,Set,JSON,Intl,' +
-    'require' // for Webpack/Browserify
+    'require,' + // for Webpack/Browserify
+    'BigInt' // for BigInt support issue #11126
```
posva: let's put it up after the Intl, the comment is not really necessary

shadowings-zy: OK~, I fixed it in commit#4241f12

posva (changes requested):
@shadowings-zy can you undo the merge you did? it seems to have messed up the PR

shadowings-zy:
> @shadowings-zy can you undo the merge you did? it seems to have messed up the PR

@posva Sorry, I forgot I use `dev` branch to open this PR, and now I undo my merge.

posva (approved):

PR #11191, opened 2020-03-08
fix(props): support BigInt in props type validation

### LINKED ISSUE #11126: Allow BigInt usage in templates

AjiTae:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Just a simple addition to make BigInt validation work.

Close #11126

### DIFF
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -146,7 +146,7 @@ function assertProp (
   }
 }
 
-const simpleCheckRE = /^(String|Number|Boolean|Function|Symbol)$/
+const simpleCheckRE = /^(String|Number|Boolean|Function|Symbol|BigInt)$/
 
 function assertType (value: any, type: Function): {
   valid: boolean;
--- a/test/unit/features/options/props.spec.js
+++ b/test/unit/features/options/props.spec.js
@@ -242,6 +242,16 @@ describe('Options props', () => {
         expect('Expected Symbol, got Object').toHaveBeenWarned()
       })
     }
+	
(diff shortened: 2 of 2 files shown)


anhmv:
I don't think it has to support BigInt, BigInt rarely uses and not supported by most browsers.

AjiTae:
> I don't think it has to support BigInt, BigInt rarely uses and not supported by most browsers.
Rarely, yes, but nonetheless. There is no way to do a BigInt type check now even if I want to. Like [Number, BigInt] - won't work.

anhmv:
@AjiTae you might have to create the validator function for your props, for example:

```js
yourBigInt: {
  validator: v => typeof v === 'bigint',
}
```

mireinaa24hu (commented):
okw

PR #11195, opened 2020-03-09
modify attribute regexp

iiijarvis:
delete symbol "+"


posva:
what is this solving?

PR #11487, opened 2020-07-03
fix(vdom): avoid executing root level script tags

### LINKED ISSUE #11483: Root level script tag should not be executed
### Version
2.6.11

### Reproduction link
[https://codesandbox.io/s/cranky-moser-p90di?file=/src/main.js](https://codesandbox.io/s/cranky-moser-p90di?file=/src/main.js)

### Steps to reproduce
Put valid <script> element in template string

### What is expected?
Script should be blocked from executing

Script executes

---
As the example illustrates, if a <script> element is the sole element passed into the template string of a component, it will execute said script block. However if it's pre-ceded by a valid element, eg: a div the script is blocked from execution by the renderer. You do not have to close said element as shown in examples, simply preceding the script tag with a valid opening tag suffices. I believe this is because the renderer automatically closes elements with missing closures.

shadowings-zy:

- [ x ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ x ] No

- [ x ] It's submitted to the `dev` branch for v2.x (or to a previous version branch), _not_ the `master` branch

**Other information:**
Root level <script> tags should not be executed, for consistent behavior. So I remove the code in <script> tag when the <script> tag is the root element of the template.

fix #11483

### DIFF
--- a/src/compiler/codegen/index.js
+++ b/src/compiler/codegen/index.js
@@ -45,7 +45,8 @@ export function generate (
   options: CompilerOptions
 ): CodegenResult {
   const state = new CodegenState(options)
-  const code = ast ? genElement(ast, state) : '_c("div")'
+  // fix #11483, Root level <script> tags should not be rendered.
+  const code = ast ? (ast.tag === 'script' ? 'null' : genElement(ast, state)) : '_c("div")'
   return {
     render: `with(this){return ${code}}`,
     staticRenderFns: state.staticRenderFns
--- a/test/unit/features/component/component.spec.js
+++ b/test/unit/features/component/component.spec.js
@@ -426,4 +426,17 @@ describe('Component', () => {
       vm.$destroy()
     }).then(done)
   })
+
+  it('render vnode with <script> tag as root element', () => {
+    const vm = new Vue({
+      template: '<scriptTest></scriptTest>',
+      components: {
+        scriptTest: {
+          template: '<script>console.log(1)</script>'
+        }
+      }
+    }).$mount()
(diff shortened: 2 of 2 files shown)


shadowings-zy:
Aha, I've thought about fix this bug at the compilation level！
But when I read the code I found that if a render function that returns null, the template will render a <div> as a default element (related codes in file 'src/compiler/codegen/index.js', in 'generate' function). Does this meet our expectations?

If so, I will work on it.

posva:
if the render function returns null, Vue will output a comment node. But still, it is better that way, yes

shadowings-zy:
Hi, now I fix this bug at the compilation level.
and `Vue.compile('<script` />')` will creates a render function that returns null.

The output of the render function is as follows:

```
function anonymous(
) {
with(this){return null}
}
```

: )

src/compiler/codegen/index.js
```diff
@@ -45,7 +45,7 @@ export function generate (
   options: CompilerOptions
 ): CodegenResult {
   const state = new CodegenState(options)
-  const code = ast ? genElement(ast, state) : '_c("div")'
+  const code = ast ? (ast.tag === 'script' ? 'null' : genElement(ast, state) ) : '_c("div")'
```
posva, line 48: can you add a comment above referencing the original issue?

shadowings-zy: ok, I will add the issue

shadowings-zy:
I had added a comment referencing the original issue in the PR.

: )

PR #11527, opened 2020-07-18
[Fix] common misspelling errors

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines


posva:
Thank you but all original words were correct

Issue #11569, opened 2020-08-06
Safari doesn't see frame inside vue app


dscherbakov:
### Version
2.6.11

### Reproduction link
[problem: https://jsfiddle.net/cqb5zr8s/ not problem:https://jsfiddle.net/e0c8tos9/](problem: https://jsfiddle.net/cqb5zr8s/ not problem:https://jsfiddle.net/e0c8tos9/)

### Steps to reproduce
Safari: <div id="vue_app"><iframe name="frame"></iframe> </div> <form target="frame"></form> => the form or link will open in a new tab
Safari: <iframe name="frame"></iframe> </div> <div id="vue_app"><form target="frame"></form> => the form or link will be loaded into the iframe

### What is expected?
Open a form or link in the specified frame.

### What is actually happening?
Safari doesn't see iframe within vue app. When you specify in the form(link) the target name iframe safari opens in a new tab, as if the iframe is not there.

posva:
This is weird. I would guess Vue is adding some attribute or property that is making this fail in Safari but can't tell which one

christoph-schaeffer:
Hey @posva may i give this a shot? i'd like to contribute

posva:
for sure @christoph-schaeffer !

christoph-schaeffer:
I've created a PR with a detailed explanation on what caused this.

PR #11578, opened 2020-08-08
fix: iframe-window-name-update (fixes Issue #11569)

christoph-schaeffer:
fix: When changing the attribute 'name' of an iframe element, the iframes window must be updated aswell. Otherwise anchor tags with a 'target' attribute won't hit the right iframe.

- [x ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This Issue actually wasn’t only present in safari. When iframes are created a window will be attached to the iframe which actually is the target. When updating the name attribute of an existing iframe the window inside the iframe doesn’t get updated.

In native javascript you could fix this with:
document.querySelector('iframe').documentWindow.name = 'theNewName';

The reproduction fiddle in the Issue #11569 only is bugged in safari, but if you try to change the attribute dynamically it doesn’t work in chrome or firefox either.

Check this fiddle for the cross-platform issue : https://jsfiddle.net/ug2jhmL6/1/

### DIFF
--- a/src/platforms/web/runtime/modules/attrs.js
+++ b/src/platforms/web/runtime/modules/attrs.js
@@ -109,7 +109,14 @@ function baseSetAttr (el, key, value) {
       // $flow-disable-line
       el.__ieph = true /* IE placeholder patched */
     }
-    el.setAttribute(key, value)
+    // When changing the attribute 'name' of an iframe element, the iframes window must be updated aswell. Otherwise
+    // anchor tags with a 'target' attribute won't hit the right iframe.
+    if (el.tagName === 'IFRAME' && key === 'name') {
+      setTimeout((el: HTMLIFrameElement) => {
+        el.contentWindow.name = value;
+      }, 0, el);
+    }
+    el.setAttribute(key, value);
   }
 }
 


src/platforms/web/runtime/modules/attrs.js
```diff
@@ -109,7 +109,14 @@ function baseSetAttr (el, key, value) {
       // $flow-disable-line
       el.__ieph = true /* IE placeholder patched */
     }
-    el.setAttribute(key, value)
+    // When changing the attribute 'name' of an iframe element, the iframes window must be updated aswell. Otherwise
+    // anchor tags with a 'target' attribute won't hit the right iframe.
+    if (el.tagName === 'IFRAME' && key === 'name') {
+      setTimeout((el: HTMLIFrameElement) => {
```
posva, line 115: Is there any reason we need a parameter instead of directly using `el` inside of the function?

posva (commented):
Thanks for this and sorry for the late review!

src/platforms/web/runtime/modules/attrs.js
```diff
@@ -109,7 +109,13 @@ function baseSetAttr (el, key, value) {
       // $flow-disable-line
       el.__ieph = true /* IE placeholder patched */
     }
-    el.setAttribute(key, value)
+    // When changing the attribute 'name' of an iframe element, the iframes window must be updated as well. Otherwise
+    // anchor tags with a 'target' attribute won't hit the right iframe.
+    if (el.tagName === 'IFRAME' && key === 'name' && el.contentWindow) {
+        // $flow-disable-line
+        el.contentWindow.name = value;
```
krishan, line 116: I could be wrong, but AFAICS this could trigger a browser security error, if the `iFrame` is cross-origin, e.g. like this:

```
Uncaught DOMException: Blocked a frame with origin "http://localhost:8080" from accessing a cross-origin frame.
```

PR #11761, opened 2020-11-09
fix #11757 props on Object.prototype

### LINKED ISSUE #11757: when use key in Object.prototype carelessly as prop or computed, it will work wrong way
### Version
2.6.11

### Steps to reproduce
```html
<main>{{ valueOf }}{{ toString }}</main>
<script>
new Vue({
  el: 'main',
  props: { valueOf: { default: 'abc' } },
  computed: { toString () { return 'def'; } },
});
</script>
```

### What is expected?
They are all keys match the uni simple rule: not start with `_` or `$`

they won't be set follow options, and default feature not work

---

1. the way of checking in these 2 places need to change (if not change the proto chain root of Vue.prototype away from Object.prototype):

https://github.com/vuejs/vue/blob/b800e8e9ee4fa4be9b3d6130b5ee82d668066870/src/core/instance/state.js#L105

https://github.com/vuejs/vue/blob/b800e8e9ee4fa4be9b3d6130b5ee82d668066870/src/core/instance/state.js#L198

2. default propsData here need to be replaced for props `default`:

https://github.com/vuejs/vue/pull/11756/files#diff-5d5d019329384d9be078c97de1c463a1afd5557340edc87039463e204535af83R65

3. there still leave `constructor` which can't be used, it need delete Vue.prototype.constructor, and give an alt way to do what `vm.constructor` do in current source code.

LongTengDao:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

I found that `new ( Vue.extend({}) )()` and components inside will work right, just void to `new Vue({})` (root case), things will be ok.

So at least fix the `propsData` line, that will cause bug when prop is declared but no prop pass in (vm will get function on proto, not undefined nor default), that's easy to fix, and can't avoid like above in userland.

### DIFF
--- a/src/core/instance/state.js
+++ b/src/core/instance/state.js
@@ -62,7 +62,7 @@ export function initState (vm: Component) {
 }
 
 function initProps (vm: Component, propsOptions: Object) {
-  const propsData = vm.$options.propsData || {}
+  const propsData = vm.$options.propsData || Object.create(null)
   const props = vm._props = {}
   // cache prop keys so that future props updates can iterate using Array
   // instead of dynamic object key enumeration.


sirlancelot:
I suspect you'll encounter many issues more than just this if your codebase is polluting `Object.prototype`.

posva:
I already answered this...

PR #11882, opened 2021-01-21
fix: fix #11881

### LINKED ISSUE #11881: keep-alive component's `include` props cannot trim blank space if there's any
### Version
2.6.11

### Steps to reproduce

following code makes a simple example:
```
<keep-alive include="a,    b,c">
    <router-view> </router-view>
</keep-alive>
```

### What is expected?
component a, b, c should all be cached.

`a` and `c` are fine except `b` with blank is failed to be kept alive.

---
When do the matching, should consider this situation because some devs like to add a space between words or formatted by some tools.

JerryYuanJ:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/src/core/components/keep-alive.js
+++ b/src/core/components/keep-alive.js
@@ -13,7 +13,7 @@ function matches (pattern: string | RegExp | Array<string>, name: string): boole
   if (Array.isArray(pattern)) {
     return pattern.indexOf(name) > -1
   } else if (typeof pattern === 'string') {
-    return pattern.split(',').indexOf(name) > -1
+    return pattern.split(',').map(i => i.trim()).indexOf(name) > -1
   } else if (isRegExp(pattern)) {
     return pattern.test(name)
   }
--- a/test/unit/features/component/component-keep-alive.spec.js
+++ b/test/unit/features/component/component-keep-alive.spec.js
@@ -254,6 +254,28 @@ describe('Component keep-alive', () => {
     sharedAssertions(vm, done)
   })
 
+  it('include (string with space)', done => {
+    const vm = new Vue({
+      template: `
+        <div v-if="ok">
+          <keep-alive include="one, two">
+            <component :is="view"></component>
+          </keep-alive>
+        </div>
+      `,
+      data: {
+        view: 'one',
+        ok: true
+      },
+      components
+    }).$mount()
+    vm.view = 'two'
+    waitForUpdate(() => {
+      assertHookCalls(one, [1, 1, 1, 1, 0])
+      assertHookCalls(two, [1, 1, 1, 0, 0])
+    }).then(done)
+  })
+
   it('include (regex)', done => {
     const vm = new Vue({
       template: `


JerryYuanJ:
Hey devs, I notice that it's been a week since I open and create PR for this one, is there anything I'm missing ? Maybe it's too tiny for you guys , let me know if there's any info I can provide or we don't need this so I can close this one, thanks 😉

src/core/components/keep-alive.js
```diff
@@ -13,7 +13,7 @@ function matches (pattern: string | RegExp | Array<string>, name: string): boole
   if (Array.isArray(pattern)) {
     return pattern.indexOf(name) > -1
   } else if (typeof pattern === 'string') {
-    return pattern.split(',').indexOf(name) > -1
+    return pattern.split(',').map(i => i.trim()).indexOf(name) > -1
```
Progowi, line 16: just asking - maybe it's a good idea to trim the name too?

JerryYuanJ: makes sense, sorry for missing that, thanks.

3imed-jaberi (approved):

Leo-liiiiii (approved):

w3bdesign (approved):

src/core/components/keep-alive.js
```diff
@@ -10,10 +10,11 @@ function getComponentName (opts: ?VNodeComponentOptions): ?string {
 }
 
 function matches (pattern: string | RegExp | Array<string>, name: string): boolean {
+  name = toString(name).trim()
```
nurlansu, line 13: Just wondering, do we really need to call `toString` here? Isn't `name` already a string?

nurlansu (changes requested):

src/core/components/keep-alive.js
JerryYuanJ: Thanks for pointing out. The name is fetched from this function:
```js
function getComponentName (opts: ?VNodeComponentOptions): ?string {
  return opts && (opts.Ctor.options.name || opts.tag)
}
```
Since the `options.name` is not required and `opt.tag` is optional: 
```js
export interface VNodeComponentOptions {
  Ctor: typeof Vue;
  propsData?: object;
  listeners?: object;
  children?: VNode[];
  tag?: string;
}
```
so here I use `toString` for double check.

nurlansu: Thank you for your answer. You are right, we are getting `name` via `getComponentName`, but before calling `matches`, we are checking that `name` is not `null|undefined`:

```js
// included
(include && (!name || !matches(include, name))) ||
// excluded
(exclude && name && matches(exclude, name))
```

nurlansu: What I'm trying to say is that `matches` will always be called with `name` as string. It's declared as `string` not `?string`.

- `matches` is called in `pruneCache`, there is a verification that `name` is not null.

- `matches` is called in `render`, there is a verification that `name` is not null.

JerryYuanJ: Yes, you're right, I'll update it. Thanks.

PR #11906, opened 2021-02-11
fix(types): async Component types

mathe42:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No
- [x] Not sure if this is considered a breaking change 

**Other information:**
Look at this codeexample:
```ts
import { AsyncComponent, Component } from 'vue'
const a: AsyncComponent = () => ({
  component: new Promise<Component>((res, rej) => {
    res({})
  }),
})

const b: AsyncComponent = () => ({
  component: () =>
    new Promise<Component>((res, rej) => {
      res({})
    }),
})
```

The first example is like in the docs https://vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State and the second example should not work (I also looked at the implementation and I think this would not work).

The types are currently like such that the first example throws a type Error and the second throws no error. This PR fixes that.

### DIFF
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -12,14 +12,10 @@ export type Component<Data=DefaultData<never>, Methods=DefaultMethods<never>, Co
   | FunctionalComponentOptions<Props>
   | ComponentOptions<never, Data, Methods, Computed, Props>
 
-interface EsModuleComponent<
-  Data = DefaultData<never>,
-  Methods = DefaultMethods<never>,
-  Computed = DefaultComputed,
-  Props = DefaultProps
-> {
-  default: Component<Data, Methods, Computed, Props>;
-}
+type EsModule<T> = T | { default: T }
+
+type ImportedComponent<Data=DefaultData<never>, Methods=DefaultMethods<never>, Computed=DefaultComputed, Props=DefaultProps>
+  = EsModule<Component<Data, Methods, Computed, Props>>
 
 export type AsyncComponent<Data=DefaultData<never>, Methods=DefaultMethods<never>, Computed=DefaultComputed, Props=DefaultProps>
   = AsyncComponentPromise<Data, Methods, Computed, Props>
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Can you add a test?

mathe42:
> Can you add a test?

What exactly should be tested? -  As I understand the repro there are no type tests. I just found https://github.com/vuejs/vue/blob/dev/test/unit/features/component/component-async.spec.js#L322

I can add there a test that the second example doesn't work (see https://codepen.io/mathe42/pen/poNNNYx) 
there are tests that the first example works.

posva:
In this case a type test case would suffice: https://github.com/vuejs/vue/tree/dev/types/test

mathe42:
Thanks didn't saw that. Just pushed some tests (also for the simpe usages of AsyncComponent).

Not shure about my `const b` test. This tests that the syntax that currently is allowed (by the types) failes with this PR.

types/options.d.ts
```diff
@@ -25,13 +25,18 @@ export type AsyncComponentPromise<Data=DefaultData<never>, Methods=DefaultMethod
...
+  Computed = DefaultComputed,
+  Props = DefaultProps
+> = () => {
+  component: Promise<Component | EsModuleComponent>;
   loading?: Component | EsModuleComponent;
   error?: Component | EsModuleComponent;
   delay?: number;
   timeout?: number;
-}
+};
```
posva, line 39:
```suggestion
}
```

types/options.d.ts
```diff
@@ -25,13 +25,18 @@ export type AsyncComponentPromise<Data=DefaultData<never>, Methods=DefaultMethod
...
 
-export type AsyncComponentFactory<Data=DefaultData<never>, Methods=DefaultMethods<never>, Computed=DefaultComputed, Props=DefaultProps> = () => {
-  component: AsyncComponentPromise<Data, Methods, Computed, Props>;
+export type AsyncComponentFactory<
+  Data = DefaultData<never>,
+  Methods = DefaultMethods<never>,
+  Computed = DefaultComputed,
+  Props = DefaultProps
+> = () => {
+  component: Promise<Component | EsModuleComponent>;
```
posva, line 34: Shouldn't component be typed?
```suggestion
  component: Promise<Component<Data, Methods, Computed, Props> | EsModuleComponent>;
```

We could also extend `EsModuleComponent` to accept the same generics

types/test/async-component-test.ts
```diff
@@ -0,0 +1,61 @@
+import Vue, { AsyncComponent, Component } from "../index";
+
+const a: AsyncComponent = () => ({
+  component: new Promise<Component>((res, rej) => {
+    res({ template: "" });
+  }),
+});
```
posva, line 7: can you adapt the style to follow the existing one? eslint --fix should remove the semi colons and extra commas

posva (changes requested):
Thanks!

posva (changes requested):
Can you leave all unrelated changes out of the PR please? I think we will handle the formatting in a different commit

mathe42:
Yes sorry, there was more in the commits than I wanted. Working on it.

mathe42:
@posva Removed the styling changes.

Hope this is fine now.

types/test/async-component-test.ts
```diff
@@ -0,0 +1,61 @@
...
+  })
+})
+
+Vue.component("async-compponent1", () => ({
+  component: new Promise<Component>((res, rej) => {
+    res({
+      template: ""
+    })
+  })
+}))
```
posva, line 49: Isn't this the same as:

```suggestion
Vue.component("async-compponent1", b)
```

types/test/async-component-test.ts
```diff
@@ -0,0 +1,61 @@
...
+  "async-compponent2",
+  () =>
+    new Promise<{ default: Component }>((res, rej) => {
+      res({
+        default: {
+          template: ""
+        }
+      })
+    })
+)
```
posva, line 61: same here?

```suggestion
Vue.component("async-compponent2", d)
```

types/test/async-component-test.ts
mathe42: Its not the same but youre right we can remove them...

mathe42:
I repleced the Vue.component tests with a single one to test if it accepts AsyncComponents...

posva (approved):
Thanks!

PR #11910, opened 2021-02-13
fix(model): keep -0 to '-0'

### LINKED ISSUE #11909: Number -0 cannot be correctly rendered in input :value
### Version
2.6.11

### Reproduction link
[https://codepen.io/yuchaosydney/pen/xxRgEPP](https://codepen.io/yuchaosydney/pen/xxRgEPP)

### Steps to reproduce
1. Open reproduction link
2. Compare display value from input tag rendered by Vue and plain input html tag

### What is expected?
Input dom should correctly render value -0

It renders as 0

---
In javascript, Number `-0` is legitimate value, this can be verified by checking `Number(-0) === -0`. Therefore, when input `:value` is number `-0`, input tag should correctly display it, just as a plain input html tag does.

yuwu9145:
fix #11909

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
In javascript, Number `-0` is legitimate value, this can be verified by checking `Number(-0) === -0`. However, before set a number value to input tag, Vue uses `String(value)` to convert the value to string but `String(-0)` will result in `0` rather than `-0`.

### DIFF
--- a/src/platforms/web/runtime/modules/dom-props.js
+++ b/src/platforms/web/runtime/modules/dom-props.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { isDef, isUndef, extend, toNumber, isNumNegZero } from 'shared/util'
+import { isDef, isUndef, extend, toNumber } from 'shared/util'
 import { isSVG } from 'web/util/index'
 
 let svgContainer
@@ -44,7 +44,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
       // non-string values will be stringified
       elm._value = cur
       // avoid resetting cursor position when value is the same
-      const strCur = isNumNegZero(cur) ? '-0' : isUndef(cur) ? '' : String(cur)
+      const strCur = isUndef(cur) ? '' : String(cur)
       if (shouldUpdateValue(elm, strCur)) {
         elm.value = strCur
       }
--- a/src/platforms/web/runtime/modules/dom-props.js
+++ b/src/platforms/web/runtime/modules/dom-props.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { isDef, isUndef, extend, toNumber } from 'shared/util'
+import { isDef, isUndef, extend, toNumber, isNumNegZero } from 'shared/util'
 import { isSVG } from 'web/util/index'
 
 let svgContainer
@@ -44,7 +44,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
       // non-string values will be stringified
       elm._value = cur
       // avoid resetting cursor position when value is the same
-      const strCur = isUndef(cur) ? '' : String(cur)
+      const strCur = isNumNegZero(cur) ? '-0' : (isUndef(cur) ? '' : String(cur))
       if (shouldUpdateValue(elm, strCur)) {
         elm.value = strCur
       }
--- a/src/platforms/web/runtime/modules/dom-props.js
+++ b/src/platforms/web/runtime/modules/dom-props.js
@@ -44,7 +44,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
       // non-string values will be stringified
       elm._value = cur
       // avoid resetting cursor position when value is the same
-      const strCur = isNumNegZero(cur) ? '-0' : (isUndef(cur) ? '' : String(cur))
+      const strCur = isNumNegZero(cur) ? '-0' : isUndef(cur) ? '' : String(cur)
       if (shouldUpdateValue(elm, strCur)) {
         elm.value = strCur
       }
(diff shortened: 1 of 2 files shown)


src/platforms/web/runtime/modules/dom-props.js
```diff
@@ -44,7 +44,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
       // non-string values will be stringified
       elm._value = cur
       // avoid resetting cursor position when value is the same
-      const strCur = isUndef(cur) ? '' : String(cur)
```
yuwu9145, line 47: Input tag should be able to display -0. However, String(-0) will result in 0.

src/shared/util.js
```diff
@@ -8,6 +8,10 @@ export function isUndef (v: any): boolean %checks {
   return v === undefined || v === null
 }
 
+export function isNumNegZero (v: any): boolean %checks {
+  return !isNaN(v) && 1/Number(v) === -Infinity
```
yuwu9145, line 12: According [this eslint rule](https://eslint.org/docs/rules/no-compare-neg-zero), `Number(v) === -0` shall not be used. In addition, due to `Object.is` is not supported by IE. Therefore, `1/Number(v) === -Infinity` is used to determine whether `Number(v)` is `-0`

sirlancelot: In my testing, the `isNaN()` check is unnecessary, as is the `Number(v)` conversion. `return (1 / v) === -Infinity` should do it.

src/platforms/web/runtime/modules/dom-props.js
```diff
@@ -44,7 +44,7 @@ function updateDOMProps (oldVnode: VNodeWithData, vnode: VNodeWithData) {
...
+      const strCur = isNumNegZero(cur) ? '-0' : isUndef(cur) ? '' : String(cur)
```
sirlancelot, line 47: I think in the interest of least impactful, this collection of ternaries should keep the `isUndef()` at the top like so:

```js
const strCur = isUndef(cur) ? '' : isNumNegZero(cur) ? '-0' : String(cur)
```

This keeps the case of `cur` being empty as the fastest/shortest path.

Justineo:
Hi, can you add some test cases to reflect these changes? (which only affects `v-model.number` while `v-model`'s original behavior is kept, according to the discussions in #11909)

yuwu9145:
> Hi, can you add some test cases to reflect these changes? (which only affects `v-model.number` while `v-model`'s original behavior is kept, according to the discussions in #11909)

HI @Justineo , yes, I am happy to add test case. I will also consider and include potential changes suggested by @sirlancelot

posva:
As said, I don't think it should modify don props, which is perf critical. It should only apply to vmodel number

yuwu9145:
> As said, I don't think it should modify don props, which is perf critical. It should only apply to vmodel number

@posva What about making `v-model` & `v-model.number` output `'-0': string` only for `-0: number`? Then pass `'-0': string` to `dom-props`? So it does not have to make any change in `dom-props`.

yuwu9145:
After thinking through all concerns & reminding this is a very rare edge case. Probably let `v-model.number` handle `-0:number` is enough, so the only change has been made is:

`-0: number -> v-model.number -> '-0': string`

Please advice if more test case or changes are required

src/shared/util.js
```diff
@@ -92,10 +96,11 @@ export function toString (val: any): string {
 
 /**
  * Convert an input value to a number for persistence.
+ * Check for special case of signed zero.
  * If the conversion fails, return original string.
  */
 export function toNumber (val: string): number | string {
-  const n = parseFloat(val)
+  const n = isNumNegZero(val) ? '-0' : parseFloat(val)
```
posva, line 103: let's inline the check and remove the function

```suggestion
  const n = 1 / val === -Infinity ? '-0' : parseFloat(val)
```

posva (changes requested):
I don't think this is still good as, I will say it again, it should be an edge case of vmodel. It's also important that v-model.number sets the value as `-0` and not `'-0'` when typing `-0`. Could you add both unit tests?

yuwu9145:
@posva the problem is if `v-model.number` outputs `-0: number`, dom-props will do `String(-0)` and result in '0: string' still. I really think a fix should be done in dom-props as well. But if it has a performance concern, then it has to make a decision between **performance** and **"-0 support"**

posva:
It's not just perf, it's also being consistent with js APIs and not breaking the existing v-model.number when the input has `-0` in it. To be fair, this is still just a small patch, it won't allow people to have a number input that handles the other mentioned edge cases, so I think it's better to just drop the idea. Thanks a lot for your efforts!

yuwu9145:
@posva no worries. It's good know the concern

PR #11943, opened 2021-03-07
fix: pause dep collection during immediate watcher invocation

### LINKED ISSUE #11942: Dep collection not paused during first invocation of immediate watchers, leading to unnecessary re-render
### Version
2.6.11

### Reproduction link
[https://jsfiddle.net/gyejmL14/32/](https://jsfiddle.net/gyejmL14/32/)

### Steps to reproduce
Press the button "change value of testKey".

### What is expected?
No components should be updated

The root components render function is being run again.

During component-a's creation, the immediate watcher for testKey is invoked without a pushTarget(). At this point the updateComponent() of the parent is still the Dep.target. Everything accessed within the immediate watcher is then collected as a dependency of the parents updateComponent watcher. When the button is then pressed, this triggers an unnecessary update of the Root component.

delaneyb:
Close #11942

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: **Performance improvement**

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/test/unit/features/instance/methods-lifecycle.spec.js
+++ b/test/unit/features/instance/methods-lifecycle.spec.js
@@ -81,7 +81,7 @@ describe('Instance methods lifecycle', () => {
       expect(calls).toBe(1)
       childData.a++
       waitForUpdate(() => {
-        expect(parentUpdate.calls.count()).toBe(0)
+        expect(parentUpdate).not.toHaveBeenCalled()
       }).then(done)
     })
   })
--- a/test/unit/features/instance/methods-lifecycle.spec.js
+++ b/test/unit/features/instance/methods-lifecycle.spec.js
@@ -53,6 +53,31 @@ describe('Instance methods lifecycle', () => {
         }
       }).$mount()
     })
+
+    it('Dep.target should be undefined during invocation of child immediate watcher', () => {
+      let calls = 0;
+      new Vue({
+        template: '<div><my-component></my-component></div>',
+        components: {
+          myComponent: {
+            template: '<div>hi</div>',
+            data() {
+              return { a: 123 }
+            },
+            watch: {
+              a: {
+                handler() {
+                  ++calls
+                  expect(Dep.target).toBe(undefined)
+                },
+                immediate: true
+              }
+            }
+          }
+        }
+      }).$mount()
+      expect(calls).toBe(1)
+    })
   })
 
   describe('$destroy', () => {
--- a/test/unit/features/instance/methods-lifecycle.spec.js
+++ b/test/unit/features/instance/methods-lifecycle.spec.js
@@ -54,15 +54,17 @@ describe('Instance methods lifecycle', () => {
       }).$mount()
     })
 
-    it('Dep.target should be undefined during invocation of child immediate watcher', () => {
-      let calls = 0;
+    it('Dep.target should be undefined during invocation of child immediate watcher', done => {
+      let calls = 0, childData
+      const parentUpdate = jasmine.createSpy()
       new Vue({
         template: '<div><my-component></my-component></div>',
+        updated: parentUpdate,
         components: {
           myComponent: {
-            template: '<div>hi</div>',
+            template: '<div>{{ a }}</div>',
             data() {
-              return { a: 123 }
+              return childData = { a: 123 }
             },
             watch: {
               a: {
@@ -77,6 +79,10 @@ describe('Instance methods lifecycle', () => {
         }
       }).$mount()
       expect(calls).toBe(1)
+      childData.a++
+      waitForUpdate(() => {
+        expect(parentUpdate.calls.count()).toBe(0)
+      }).then(done)
     })
   })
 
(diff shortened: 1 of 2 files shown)


posva (changes requested):
can you add a unit test?

test/unit/features/instance/methods-lifecycle.spec.js
```diff
@@ -53,6 +53,31 @@ describe('Instance methods lifecycle', () => {
...
+          myComponent: {
+            template: '<div>hi</div>',
+            data() {
+              return { a: 123 }
+            },
+            watch: {
+              a: {
+                handler() {
+                  ++calls
+                  expect(Dep.target).toBe(undefined)
```
posva, line 71: Is this necessary?
I think I shouldn't have said unit test. A regression test closer to what you provided as a reproduction would be perfect

delaneyb: Is this more like what you were thinking?
image

delaneyb:
Related Vue 3 PR: https://github.com/vuejs/vue-next/pull/2765

I see the solution there pauses tracking during both the immediate and subsequent invocations, whereas my issue here covers only that immediate case. Would wrapping the calls here too interfere with other uses of the Watcher class?

https://github.com/vuejs/vue/blob/5255841aaff441d275122b4abfb099b881de7cb5/src/core/observer/watcher.js#L193-L201

test/unit/features/instance/methods-lifecycle.spec.js
posva: Yes, that looks like a better test for this scenario

test/unit/features/instance/methods-lifecycle.spec.js
```diff
@@ -53,6 +53,37 @@ describe('Instance methods lifecycle', () => {
...
+        updated: parentUpdate,
+        components: {
+          myComponent: {
+            template: '<div>{{ a }}</div>',
+            data() {
+              return childData = { a: 123 }
+            },
+            watch: {
+              a: {
+                handler() {
```
posva, line 71: let's use a spy here too since the Dep.target is an implementation detail

test/unit/features/instance/methods-lifecycle.spec.js
```diff
@@ -53,6 +53,37 @@ describe('Instance methods lifecycle', () => {
...
+                immediate: true
+              }
+            }
+          }
+        }
+      }).$mount()
+      expect(calls).toBe(1)
+      childData.a++
+      waitForUpdate(() => {
+        expect(parentUpdate.calls.count()).toBe(0)
```
posva, line 84:
```suggestion
        expect(parentUpdate).not.toHaveBeenCalled()
```

or

```suggestion
        expect(parentUpdate).toHaveBeenCalledTimes(0)
```

posva (changes requested):

test/unit/features/instance/methods-lifecycle.spec.js
delaneyb: Are you suggesting not having our own function there? When the watcher callback is called it either needs to check Dep.target or touch childData.a to cause the issue. Should I just replace the Dep.target check with a `let x = this.a`, and then we rely on the `expect(parentUpdate).not.toHaveBeenCalled()`?

test/unit/features/instance/methods-lifecycle.spec.js
```diff
@@ -53,6 +53,37 @@ describe('Instance methods lifecycle', () => {
...
+      let calls = 0, childData
+      const parentUpdate = jasmine.createSpy()
+      new Vue({
+        template: '<div><my-component></my-component></div>',
+        updated: parentUpdate,
+        components: {
+          myComponent: {
+            template: '<div>{{ a }}</div>',
+            data() {
+              return childData = { a: 123 }
```
posva, line 67: Let's move the data variable outside of new Vue, and return it here

```suggestion
              return childData
```

test/unit/features/instance/methods-lifecycle.spec.js
```diff
@@ -53,6 +53,37 @@ describe('Instance methods lifecycle', () => {
...
+            },
+            watch: {
+              a: {
```
posva, line 70: Here it should be something else
```suggestion
              anything: {
```

test/unit/features/instance/methods-lifecycle.spec.js
```diff
@@ -53,6 +53,37 @@ describe('Instance methods lifecycle', () => {
         }
       }).$mount()
     })
+
+    it('Dep.target should be undefined during invocation of child immediate watcher', done => {
+      let calls = 0, childData
```
posva, line 58:
```suggestion
      let calls = 0
      const childData = { a: 1 }
```

test/unit/features/instance/methods-lifecycle.spec.js
posva: I'm realizing I was giving you hints as if they were jest spies but they are not. The `toHaveBeenCalled()` seems to work though. There is probably a way to mock the implementation of the spy but it's not necessary, it might be an outdated version of Jasmine.
Reading `this.a` is better, yes

posva (changes requested):

posva (approved):
Thanks!

PR #11963, opened 2021-03-16
fix: handle async placeholders in normalizeScopedSlot 

### LINKED ISSUE #10391: Nested async components passed as slots are not rendered in SSR
### Version
2.6.10

### Reproduction link
[https://codesandbox.io/s/vue-ssr-nested-async-components-repro-script-ozze7](https://codesandbox.io/s/vue-ssr-nested-async-components-repro-script-ozze7)

### Steps to reproduce
When the codesandbox starts up, it should run the ```npm start``` script. If it doesn't, open a bash terminal and run ```npm start``` to see the HTML generated from SSR. You can also take a look at the components in the ```src``` folder to see how they are configured.

### What is expected?
Expected the HTML for the nested async component (called "SomeComponent") to be rendered during SSR:

```html
<div data-server-rendered="true">
  <div class="Grid">
    <div class="Grid__left">Left
      <div class="SomeComponent">Somecomponent with value: first</div>
    </div>
    <div class="Grid__right">Right
      <div class="SomeComponent">Somecomponent with value: second</div>
    </div>
  </div>
</div>
```

The HTML for the nested async component (called "SomeComponent") is not rendered during SSR. Only the HTML for the outer async component (called "Grid" ) is rendered:

```html
<div data-server-rendered="true">
  <div class="Grid">
    <div class="Grid__left">Left
    </div>
    <div class="Grid__right">Right
    </div>
  </div>
</div>
```

---
I have an App that is composed of many async components. Some of these components have named slots. I was trying to build a render function that will accept a config object to render these components in various setups (for example, an async component set to a named slot of another async component). I noticed that if I nest async components within another, the HTML for the async component does not render during SSR. The codesandbox is simplified scenario of this

pi0:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Context:**

- Reproduction: https://replit.com/@pi0/vuejsvue10391
- resolves #10391, resolves nuxt/nuxt.js#8904. resolves nuxt/components#164

### DIFF
--- a/src/core/vdom/helpers/normalize-scoped-slots.js
+++ b/src/core/vdom/helpers/normalize-scoped-slots.js
@@ -3,6 +3,7 @@
 import { def } from 'core/util/lang'
 import { normalizeChildren } from 'core/vdom/helpers/normalize-children'
 import { emptyObject } from 'shared/util'
+import { isAsyncPlaceholder } from './is-async-placeholder'
 
 export function normalizeScopedSlots (
   slots: { [key: string]: Function } | void,
@@ -60,9 +61,10 @@ function normalizeScopedSlot(normalSlots, key, fn) {
     res = res && typeof res === 'object' && !Array.isArray(res)
       ? [res] // single vnode
       : normalizeChildren(res)
+    let vnode: VNode = res && res[0]
     return res && (
-      res.length === 0 ||
-      (res.length === 1 && res[0].isComment) // #9658
+      !vnode ||
+      (vnode.isComment && !isAsyncPlaceholder(vnode)) // #9658, #10391
     ) ? undefined
       : res
   }
--- a/src/core/vdom/helpers/normalize-scoped-slots.js
+++ b/src/core/vdom/helpers/normalize-scoped-slots.js
@@ -61,7 +61,7 @@ function normalizeScopedSlot(normalSlots, key, fn) {
     res = res && typeof res === 'object' && !Array.isArray(res)
       ? [res] // single vnode
       : normalizeChildren(res)
-    let vnode: VNode = res && res[0]
+    let vnode: VNode = res && res.length === 1 && res[0]
     return res && (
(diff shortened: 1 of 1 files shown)


src/core/vdom/helpers/normalize-scoped-slots.js
```diff
@@ -60,9 +61,10 @@ function normalizeScopedSlot(normalSlots, key, fn) {
     res = res && typeof res === 'object' && !Array.isArray(res)
       ? [res] // single vnode
       : normalizeChildren(res)
+    let vnode: VNode = res && res[0]
```
posva, line 64: shouldn't we keep the length check of 1?
```suggestion
    let vnode: VNode = res && res.length === 1 && res[0]
```

posva (commented):
Thanks for this!

src/core/vdom/helpers/normalize-scoped-slots.js
```diff
@@ -60,9 +61,10 @@ function normalizeScopedSlot(normalSlots, key, fn) {
     res = res && typeof res === 'object' && !Array.isArray(res)
       ? [res] // single vnode
       : normalizeChildren(res)
+    let vnode: VNode = res && res.length === 1 && res[0]
```
posva, line 64:
```suggestion
    let vnode: VNode = res && res[0]
```

I was wrong I guess

patrioticcow:
Can we merge this in? I need this

src/core/vdom/helpers/normalize-scoped-slots.js
caugner: This conversation was marked as resolved, but the suggested change didn't make it into the MR, and possibly caused #12102.

PR #11967, opened 2021-03-18
Make an original event context accessible as [$el] in events

thisYogesh:

- [ ] Bugfix
- [x] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**PR details:**
Currently we don't have direct access to the original event context i.e. an element upon which event has been applied. This feature is kind of an alternative for `ref` but not complete replacement, of course, And I think people would like to use it.

**Usage of this feature:**
- `<div id="item" @click="handler($el)"></div>` 
- Where `$el` is the respective element i.e. `#item`.


posva:
Thank you for the proposal. I don't think we should add this: the `$event` holds already the target. If you think this should be added to Vue, you should open an RFC in our rfcs repository

thisYogesh:
Thanks for your response @posva!
I'm aware of `$event` and `event target`, but `event target` is not always the element upon which event is applied!

PR #12298, opened 2021-09-26
refactor(observer): Hijack when taking value

jianglailai:
Why is it necessary to recurse as soon as you get it? Can't you hijack the property when you get it

### DIFF
--- a/src/core/observer/index.js
+++ b/src/core/observer/index.js
@@ -153,11 +153,12 @@ export function defineReactive (
     val = obj[key]
   }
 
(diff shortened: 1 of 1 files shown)


posva:
Thanks but this is simply not the same

jianglailai:
Can you explain in detail why it can't be like this? thank you

Justineo:
It should be the other way round: you are proposing a change to the code that worked well for a long time so you should explain what improvements does the change brings in the first place.

PR #12455, opened 2022-02-03
fix(core): add prop name to invalid prop type warning message (fix #12454)

### LINKED ISSUE #12454: Property validator should report the property that failed type / validator check
Components with multiple properties might fail property validation at the most inopportune time: assigning one property will trigger property validation for all `function initProps (vm, propsOptions)`. Stack trace will show that the issue was triggered by assigning a specific property "prop1" while validation will actually fail on "prop2". If this is combined with component actually rendering for the first time ("prop1" was the trigger), it will be unclear what property actually failed.

In my case, reported error message was:

    [Vue warn]: Invalid prop type: "[object Object]" is not a constructor

This is from method `function assertType (value, type, vm)`

Clarify for the actual n00b error message I received: I specified prop type to be of an "enum" class declared with Object.freeze. Actual values are, naturally, Number.

It would be clearer to add prop name to the message like this:

    [Vue warn]: Invalid prop "propName" type: "[object Object]" is not a constructor

This would require to add `propName` (or just `name`) parameter to `assertType` in order to be able to include it in the message

espresso:
Previously, the "Invalid prop type" warning message returned by the property validator did not
include the name of the offending property. In certain scenarios this obscurity can make it
difficult to identify the source of the issue. To address this, I have added the name of the
offending property to the output as suggested in #12454.

fix #12454

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/core/util/props.js
+++ b/src/core/util/props.js
@@ -122,7 +122,7 @@ function assertProp (
       type = [type]
     }
     for (let i = 0; i < type.length && !valid; i++) {
-      const assertedType = assertType(value, type[i], vm)
+      const assertedType = assertType(value, type[i], name, vm)
       expectedTypes.push(assertedType.expectedType || '')
       valid = assertedType.valid
     }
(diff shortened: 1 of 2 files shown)


sakupan102:
hello is it still open i would love to work on this

PR #12470, opened 2022-02-17
changed to GPLv3 license

libererer:


Justineo:
We have no plans to change our licence, thank you.

PR #12551, opened 2022-06-03
fix(core): hydrated node assertion in production

Timsonrobl:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
The `tag name` and `type` of hydration candidate nodes are currently checked only in development mode, for some reason.
This causes instability in production mode as TEXT type node can be chosen as hydration target for ELEMENT type node if there is a DOM structure mismatch, which can cause app crash when hydration process tries to access .appendChild() method non existing in text node throwing an unhandled exception and making the whole app unresponsive.

### DIFF
--- a/src/core/vdom/patch.js
+++ b/src/core/vdom/patch.js
@@ -604,10 +604,8 @@ export function createPatchFunction (backend) {
       return true
     }
     // assert node match
-    if (process.env.NODE_ENV !== 'production') {
-      if (!assertNodeMatch(elm, vnode, inVPre)) {
-        return false
-      }
+    if (!assertNodeMatch(elm, vnode, inVPre)) {
+      return false
     }
     if (isDef(data)) {
       if (isDef(i = data.hook) && isDef(i = i.init)) i(vnode, true /* hydrating */)


Timsonrobl:
Any hints on how to make reproduction would be appreciated. Is there any sandbox with SSR out there?

Basically you need to make a render function that renders conditionally:
On the server:
```HTML
<div>Text node</div>
```
On the client:
```HTML
<div>
  <a>Text node</a>
</div>
```
and it should crash the Vue instance in production mode.

You can also Google " Nuxt Failed to execute 'appendChild' on 'Node': This node type does not support this method." and most of the results would be symptoms of this bug.
Also lots of issues in the Nuxt repo for [this search query](https://github.com/nuxt/nuxt.js/issues?q=is%3Aissue+Failed+to+execute+%27appendChild%27+on+%27Node).

posva:
It’s normal for the app to crash. You are supposed to catch that kind of bug during development. The check is removed for perf reasons

PR #12564, opened 2022-06-20
chore(compiler-sfc): replace `substr` with `slice`

webfansplz:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The `substr` is deprecated. 
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

### DIFF
--- a/packages/compiler-sfc/src/templateCompilerModules/srcset.ts
+++ b/packages/compiler-sfc/src/templateCompilerModules/srcset.ts
@@ -38,7 +38,7 @@ function transform(
         }
 
         const imageCandidates: ImageCandidate[] = value
-          .slice(1, value.length - 2)
+          .slice(1, value.length - 1)
           .split(',')
           .map(s => {
             // The attribute value arrives here with all whitespace, except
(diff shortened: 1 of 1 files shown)


packages/compiler-sfc/src/templateCompilerModules/srcset.ts
```diff
@@ -38,7 +38,7 @@ function transform(
         }
 
         const imageCandidates: ImageCandidate[] = value
-          .substr(1, value.length - 2)
+          .slice(1, value.length - 2)
```
posva, line 41: I think this becomes

```suggestion
          .slice(1, value.length - 1)
```

jacekkarczmarczyk:
isn't `-1` even better?

image

packages/compiler-sfc/src/templateCompilerModules/srcset.ts
```diff
@@ -38,7 +38,7 @@ function transform(
         }
 
         const imageCandidates: ImageCandidate[] = value
-          .substr(1, value.length - 2)
+          .slice(1, value.length - 1)
```
BlackCrowxyz, line 41:
```suggestion
          .slice(1, -1)
```

posva (approved):

PR #12571, opened 2022-06-22
fix: types/v3-directive not published

lsdsjy:
Currently running TS type-checking will report error:
`vue/types/index.d.ts:70:8 - error TS2307: Cannot find module './v3-directive' or its corresponding type declarations.`


xiaoxiangmoe:
Looks good to me！

Issue #12745, opened 2022-08-16
use decorator before export default may cause compilation problems


linshuohao:
### Version
2.7.8

### Reproduction link
[stackblitz.com](https://stackblitz.com/edit/vitejs-vite-t9j2nw)

### Steps to reproduce
After open the above reproduction, you will see the console shows the error: [vite] Internal server error: Cannot overwrite across a split point

### What is expected?
rewriteDefault can transform
```
@Component
export default class App extends Vue {
  a = `;export xxx default`;
}
```
into
```
@Component
class App extends Vue {
  a = `;export xxx default`;
}
const ${as} = App
```
so that vite can work correctly

### What is actually happening?
same as https://github.com/vuejs/core/issues/6318

PR #12747, opened 2022-08-16
fix(compiler-sfc): make sure rewriteDefault get current result when u…

linshuohao:
fix https://github.com/vuejs/vue/issues/12745


yyx990803:
Thanks! Can you submit the same PR for Vue 3 as well?

yyx990803:
Oh nvm, I see the PR.

Issue #12892, opened 2022-12-01
[2.7.14][class component][sfc][vite] Method names containing "export" and "default" break lifecycle hooks


hmkrivoj:
### Version
2.7.14

### Reproduction link
[codesandbox.io](https://codesandbox.io/s/eloquent-tess-oconp9?file=/src/components/HelloWorld.vue)

### Steps to reproduce
1. Create a typescript/vite/vue 2 app with these dependencies
```json
{
  "dependencies": {
    "vue": "^2.7.14",
    "vue-class-component": "^7.2.6"
  },
  "devDependencies": {
    "@types/node": "^16.11.45",
    "@vitejs/plugin-legacy": "^2.0.0",
    "@vitejs/plugin-vue2": "^1.1.2",
    "@vue/tsconfig": "^0.1.3",
    "npm-run-all": "^4.1.5",
    "terser": "^5.14.2",
    "typescript": "~4.7.4",
    "vite": "^3.0.2",
    "vue-tsc": "^0.38.8"
  }
}
```

2. Create a SFC class component and include it into your app. Add methods `mounted()`, `exportieren()` and `defaultWert()`:
```vue
<script lang="ts">
import Vue from "vue";
import Component from "vue-class-component";
@Component({})
export default class HelloWorld extends Vue {
  test = "";
  mounted() {
    console.log("mounted!");
    this.test = "Hallo Welt!";
  }
  exportieren(): void {
    // do nothing
  }
  defaultWert(): void {
    // do nothing
  }
}
</script>

<template>
  <p>{{ test }}</p>
</template>
```

### What is expected?
`exportieren()` and `defaultWert()` should not prevent lifecycle hooks from being called.

### What is actually happening?
Lifecycle hook `mounted()` is not called. Renaming/rearranging/removing `exportieren()` and `defaultWert()` fixes the issue.

---
I'm guessing it's a problem with the sfc-compiler, as I can't reproduce this bug with a plain js/ts component:
```typescript
import { Component, Vue } from "vue-property-decorator";
@Component({
    template: "<p>{{ test }}</p>"
})
export default class HelloWorld extends Vue {
	test = "";
	mounted() {
	  console.log("mounted!");
	  this.test = "Hallo Welt!";
	}
	exportieren(): void {
	  // do nothing
	}
	defaultWert(): void {
	  // do nothing
	}
}
```

[...]

PR #12906, opened 2022-12-19
fix(compiler-sfc): fix rewriteDefault problem when using @babel/parser@^7.20.0

linshuohao:
close https://github.com/vuejs/vue/issues/12892

https://github.com/babel/babel/pull/15032 change the location of the export declaration when use decorators before export

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "csstype": "^3.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks for the PR, but the lockfile updates are a bit difficult to resolve directly from your branch.

landed separately in 10c2a87c and 25f97a50 (which also uses the exact same logic from Vue 3)

Issue #3761, opened 2016-09-25
v-show should be !important


codysherman:
### Vue.js version
2.0.0-rc.7

### Reproduction Link
http://codepen.io/codysherman/pen/qaArqJ?editors=1010

### Steps to reproduce
Have an element already have an !important display trait, such as one of Bootstrap's 'visible-*' classes.

### What is Expected?
Vue should hide the element based on the v-show.

### What is actually happening?
Because v-show doesn't have an !important tag, it is overruled. In my opinion, in the event of a conflict, the tie should obviously go to Vue, as it is getting its logic dynamically. Angular's ng-show applies the !important tag for this reason.

sylvainpolletvillard:
Please no.  The *!important* level is already used in some components to purposely override Vue v-show behaviour. The current level as inline style is the most appropriate.

Just put the condition on your *visible-* class instead of using v-show. It makes no sense to have a hidden element with some *visible-* classes anyway.

posva:
As @sylvainpolletvillard said, if you want to apply `!important` you should create a class for it.

BrainBacon:
@posva v-hide does not apply a class to the element when hidden. If it did this would be a non-issue. Wouldn't a class have to be applied via v-bind? Seems like a contrived way to go about hiding an element and would make v-hide completely useless when doing responsive design in a popular css framework.

posva:
`v-hide` doesn't exist but you can create your own using the [v-show directive](https://github.com/vuejs/vue/blob/dev/src/directives/public/show.js) as a reference. You can even publish it to npm a post it here 😉 
Yes, the class have to be applied with a v-bind

BrainBacon:
@posva Sorry I meant `v-show`, sounds like I had a bit of a Freudian slip haha. That being said and expanding on what @sylvainpolletvillard said, is component overriding directives really something that should be widely supported? Seems to me something like that would be an edge-case that should be either handled by Vue's internal logic or some kind of decorator whereas something as common as responsive framework usage should be supported? I assume there's also a good reason why the directive doesn't apply a class for easy selecting, but I don't know enough to say what that is? Perhaps you can shed a bit of light on the reasoning?

posva:
By applying classes I mean something like this:

``` html
<div :class="{hide: hide}"></div>
```

``` css
.hide {
  display: none !important;
}
```

Maybe a modifier to `v-show` can come in handy `v-show.important=""`

sylvainpolletvillard:
[...]

PR #12942, opened 2023-01-24
Feat/show important directive

ajomuch92:
Adding support for `important `modifier on `v-show` directive

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] It's submitted to the `main` branch for v2.x (or to a previous version branch)

The explanation to include this feature is when you have a component (from an external library or own component) and this component has a display flex (or other display property on CSS), the v-show directive doesn't work showing the content you are trying to hide/show. You can make a workaround, like switching the class with the condition used into the directive, but it could be easier for developers only add this modifier

### DIFF
--- a/packages/server-renderer/src/directives/show.ts
+++ b/packages/server-renderer/src/directives/show.ts
@@ -4,9 +4,9 @@ export default function show(node: VNodeWithData, dir: VNodeDirective) {
   if (!dir.value) {
     const style: any = node.data.style || (node.data.style = {})
     if (Array.isArray(style)) {
-      style.push({ display: 'none' })
+      style.push({ display: dir?.modifiers?.important ? 'none !important' : 'none' })
     } else {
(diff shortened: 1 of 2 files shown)


posva:
Did you see this idea was already rejected at https://github.com/vuejs/vue/issues/3761 ?

ajomuch92:
@posva Yes, I read that issue, but I think this is relevant, because I have met with this problem many times, and other devs faced this too. 

I think it's not big deal to implement it, another library did it

ajomuch92:
@posva Also, if we implement a custom directive to display none important, this new one doesn't trigger the transition classes (if it's wrapped by a transition component), keep in mind

ajomuch92:
@posva any news from this PR?

PR #13107, opened 2023-10-24
fix: type augmentation and compiler-sfc types w/moduleResolution: bundler (fix #13106)

### LINKED ISSUE #13106: Type augmentation is broken when using typescript's moduleResolution: "Bundler" option
### Version
2.7.15

### Reproduction link
[codesandbox.io](https://codesandbox.io/p/sandbox/bold-resonance-ssg894)

### Steps to reproduce
```bash
yarn install
yarn test
```

### What is expected?
The project should typecheck without errors

The type augmentation fails and as a result, there are type errors that need to be worked around.

---
When using typescript's "moduleResolution": "bundler" option, we currently get a bunch of type errors, particularly when trying to extend the types, or when consuming other packages which extend the types (vue-router, pinia, etc).

For example, vue-router extends the types as follows:  
```ts
declare module 'vue/types/vue' {
  interface Vue {
    $router: VueRouter
    $route: Route
  }
}

declare module 'vue/types/options' {
  interface ComponentOptions<V extends Vue> {
    router?: VueRouter
    beforeRouteEnter?: NavigationGuard<V>
    beforeRouteLeave?: NavigationGuard<V>
    beforeRouteUpdate?: NavigationGuard<V>
  }
}
```

But this silently fails, and the types aren't available on the vue constructor options or the component instance.

image  

```
Object literal may only specify known properties, and 'router' does not exist in type 'ComponentOptions<Vue<Record<string, any>, Record<string, any>, never, never, (event: string, ...args: any[]) => Vue<Record<string, any>, Record<string, any>, never, never, ...>>, ... 7 more ..., ComponentOptionsMixin>
```

The reason for this can be seen when copying the code into your own project with `"moduleResolution": "bundler"`:

image

```
Invalid module name in augmentation, module 'vue/types/options' cannot be found.ts(2664)
module "vue/types/options"
```

What is happening is that once "moduleResolution" is set to "bundler", the "exports" maps in package.json starts being used, and there are some peculiarities with wildcard matching that we don't appear to be accounting for here. In particular, wildcard matching is going to require that the full path to the file is provided. So, when we specify `declare module 'vue/types/vue'` it is going to look for a file on disk with the relative path: `./types/vue`, and since there is no extensionless file with that name, it fails.

You could technically work around this by explicitly specifying `declare module 'vue/types/vue.js'` or `declare module 'vue/types/vue.d.ts'` when you are doing the type augmentation, but I think it would be better to just fix the "exports" definition to allow you to specify the file name without the extension as otherwise we'd need to release updates to many other packages and user code vs just one.

thebanjomatic:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] It's submitted to the `main` branch for v2.x (or to a previous version branch)

**Other information:**

Each commit contains a pretty lengthy description to try to make things as clear as possible. Rather than reproduce that here, the short version is that this PR contains two bug fixes:
1) Fixes #13106 which deals with type augmentation being broken in the ecosystem when enabling the "moduleResolution": "bundler" typescript option. This is due to the export map requiring full filenames with extensions, even though they are almost never provided as such
2) Fixes an issue with the types being unavailable from a `"type": "module"` context when using "moduleResolution": "bundler" | "node16"

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,12 @@
   ],
   "exports": {
     ".": {
+      "types": "./types/index.d.ts",
       "import": {
         "node": "./dist/vue.runtime.mjs",
         "default": "./dist/vue.runtime.esm.js"
       },
-      "require": "./dist/vue.runtime.common.js",
-      "types": "./types/index.d.ts"
+      "require": "./dist/vue.runtime.common.js"
     },
     "./compiler-sfc": {
       "types": "./compiler-sfc/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
       "types": "./types/index.d.ts"
     },
     "./compiler-sfc": {
+      "types": "./compiler-sfc/index.d.ts",
       "import": "./compiler-sfc/index.mjs",
       "require": "./compiler-sfc/index.js"
     },
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
       "require": "./compiler-sfc/index.js"
     },
     "./dist/*": "./dist/*",
-    "./types/*": "./types/*",
+    "./types/*": ["./types/*.d.ts", "./types/*"],
     "./package.json": "./package.json"
   },
   "sideEffects": false,


package.json
```diff
@@ -18,19 +18,20 @@
...
+      "require": "./dist/vue.runtime.common.js"
     },
     "./compiler-sfc": {
+      "types": "./compiler-sfc/index.d.ts",
       "import": "./compiler-sfc/index.mjs",
       "require": "./compiler-sfc/index.js"
     },
     "./dist/*": "./dist/*",
-    "./types/*": "./types/*",
+    "./types/*": ["./types/*.d.ts", "./types/*"],
```
thebanjomatic, line 34: To help clarify the conversion from import name to file lookups, I tried to follow the pseudo-code from the documentation here: https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports and list out what would happen in the the following scenarios:

Before this change:
```
"vue/types/options" => ["./types/options"] // fails to find types
"vue/types/options.js" => ["./types/options.ts", "./types/options.d.ts", "./types/options.js"] // finds on second try
"vue/types/options.d.ts" => ["./types/options.d.ts"] // finds on first try
```

After this change:
```
"vue/types/options" => ["./types/options.d.ts", "./types/options"] // finds on first try
"vue/types/options.js" => ["./types/options.js.d.ts", "./types/options.ts", "./types/options.d.ts", "./types/options.js"] // finds on third try
"vue/types/options.d.ts" => ["./types/options.d.ts.d.ts", "./types/options.d.ts"] // finds on second try
```

Since the most common case is to write these imports without the extensions, I thought it would be best to write this as `["./types/*.d.ts", "./types/*"]`. This makes the common case more direct to resolve.

thebanjomatic:
@yyx990803 Hey Evan is there any chance that this could be considered for a future vue 2.7.16 release before it reaches the end of the support window? This issue has been starting to show up more and more as a blocker recently as I have been running into cases where some libraries (such as Got) require configuring "moduleResolution": "bundler", but a bunch of stuff in the vue 2.7 ecosystem stops working when you enable this option. The actual changes required to fix things are minimal and just amount to small changes to the package.json exports map, and I tried to be extra thorough in my descriptions of what these things do to provide confidence in the overall safety of making these changes.

For more in depth information, please see the attached issue (which contains a reproducer), or the commit message [here](https://github.com/vuejs/vue/pull/13107/commits/7458a5af98583606a64ddd2e98dfefa5aa5d4038)

PR #13289, opened 2025-11-12
docs: make README top-level H1 to satisfy markdownlint

maynkxx:
Make the README's first line a top-level heading (#) so it complies with markdownlint rule MD041 (first-line-heading). This is a small documentation-only change that fixes linter warnings.


Burnsedia:
I can do a code review

Issue #12965, opened 2023-02-21
Nested pre tag working exception in vue-template-compiler


daysv:
### Version
2.7.14

### Reproduction link
[codepen.io/daysv/pen/JjaGLVP](https://codepen.io/daysv/pen/JjaGLVP)

### Steps to reproduce
In pure html: 
https://codepen.io/daysv/pen/yLxeKqj

Demo:
```js
const compiler = require('vue-template-compiler')

const str = `<div v-pre>
1

2

<div style="padding-left: 20px">aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
c
</div>
3
</div>`

const result = compiler.compile(str, { whitespace: 'condense' })
console.log(result.staticRenderFns)
```

### What is expected?
`with(this){return _c('pre',[_v("1\\n\\n2\\n\\n"),_c('pre',{staticStyle:{"padding-left":"20px"}},[_v("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaa\\nbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb\\nc\\n")]),_v("\\n3\\n")])}`

### What is actually happening?
 `with(this){return _c('pre',[_v("1\\n\\n2\\n\\n"),_c('pre',{staticStyle:{"padding-left":"20px"}},[_v("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaa\\nbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb\\nc\\n")]),_v(" 3 ")])}`

---
I think `inPre` should use number instaed of bool.

```js
let inPre = 0;

// ....

if (platformIsPreTag(element.tag)) {
  inPre --;
}

// ...

if (platformIsPreTag(element.tag)) {
  inPre ++;
}
```

Subham-KRLX:
Sir i would like to fix this issue can you please assign it to me.

seekskyworld:
[...]

PR #13299, opened 2026-01-14
fix: preserve nested pre whitespace

seekskyworld:
Issue link: https://github.com/vuejs/vue/issues/12965

Summary:
- track nested <pre> depth so condense mode keeps whitespace inside the outer <pre>
- add regression coverage for nested <pre> with whitespace: 'condense'

Motivation:
- avoid trimming newlines after an inner <pre> closes inside an outer <pre>

Validation:
- pnpm test:unit -- test/unit/modules/compiler/parser.spec.ts


edison1105:
Please note Vue 2 has reached End of Life. PRs will never be merged anymore.

PR #13307, opened 2026-02-03
Fix memory leak in functional components by tearing down render watchers

KashviYadav09:
Functional components create render watchers that were not properly cleaned up
on unmount, leading to retained dependencies and memory leaks.

This PR tracks functional watchers at vnode level and ensures teardown during
patch destroy and lifecycle cleanup. Includes a regression test.


mufasacubana8-pixel (changes requested):
Response ()

KashviYadav09:
Yes Please....

PR #13380, opened 2026-07-03
fix(compiler): warn when binding Vue directives

### LINKED ISSUE #10191: Warn if colon shorthand is used on v-if/v-html/etc.

Galavic:
## Summary
- warn when built-in Vue directives are accidentally used with v-bind/v-on syntax, such as :v-if or @v-html`n- add parser coverage for the new warning

Fixes #10191

## Test
- .\node_modules\.bin\vitest.cmd run test/unit/modules/compiler/parser.spec.ts


daltino (commented):
The implementation is missing the actual `warnIfDirectiveBinding` function definition, the patch only shows the call site but never defines the function, so this would fail to compile. Also, the directive list in `makeMap` includes `'bind'` and `'on'` themselves, which means `:v-bind` and `@v-on` would also trigger the warning (arguably correct, but worth a deliberate decision), and `'slot'` is included even though `v-slot` is a legitimate directive users interact with via shorthand `#`. More importantly, the check only runs for `v-bind` and `v-on` shorthand paths, someone writing `:v-if` with the full `v-bind:v-if` syntax would silently slip through without any warning, which seems like an inconsistency worth addressing.

